PR #1, opened 2024-08-09
Comments

ylacombe:
I've created this PR to be able to make some comments


ylacombe:
It's not the best format to comment the code, but let's try it.

First of all, congrats on the code, it's looking already really strong!

For now, and before being able to play with it, here are some remarks:

**[Parler Streamer](https://github.com/eustlb/speech-to-speech/blob/62c4d6e2d36f943d0e978ba271a117b3709cb9c6/utils.py#L131):** Unless you've made some changes on the class, we've integrated the streamer directly in the Parler repo ([here](https://github.com/huggingface/parler-tts/blob/main/parler_tts/streamer.py)), you can see how to import it [here](https://github.com/huggingface/parler-tts/blob/main/INFERENCE.md#how-can-i-use-it). All in all, it'll make the code easier to read.

**Requirements:** It'd be great to have a requirements.txt or a [setup.py](https://github.com/huggingface/parler-tts/blob/main/setup.py), the former being probably more than enough for now. As it is, we don't really know what your project relies on!

**README:** This is absolutely required. For now, you can create now for a really quick reproducer (for us to be able to try the code and understand the logic). Before release, it'd be nice to have something that:
1. Introduce the work
2. Explain how to install it and use it
3. Explain the motivation behind it
4. Maybe a FAQ
In that regards, the [Distil-Whisper README](https://github.com/huggingface/distil-whisper/blob/main/README.md) or the [Data-Speech README](https://github.com/huggingface/dataspeech) are great examples of what we're trying to achieve

I'll do a more thorough review once the previous comments are addressed, but some small ones in the meantime:
https://github.com/eustlb/speech-to-speech/blob/62c4d6e2d36f943d0e978ba271a117b3709cb9c6/listen_and_play.py#L7-L14

Would be great to have those configurable (maybe not CHANNEL), depending on the models and config chosen.


https://github.com/eustlb/speech-to-speech/blob/62c4d6e2d36f943d0e978ba271a117b3709cb9c6/s2s_pipeline.py#L452-L453
Here you should be able to make it configurable with the CLI. You can get inspiration from [this code](https://github.com/huggingface/dataspeech/blob/main/main.py)

andimarafioti:
Thank you for your comments @ylacombe , these are a bit outdated now so I'm closing this 🙏

PR #7, opened 2024-08-15
Fix missing sounddevice module

AlexHayton:
Without this I get the following after installation of pip packages:

```
$ python listen_and_play.py --host localhost

  File "/Users/alex.hayton/funtime/speech-to-speech/listen_and_play.py", line 6, in <module>
    import sounddevice as sd
ModuleNotFoundError: No module named 'sounddevice'
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 nltk==3.8.2
 parler_tts @ git+https://github.com/huggingface/parler-tts.git
 torch==2.4.0
+sounddevice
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
 nltk==3.8.2
 parler_tts @ git+https://github.com/huggingface/parler-tts.git
 torch==2.4.0
+sounddevice
```
andimarafioti, line 4:
```suggestion
sounddevice==0.5.0
```

andimarafioti: I would specify the version number to stay consistent.

andimarafioti (changes requested):
Great suggestion! let's specify the version number and merge this one 👍

AlexHayton:
@andimarafioti Great! Version number updated

andimarafioti (approved):
awesome! thanks!

PR #40, opened 2024-08-23
Allow LM selection and MLX Gemma

RonanKMcGovern:
Allow the language model to be selected via command line when running on Mac with MLX.
Remove the system message if using Gemma (otherwise the chat templating will throw an error).


andimarafioti:
Cool stuff! The only issue I see is that the default model in the language model won't work for MLX, so we would need to change that.

andimarafioti:
I'll merge this and make that change quickly

RonanKMcGovern:
great thanks

PR #48, opened 2024-08-26
feat:add paraformer_zh asr

wuhongsheng:
dd paraformer_zh asr


andimarafioti:
Really cool! Do we need to update the requirements? does this also work for mac? THere is a requirements_mac as well

PR #60, opened 2024-08-30
Add support for multiple languages

andimarafioti:
This PR adds support for:

- Chinese
- French
- Spanish
- Japanese
- Korean
- multi-language (all of the above + English)

For multi-language, I'm using this command:
`python s2s_pipeline.py --recv_host 0.0.0.0 --send_host 0.0.0.0 --lm_model_name meta-llama/Meta-Llama-3.1-8B-Instruct --init_chat_role system --tts melo --stt_model_name openai/whisper-large-v3 --init_chat_prompt "You are a helpful and friendly AI assistant. You are polite, respectful, and aim to provide concise responses of less than 20 words. Users will talk to you in several different languages, and will change language during a conversation. Always switch immediately to the language from the user's last message." --language None`

For any other language, you can use:
`python s2s_pipeline.py --recv_host 0.0.0.0 --send_host 0.0.0.0 --lm_model_name meta-llama/Meta-Llama-3.1-8B-Instruct --init_chat_role system --tts melo --stt_model_name openai/whisper-large-v3 --language zh`

We can't use distil whisper for languages other than English, and we can't use parler for languages other than English. Also, the LM model needs to be able to speak the language you choose.

### DIFF
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -101,7 +101,7 @@ class LanguageModelHandler(BaseHandler):
                 f"{self.__class__.__name__}:  warmed up! time: {start_event.elapsed_time(end_event) * 1e-3:.3f} s"
             )
 
-    def process(self, prompt):
+    def process(self, prompt, language_id=None):
         logger.debug("infering language model...")
 
         self.chat.append({"role": self.user_role, "content": prompt})
@@ -128,4 +128,4 @@ class LanguageModelHandler(BaseHandler):
         self.chat.append({"role": "assistant", "content": generated_text})
 
         # don't forget last sentence
-        yield printable_text
+        yield (printable_text, language_id)
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -101,8 +101,11 @@ class LanguageModelHandler(BaseHandler):
                 f"{self.__class__.__name__}:  warmed up! time: {start_event.elapsed_time(end_event) * 1e-3:.3f} s"
             )
 
-    def process(self, prompt, language_id=None):
+    def process(self, prompt):
         logger.debug("infering language model...")
+        language_id = None
+        if isinstance(prompt, tuple):
+            prompt, language_id = prompt
 
         self.chat.append({"role": self.user_role, "content": prompt})
         thread = Thread(
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -69,7 +69,7 @@ class LanguageModelHandler(BaseHandler):
     def warmup(self):
         logger.info(f"Warming up {self.__class__.__name__}")
 
-        dummy_input_text = "Write me a poem about Machine Learning."
+        dummy_input_text = "Repeat the word 'home'."
         dummy_chat = [{"role": self.user_role, "content": dummy_input_text}]
         warmup_gen_kwargs = {
             "min_new_tokens": self.gen_kwargs["min_new_tokens"],
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -18,6 +18,15 @@ logger = logging.getLogger(__name__)
 console = Console()
 
 
+WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
+    "<|en|>": "english",
+    "<|fr|>": "french",
+    "<|es|>": "spanish",
+    "<|zh|>": "chinese",
+    "<|ja|>": "japanese",
+    "<|ko|>": "korean",
+}
+
 class LanguageModelHandler(BaseHandler):
     """
     Handles the language model part.
@@ -106,6 +115,7 @@ class LanguageModelHandler(BaseHandler):
         language_id = None
         if isinstance(prompt, tuple):
             prompt, language_id = prompt
+            prompt = f"Please reply to my message in {WHISPER_LANGUAGE_TO_LLM_LANGUAGE[language_id]}. " + prompt
 
         self.chat.append({"role": self.user_role, "content": prompt})
         thread = Thread(
@@ -125,7 +135,7 @@ class LanguageModelHandler(BaseHandler):
                 printable_text += new_text
                 sentences = sent_tokenize(printable_text)
                 if len(sentences) > 1:
-                    yield (sentences[0])
+                    yield (sentences[0], language_id)
                     printable_text = new_text
 
         self.chat.append({"role": "assistant", "content": generated_text})
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -19,12 +19,12 @@ console = Console()
 
 
 WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
-    "<|en|>": "english",
-    "<|fr|>": "french",
-    "<|es|>": "spanish",
-    "<|zh|>": "chinese",
-    "<|ja|>": "japanese",
-    "<|ko|>": "korean",
+    "en": "english",
+    "fr": "french",
+    "es": "spanish",
+    "zh": "chinese",
+    "ja": "japanese",
+    "ko": "korean",
 }
 
 class LanguageModelHandler(BaseHandler):
@@ -112,10 +112,10 @@ class LanguageModelHandler(BaseHandler):
 
     def process(self, prompt):
         logger.debug("infering language model...")
-        language_id = None
+        language_code = None
         if isinstance(prompt, tuple):
-            prompt, language_id = prompt
-            prompt = f"Please reply to my message in {WHISPER_LANGUAGE_TO_LLM_LANGUAGE[language_id]}. " + prompt
+            prompt, language_code = prompt
+            prompt = f"Please reply to my message in {WHISPER_LANGUAGE_TO_LLM_LANGUAGE[language_code]}. " + prompt
 
         self.chat.append({"role": self.user_role, "content": prompt})
         thread = Thread(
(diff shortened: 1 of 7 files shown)


STT/whisper_stt_handler.py
```diff
@@ -100,11 +117,24 @@ def process(self, spoken_prompt):
...
+
+        if language_id not in SUPPORTED_LANGUAGES:  # reprocess with the last language
+            logger.warning("Whisper detected unsupported language:", language_id)
+            gen_kwargs = copy(self.gen_kwargs)
+            gen_kwargs['language'] = self.last_language
+            language_id = self.last_language
+            pred_ids = self.model.generate(input_features, **gen_kwargs)
+        else:
+            self.last_language = language_id
+
```
eustlb, line 130: Not sure to get the logic here. From what I understand, when the detected language is not supported by what's coming after (namely MeloTTS), we re-transcribe using the previously detected language. Even if in practice it tends to work (if you have a french sentence and transcribe by setting language to spanish it will actually translate), this is not an expected behavior for Whisper and might lead to hallucinations (see [here](https://github.com/openai/whisper?tab=readme-ov-file#approach) for expected usage). Maybe it would be better to prompt the LLM with `"Repeat: The language you have used is not supported. Please talk in one of the expected languages"` and set language_code to `en` in this case.

STT/whisper_stt_handler.py
```diff
@@ -100,11 +117,24 @@ def process(self, spoken_prompt):
...
+            self.last_language = language_id
+        
         pred_text = self.processor.batch_decode(
             pred_ids, skip_special_tokens=True, decode_with_timestamps=False
         )[0]
+        language_id = self.processor.tokenizer.decode(pred_ids[0, 1])
 
         logger.debug("finished whisper inference")
         console.print(f"[yellow]USER: {pred_text}")
+        logger.debug(f"Language ID Whisper: {language_id}")
```
eustlb, line 138:
```suggestion
        logger.debug(f"Language code (Whisper): {language_id}")
```

STT/whisper_stt_handler.py
```diff
@@ -100,11 +117,24 @@ def process(self, spoken_prompt):
...
 
-        yield pred_text
+        yield (pred_text, language_id)
```
eustlb, line 140:
```suggestion
        yield (pred_text, language_code)
```

TTS/melo_handler.py
```diff
@@ -10,21 +10,46 @@
...
+
+WHISPER_LANGUAGE_TO_MELO_SPEAKER = {
+    "<|en|>": "EN-Newest",
+    "<|fr|>": "FR",
+    "<|es|>": "ES",
+    "<|zh|>": "ZH",
+    "<|ja|>": "JP",
+    "<|ko|>": "KR",
+}
+
```
eustlb, line 30:
```suggestion
WHISPER_LANGUAGE_TO_MELO_LANGUAGE = {
    "en": "EN_NEWEST",
    "fr": "FR",
    "es": "ES",
    "zh": "ZH",
    "ja": "JP",
    "ko": "KR",
}

WHISPER_LANGUAGE_TO_MELO_SPEAKER = {
    "en": "EN-Newest",
    "fr": "FR",
    "es": "ES",
    "zh": "ZH",
    "ja": "JP",
    "ko": "KR",
}

```
Let's rather use ISO language codes.

TTS/melo_handler.py
```diff
@@ -10,21 +10,46 @@
...
         gen_kwargs={},  # Unused
         blocksize=512,
     ):
         self.should_listen = should_listen
         self.device = device
-        self.model = TTS(language=language, device=device)
-        self.speaker_id = self.model.hps.data.spk2id[speaker_to_id]
+        self.language = (
+            "<|" + language + "|>"
+        )  # 'Tokenize' the language code to do less operations
```
eustlb, line 46:
```suggestion
```

Unnecessary if we directly work with language codes.

TTS/melo_handler.py
```diff
@@ -10,21 +10,46 @@
...
+        self.model = TTS(
+            language=WHISPER_LANGUAGE_TO_MELO_LANGUAGE[self.language], device=device
+        )
+        self.speaker_id = self.model.hps.data.spk2id[
+            WHISPER_LANGUAGE_TO_MELO_SPEAKER["<|" + speaker_to_id + "|>"]
+        ]
```
eustlb, line 52:
```suggestion
            WHISPER_LANGUAGE_TO_MELO_SPEAKER[speaker_to_id]
        ]
```

TTS/melo_handler.py
```diff
@@ -33,7 +58,28 @@ def warmup(self):
         _ = self.model.tts_to_file("text", self.speaker_id, quiet=True)
 
     def process(self, llm_sentence):
+        language_id = None
+
+        if isinstance(llm_sentence, tuple):
+            llm_sentence, language_id = llm_sentence
+
```
eustlb, line 65:
```suggestion
        language_code = None

        if isinstance(llm_sentence, tuple):
            llm_sentence, language_code = llm_sentence

```

TTS/melo_handler.py
```diff
@@ -33,7 +58,28 @@ def warmup(self):
...
+                )
+                self.speaker_id = self.model.hps.data.spk2id[
+                    WHISPER_LANGUAGE_TO_MELO_SPEAKER[language_id]
+                ]
+                self.language = language_id
+            except KeyError:
+                console.print(
+                    f"[red]Language {language_id} not supported by Melo. Using {self.language} instead."
+                )
+
```
eustlb, line 82:
```suggestion

        if language_code is not None and self.language != language_code:
            try:
                self.model = TTS(
                    language=WHISPER_LANGUAGE_TO_MELO_LANGUAGE[language_codd],
                    device=self.device,
                )
                self.speaker_id = self.model.hps.data.spk2id[
                    WHISPER_LANGUAGE_TO_MELO_SPEAKER[language_code]
                ]
                self.language = language_code
            except KeyError:
                console.print(
                    f"[red]Language {language_code} not supported by Melo. Using {self.language} instead."
                )

```

eustlb (commented):
Great work Andi !! 🐐🤗

Main suggestion is to use ISO language code rather than Whisper's language tokens (i.e use `en` rather than `<|en|>`). This way we can stick with Whisper Transformer's nomenclature and use directly languages available in Whisper Transformer's implementation (see [here](https://github.com/huggingface/transformers/blob/6b7d64ac1c5d19ab4b9e99142a08e31741106e4e/src/transformers/models/whisper/tokenization_whisper.py#L87)). Also, I am unsure about the way we handle unsupported languages.

Moreover, when this PR will be merged, I think it will be the good moment for me to do a README update to include all these changes and make usage as easy as possible for users.

andimarafioti:
rebased on main

STT/whisper_stt_handler.py
```diff
@@ -24,12 +33,18 @@ def setup(
...
         torch_dtype="float16",
         compile_mode=None,
+        language=None,
         gen_kwargs={},
     ):
         self.device = device
         self.torch_dtype = getattr(torch, torch_dtype)
         self.compile_mode = compile_mode
         self.gen_kwargs = gen_kwargs
+        if language == 'None':
```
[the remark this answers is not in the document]
andimarafioti: I had it first set as None, but I think the performance is considerably better when we set a language. And as we support more and more languages, this can be an issue (whisper is not that good at detecting the spoken language). I think steering the users towards selecting one (or in the future a few) languages will generally improve their experience.

LLM/language_model.py
```diff
@@ -18,6 +18,15 @@
...
 
+WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
+    "<|en|>": "english",
+    "<|fr|>": "french",
+    "<|es|>": "spanish",
+    "<|zh|>": "chinese",
+    "<|ja|>": "japanese",
+    "<|ko|>": "korean",
+}
+
```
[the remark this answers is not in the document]
andimarafioti: This would break the logic. The reason for having a smaller dictionary is to force whispers to re-infer if it thinks that the spoken language is one we don't support yet. This considerably improves the user experience, because it doesn't happen that whisper thinks you are speaking Italian and then the LLM answers Italian, and then the voice speaks the Italian text with an English voice.

LLM/language_model.py
```diff
@@ -103,6 +112,10 @@ def warmup(self):
 
     def process(self, prompt):
         logger.debug("infering language model...")
+        language_id = None
```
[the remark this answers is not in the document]
andimarafioti: Changed it also for the whisper handler.

LLM/language_model.py
```diff
@@ -103,6 +112,10 @@ def warmup(self):
...
+        if isinstance(prompt, tuple):
+            prompt, language_id = prompt
+            prompt = f"Please reply to my message in {WHISPER_LANGUAGE_TO_LLM_LANGUAGE[language_id]}. " + prompt
```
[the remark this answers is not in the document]
andimarafioti: changed all _id to _code

STT/whisper_stt_handler.py
```diff
@@ -100,11 +117,24 @@ def process(self, spoken_prompt):
 
         input_features = self.prepare_model_inputs(spoken_prompt)
         pred_ids = self.model.generate(input_features, **self.gen_kwargs)
+        language_id = self.processor.tokenizer.decode(pred_ids[0, 1])
```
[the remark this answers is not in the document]
andimarafioti: I just put the [2:-2] inside of the first line and added a comment

STT/whisper_stt_handler.py
andimarafioti: changed

TTS/melo_handler.py
andimarafioti: changed

TTS/melo_handler.py
andimarafioti: changed

PR #61, opened 2024-08-30
feat:add DeepFilterNet for speech enhancement to obtain clear speech …

wuhongsheng:


andimarafioti (approved):
Really cool stuff! I'll set the default to False to avoid adding latency, but it's great to have this, thank you!

PR #82, opened 2024-09-04
add Facebook MMS TTS with expanded language support

rs545837:
This PR introduces support for multiple languages in the speech-to-speech pipeline by integrating Facebook's MMS TTS model.

### Key Changes:
- Added Facebook mms handler and arguments file
- Added a new option to specify the language for TTS with the `--tts_language` flag.
- Support for Hindi (and other languages) using the `facebookmms` TTS model.
- Extended the pipeline to handle multilingual speech synthesis.

### Example usage with Hindi:

```bash
python s2s_pipeline.py --recv_host 0.0.0.0 --send_host 0.0.0.0 --stt_model_name openai/whisper-medium --tts_language hi --language hi --tts facebookmms --lm_model_name meta-llama/Meta-Llama-3.1-8B-Instruct

### DIFF
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -25,6 +25,7 @@ WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
     "zh": "chinese",
     "ja": "japanese",
     "ko": "korean",
+    "hi": "hindi",
 }
 
 class LanguageModelHandler(BaseHandler):
--- a/STT/whisper_stt_handler.py
+++ b/STT/whisper_stt_handler.py
@@ -19,6 +19,7 @@ SUPPORTED_LANGUAGES = [
     "zh",
     "ja",
     "ko",
+    "hi"
 ]
 
 
--- a/TTS/facebookmms_handler.py
+++ b/TTS/facebookmms_handler.py
@@ -15,11 +15,45 @@ logger = logging.getLogger(__name__)
 console = Console()
 
 WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE = {
-    "en": "eng",
-    "fr": "fra",
-    "es": "spa",
-    "ko": "kor",
-    "hi": "hin",
+    "en": "eng", # English
+    "fr": "fra", # French
+    "es": "spa", # Spanish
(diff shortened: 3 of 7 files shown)


andimarafioti:
Is this up for review or is it a draft?

rs545837:
Yeah, Hindi is supported with this PR, I was just curious, should we add all the languages with the same PR as well?

andimarafioti:
I would add all the languages in TTS/facebookmms_handler.py such that we know which ones are supported by facebookmms at least

rs545837:
List of all the supported languages by facebook-mms is [here](https://huggingface.co/facebook/mms-tts) under a toggle button, there are 1000+ languages, and adding all will also require the lm and stt selection accordingly, maybe for a start we can add all the languages supported by whisper that is likely less than 100 languages.

eustlb:
We're seeing here issues entailed with having such a modular pipeline and parts that support different languages. I also think that since Whisper will be mostly used, starting by adding its supported languages makes sense here. 
We could benefit from the `--language` flag to print warnings and fall back to default english when language set for the STT part do not match the TTS part. Nevertheless, we also have the use case where we want STT to be in another language than TTS, e.g. a translator. We could then also go back to setting directly STT and TTS language using specific flags` --stt_language` and `--tts_language`. 
 Maybe here starting to add a clear abstraction for the STT and TTS would make sense, as we could ensure that supported languages are provided when a new implementation is added. Writing updated docs, I am realizing that some uniformisation could clearly benefit the project. WDYT ?

rs545837:
Alright, I added the languages supported by both whisper and mms-tts as of now.

The list of supported languages and their respective iso codes were picked up from [OpenAI Docs](https://platform.openai.com/docs/guides/speech-to-text/supported-languages) and  [MMS - Language Coverage](https://dl.fbaipublicfiles.com/mms/misc/language_coverage_mms.html).

andimarafioti:
I'm merging this into an intermediate branch to fix some stuff before going to main. It's been a while since you opened the PR but there are a few things to fix

andimarafioti (approved):

Issue #86, opened 2024-09-05
Exception :  Sizes of tensors must match except in dimension 1


SaddamBInSyed:
Hi ,

@andimarafioti 

when i run the  "python s2s_pipeline.py" with mentioned arguments in teh readme file ,  When i started to speak I am getting the below exception 

(/home/user/Downloads/speech-to-speech/sts) ~/PycharmProjects/speech-to-speech git:[multi-language]
python s2s_pipeline.py 	--recv_host 0.0.0.0 	--send_host 0.0.0.0 	--lm_model_name microsoft/Phi-3-mini-4k-instruct 	--init_chat_role system 	--stt_compile_mode reduce-overhead 	--tts_compile_mode default
/home/user/Downloads/speech-to-speech/sts/lib/python3.10/site-packages/df/io.py:9: UserWarning: `torchaudio.backend.common.AudioMetaData` has been moved to `torchaudio.AudioMetaData`. Please update the import path.
  from torchaudio.backend.common import AudioMetaData
[nltk_data] Downloading package averaged_perceptron_tagger_eng to
[nltk_data]     /home/user/nltk_data...
[nltk_data]   Package averaged_perceptron_tagger_eng is already up-to-
[nltk_data]       date!
Using cache found in /home/user/.cache/torch/hub/snakers4_silero-vad_master
2024-09-05 14:53:45,040 - STT.whisper_stt_handler - INFO - Warming up WhisperSTTHandler
You have passed task=transcribe, but also have set `forced_decoder_ids` to [[1, None], [2, 50360]] which creates a conflict. `forced_decoder_ids` will be ignored in favor of task=transcribe.
The attention mask is not set and cannot be inferred from input because pad token is same as eos token.As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
2024-09-05 14:54:06,376 - STT.whisper_stt_handler - INFO - WhisperSTTHandler:  warmed up! time: 21.336 s
2024-09-05 14:54:07,457 - transformers_modules.microsoft.Phi-3-mini-4k-instruct.5a516f86087853f9d560c95eb9209c1d4ed9ff69.modeling_phi3 - WARNING - `flash-attention` package not found, consider installing for better performance: No module named 'flash_attn'.
[...]

PR #87, opened 2024-09-05
fix

andimarafioti:
hot fix for #86


andimarafioti:
tested and it works 👍

PR #91, opened 2024-09-07
fix: Changed [True] to [False] in help text for audio_enhancement to align with actual default

BrutalCoding:
Fixed help comment mismatch with actual default value.

Introduced in this commit: https://github.com/huggingface/speech-to-speech/commit/5e667f94a43152fec53d561213c630d0fbc40721


andimarafioti (approved):
Easy review for me, thank you :)

PR #106, opened 2024-09-18
Refactor for inference

andimarafioti:
I needed to be able to obtain the pipeline to deploy it on an inference endpoint, so I did a little refactor of the whole file.

### DIFF
--- a/s2s_pipeline.py
+++ b/s2s_pipeline.py
@@ -164,6 +164,18 @@ def prepare_all_args(
     prepare_args(chat_tts_handler_kwargs, "chat_tts")
 
 
+def initialize_queues_and_events():
+    return {
+        "stop_event": Event(),
+        "should_listen": Event(),
+        "recv_audio_chunks_queue": Queue(),
+        "send_audio_chunks_queue": Queue(),
+        "spoken_prompt_queue": Queue(),
+        "text_prompt_queue": Queue(),
+        "lm_response_queue": Queue(),
+    }
+
+
 def build_pipeline(
     module_kwargs,
     socket_receiver_kwargs,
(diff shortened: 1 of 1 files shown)


s2s_pipeline.py
```diff
@@ -330,14 +321,69 @@ def overwrite_device_argument(common_device: Optional[str], *handler_kwargs):
...
+        language_model_handler_kwargs,
+        mlx_language_model_handler_kwargs,
+        parler_tts_handler_kwargs,
+        melo_tts_handler_kwargs,
+        chat_tts_handler_kwargs,
+    ) = parse_arguments()
+
+    setup_logger(module_kwargs.log_level)
+
+    optimal_mac_settings(
```
eustlb, line 354: Just to go a bit further in the refactoring, do you think having `optimal_mac_settings`, `check_mac_setttings` and `overwrite_device_argument` in prepare_all_args would improve clarity ?

s2s_pipeline.py
```diff
@@ -84,69 +84,40 @@ def main():
...
-            vad_handler_kwargs,
-            whisper_stt_handler_kwargs,
-            paraformer_stt_handler_kwargs,
-            language_model_handler_kwargs,
-            mlx_language_model_handler_kwargs,
-            parler_tts_handler_kwargs,
-            melo_tts_handler_kwargs,
-            chat_tts_handler_kwargs,
-        ) = parser.parse_json_file(json_file=os.path.abspath(sys.argv[1]))
+        return parser.parse_json_file(json_file=os.path.abspath(sys.argv[1]))
```
eustlb, line 88:
```suggestion
        # Parse configurations from a JSON file if specified
        return parser.parse_json_file(json_file=os.path.abspath(sys.argv[1]))
```

s2s_pipeline.py
```diff
@@ -84,69 +84,40 @@ def main():
...
-            paraformer_stt_handler_kwargs,
-            language_model_handler_kwargs,
-            mlx_language_model_handler_kwargs,
-            parler_tts_handler_kwargs,
-            melo_tts_handler_kwargs,
-            chat_tts_handler_kwargs,
-        ) = parser.parse_args_into_dataclasses()
-
-    # 1. Handle logger
+        return parser.parse_args_into_dataclasses()
```
eustlb, line 90:
```suggestion
        # Parse arguments from command line if no JSON file is provided
        return parser.parse_args_into_dataclasses()
```

s2s_pipeline.py
```diff
@@ -84,69 +84,40 @@ def main():
...
     logging.basicConfig(
-        level=module_kwargs.log_level.upper(),
+        level=log_level.upper(),
         format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
     )
     logger = logging.getLogger(__name__)
 
-    # torch compile logs
-    if module_kwargs.log_level == "debug":
+    if log_level == "debug":
```
eustlb, line 101:
```suggestion
    # torch compile logs
    if log_level == "debug":
```

eustlb (commented):
Great work in refactoring @andimarafioti! 🙏
This now makes us able to import the pipeline from the module.
Close to merging, just a nit and some comments that should not disappear IMO.

s2s_pipeline.py
andimarafioti: Not really, maybe the name for `prepare_args` is wrong, but the function has a very clear goal and adding these things would only make it more convoluted. We could rename `prepare_args` to something like -> `rename_args` and then have another method `prepare_args` where these methods also took part. WDYT?

eustlb: Yep indeed it is better this way! Let's have `prepare_args` renamed `rename_args` (I should have named it this way from the beginning), `optimal_mac_settings` and `if platform == "darwin":` part in a dedicated `prepare_module_args` and calls to `rename_args` and `overwrite_device_arguments` in `prepare_args` ?

andimarafioti: done!

eustlb:
LGTM! thanks 🤗

PR #112, opened 2024-09-23
Improve auto language

eustlb:
This PR improves multilingual handling. In the current implementation, Whisper (both Transformer and MLX) always returns a tuple. As a result, the prompt "`Please reply to my message in ...`" is always prepended to the LLM prompt, which is unnecessary when the language does not change mid-conversation.

Here, we consider two use cases:
- Either we enforce the language setting using the `--language` option, specifying the target language code (default is `en`).
- Or we set `--language` to `auto`. In this case, Whisper detects the language for each spoken prompt, and the LLM is prompted with "`Please reply to my message in ...`" to ensure the response is in the detected language.

### DIFF
--- a/STT/whisper_stt_handler.py
+++ b/STT/whisper_stt_handler.py
@@ -137,4 +137,7 @@ class WhisperSTTHandler(BaseHandler):
         console.print(f"[yellow]USER: {pred_text}")
         logger.debug(f"Language Code Whisper: {language_code}")
 
+        if self.language is None:
+            language_code += "-auto"
+            
         yield (pred_text, language_code)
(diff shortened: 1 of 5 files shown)


STT/whisper_stt_handler.py
```diff
@@ -137,4 +137,7 @@ def process(self, spoken_prompt):
         console.print(f"[yellow]USER: {pred_text}")
         logger.debug(f"Language Code Whisper: {language_code}")
 
+        if self.language is None:
+            language_code += "-auto"
+
```
andimarafioti, line 142: There is no `self.language`. Did you test this ?

andimarafioti (changes requested):

STT/whisper_stt_handler.py
eustlb: My bad, forgot to commit a piece of code (`self.language = language`). I took the opportunity to make it a bit more clear and coherent with the Whisper MLX implementation. I also tested both local and server 👌

PR #113, opened 2024-09-23
Readme update + clarity improvements

eustlb:
This PR updates the README to better reflect the newly added speech-to-speech functionalities and implements a few minor improvements for clarity (e.g., the flag `--stt_model_name` has been replaced with `--whisper_model_name` since STT models other than Whisper can now be used).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,20 +23,32 @@
 ## Approach
 
 ### Structure
-This repository implements a speech-to-speech cascaded pipeline with consecutive parts:
-1. **Voice Activity Detection (VAD)**: [silero VAD v5](https://github.com/snakers4/silero-vad)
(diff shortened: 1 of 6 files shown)


andimarafioti (approved):
I like this but it also introduces breaking changes (the param changes). We should start making releases to avoid breaking peoples applications by changing stuff like this :/

eustlb:
Indeed, I was wondering about that as well. Should we merge the parameter change now or wait for a beta release? I agree that we could stick to just updating the README for now. Having unclear parameters is still better than introducing breaking changes. WDYT?

andimarafioti:
I agree, we can do a first release without breaking anything else and then start working on breaking changes

PR #118, opened 2024-09-27
Inference endpoint

andimarafioti:
Code for the deployment to an inference endpoint.

I think here we need to be careful with a few changes to the main models that are good, but need to be consistent. For example, I'm yielding "Done" after the language model and the speech model are done. That allows me to have the client know that the server is done. It could also help us with the local and the ssh server approach. But we need to make sure that it works for them and it doesn't break. Also that all the TTS/language models handle this similarly.

### DIFF
--- a/LLM/language_model.py
+++ b/LLM/language_model.py
@@ -142,3 +142,4 @@ class LanguageModelHandler(BaseHandler):
 
         # don't forget last sentence
         yield (printable_text, language_code)
+        yield b"DONE"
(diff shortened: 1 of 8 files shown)


andimarafioti:
Another thing to consider is that this client and endpoint handler make sense with the webservice that I have over at https://github.com/andimarafioti/speech-to-speech-inference-toolkit/pull/1/files so maybe we want to get those files here as well.

PR #121, opened 2024-10-10
Fix hanging client on KeyboardInterrupt

3manifold:
Given that `socket.recv()` is blocking in `receive_data_chunk()`, shut it down to allow the thread to continue. Client hangs because `recv_thread.join()` is waiting forever. Worth noting that shutdown is about closing off the connection, and `close()` is about closing the file descriptor and discarding any data that may be in the file description's buffer. Fixes https://github.com/huggingface/speech-to-speech/issues/122 .

### DIFF
--- a/listen_and_play.py
+++ b/listen_and_play.py
@@ -113,6 +113,8 @@ def listen_and_play(
 
     finally:
         stop_event.set()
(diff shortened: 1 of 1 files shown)


andimarafioti (approved):
Great job, thank you!

Issue #97, opened 2024-09-09
Ollama support for LLM backend


rchan26:
Somewhat related to #65 

In the past I've used Ollama for local inference for LLMs, would it be useful to add this support to the library? I'd be happy to work on this to add support to use an Ollama API endpoint for the LLM

andimarafioti:
can ollama use APIs? There is a PR open for the API part.

rchan26:
@andimarafioti I think the PR is only for OpenAI API. I was just suggesting to offer Ollama API support. In particular, it's possible to run a Ollama server (either locally on a slightly better machine) and then query it using the Ollama Rest API (like [here](https://github.com/ollama/ollama?tab=readme-ov-file#rest-api)). 

The changes/additions would be similar to the open PR #81 but a class for Ollama API support and querying Ollama endpoints

mattfro:
This would be great, when some people already run ollama. Like me :)

rchan26:
happy to work on this at some point this week. there's added functionality to Ollama last week to run any GGUF model from the HF Hub too: https://huggingface.co/docs/hub/en/ollama

rchan26:
@mattfro sorry for the delay, but I finally got round to working on this. See #143 for my implementation of this :)

andimarafioti:
Since we offer support for APIs, this isn't necessary. You can connect to ollama already. We have examples with llama-cpp

PR #143, opened 2024-11-28
Add Ollama option to choice of LLMs

rchan26:
As discussed in #97. I have added an option to use Ollama as the LLM. By default the model is a quantised [SmolLM-1.7B-Instruct](https://huggingface.co/HuggingFaceTB/SmolLM-1.7B-Instruct).

For example, if you're running an [Ollama](https://ollama.com/) server from `http://127.0.0.1:8000`, you can run:
```
python s2s_pipeline.py \
    --local \
    --device mps \
    --llm ollama \
    --ollama_api_endpoint http://127.0.0.1:8000 
```

By default the `ollama_api_endpoint` argument is `http://localhost:11434` which is the default host when spinning up an Ollama endpoint.

Note that you must have downloaded the model for Ollama prior to running, so first run
```
ollama pull hf.co/HuggingFaceTB/SmolLM2-1.7B-Instruct-GGUF
```
before starting an endpoint with
```
ollama serve
```

Also updated a few help comments.

### DIFF
--- a/LLM/ollama_language_model.py
+++ b/LLM/ollama_language_model.py
@@ -1,0 +1,87 @@
+from threading import Thread
+from ollama import Client
+import torch
+
+from LLM.chat import Chat
+from baseHandler import BaseHandler
+from rich.console import Console
+import logging
+from nltk import sent_tokenize
+
+logger = logging.getLogger(__name__)
+
+console = Console()
+
+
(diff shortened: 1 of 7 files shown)


andimarafioti:
This was a good idea, but I got too late to it. Still, I think you can serve the models with ollama and connect with speech to speech to them, so it's probably already implemented in a way

PR #147, opened 2024-12-04
Update Parler-TTS base model and description

ylacombe:


eustlb:
Great, thanks for the update @ylacombe (and for the model! 🚀)

PR #170, opened 2025-08-02
chore: add missing model options to exceptions

that-one-arab:


andimarafioti:
not sure what this is and it's stale

that-one-arab:
fair enough I forgot what this was too

PR #185, opened 2026-02-16
feat/qwen3 tts integration

andimarafioti:
- Add Qwen3 TTS handler and benchmark integration
- Fix Qwen3 TTS handler: remove hardcoded paths, add HF Hub support, optional CUDA graphs


A-Mahla (changes requested):
Great work, thanks @andimarafioti! Super nice to now support `qwen3_tts`. I left a few comments.

PR #192, opened 2026-03-03
feat(vad): integrate Smart Turn v3.2 endpointing

andimarafioti:
## Summary

- uses the latest supported `pipecat-ai/smart-turn-v3` v3.2 quantized CPU ONNX checkpoint, downloaded from the Hugging Face Hub or supplied with `--smart_turn_model_path`
- enables Smart Turn by default with `onnxruntime` as a regular dependency and supports CPU inference only
- supports Smart Turn only in realtime mode and rejects incompatible modes before model setup
- preserves ordinary Silero finalization and `max_speech_ms` enforcement, then uses Smart Turn to choose the speculative response grace
- uses `--speculative_reopen_ms` (800 ms by default) for complete turns and `--smart_turn_max_wait_ms` (2 seconds by default) for incomplete turns
- emits every accepted final segment to speculative STT/LLM work immediately; resumed speech creates a newer revision and invalidates work from the previous revision before output commits\n- keeps reopen eligibility active for the full Smart Turn grace and preserves raw accumulated audio for repeated Smart Turn analysis when downstream audio enhancement is enabled\n- adds CLI configuration, user documentation, packaging coverage, and focused revision-flow tests

## Why

Silero detects that speech stopped, but it cannot distinguish a completed turn from a mid-thought pause. Smart Turn uses the accumulated turn audio and prosody to make that distinction without holding or restoring Silero buffers.

The speculative pipeline can perform STT and LLM work during either grace period, while the existing response gates prevent stale output from reaching the user. This keeps complete turns responsive and gives incomplete turns more time to resume.

## Usage

```bash
pip install speech-to-speech
speech-to-speech
```

Smart Turn uses the quantized CPU checkpoint and `CPUExecutionProvider`. Pass `--no_smart_turn` to disable it or when selecting a non-realtime mode.

## Validation

- real `smart-turn-v3.2-cpu.onnx` inference with `CPUExecutionProvider`
- `uv pip check`: all 144 installed packages compatible
- `pytest -q`: 783 passed, 1 skipped
- focused Smart Turn, speculative-turn, and CLI tests: 79 passed
- `ruff check src/ tests/`
- `mypy src/`
- wheel and sdist build plus strict Twine metadata check


src/speech_to_speech/VAD/vad_handler.py
```diff
@@ -640,7 +738,24 @@ def _process_realtime(
...
+            # that buffer before applying max_speech_ms, or valid speech can be
+            # discarded solely because its waiting silence crossed the limit.
+            smart_turn_decision: _SmartTurnDecision | None = None
+            if self._smart_turn_pending_since_sample is not None:
+                smart_turn_decision = self._smart_turn_should_finalize(
+                    array,
+                    int(active_speech_duration_ms * self.sample_rate / 1000),
+                    analysis_audio=self._combined_turn_audio(array),
+                )
+                if smart_turn_decision is _SmartTurnDecision.CONTINUE:
```
andimarafioti: This `CONTINUE` branch is a symptom of the underlying design problem: Smart Turn rehydrates `VADIterator` after Silero has already finalized a segment. That couples low-level VAD buffering to turn endpointing, which is why waiting silence, resumed speech, `max_speech_ms`, and timeout behavior now require separate exceptions.

Please simplify this by making Smart Turn realtime-only and reusing the existing speculative-turn mechanism:

1. Always let Silero finalize the segment through the normal realtime path; do not restore it into `VADIterator`.
2. Run Smart Turn on the complete turn audio before choosing the reopen grace duration.
3. For a complete prediction, use `speculative_reopen_ms`; for an incomplete prediction, use `smart_turn_max_wait_ms` as the grace passed to `SpeculativeTurnTracker.start_reopen_grace()`.
4. Emit the final revision normally. STT/LLM may run speculatively, while the existing output gates prevent assistant output from committing until the grace expires.
5. Let resumed speech use the existing pending-candidate/revision flow. The accumulated audio prefix is re-emitted under the next revision, and existing STT/LLM/TTS stale-revision filters discard the previous work.
6. Keep the ordinary `max_speech_ms` validation before Smart Turn; remove its Smart Turn-specific exceptions.
7. Remove `_resume_smart_turn_utterance`, `_smart_turn_pending_since_sample`, `_smart_turn_last_active_samples`, `_SmartTurnDecision.MAX_WAIT`, and the duplicated pending-decision branches in realtime/normal processing.
8. Reject `--smart_turn` outside realtime mode early with a clear configuration error, since non-realtime support is not required.

Please cover: incomplete prediction selects the longer grace; resumed speech creates a newer revision and invalidates the old one; no resumed speech commits after the longer grace; `max_speech_ms` remains enforced; and non-realtime Smart Turn configuration is rejected.

PR #230, opened 2026-04-10
Replace magic string constants with strongly-typed enums for inter-component queue messages

A-Mahla:
resolve https://github.com/huggingface/speech-to-speech/issues/214


andimarafioti (approved):
Great work, thank you 🙏

PR #233, opened 2026-04-15
Add mlx-audio backend for Qwen3 TTS on macOS

andimarafioti:
## What changed

This PR adds a macOS-compatible Qwen3 TTS path by extending the existing Qwen3TTSHandler to use mlx-audio on Darwin while preserving the current faster-qwen3-tts backend for non-macOS platforms.

- auto-maps Qwen/* model IDs to mlx-community/*-bf16 on Apple Silicon
- keeps the existing voice-clone, custom-voice, and voice-design flow intact
- adds MLX locking for the macOS Qwen3 TTS path
- updates macOS dependency declarations and docs
- adds focused tests for backend selection and model-name mapping

## Why this changed

The repo already supports Qwen3 TTS through faster-qwen3-tts, but that backend does not support macOS. mlx-audio now supports Qwen3 TTS, so this PR wires it into the existing handler for Apple Silicon without changing the current non-macOS path.

## Validation

- python -m py_compile TTS/qwen3_tts_handler.py arguments_classes/qwen3_tts_arguments.py arguments_classes/module_arguments.py s2s_pipeline.py tests/test_qwen3_tts_handler_backend.py tests/test_tts_input_coalescing.py
- pytest tests/test_qwen3_tts_handler_backend.py tests/test_tts_input_coalescing.py


TTS/qwen3_tts_handler.py
```diff
@@ -346,6 +653,25 @@ def _process_custom_voice(self, text):
...
+                        instruct=self.instruct,
+                        language=self.language,
+                        max_tokens=self.max_new_tokens,
+                        verbose=False,
+                        stream=True,
+                        streaming_interval=self._mlx_streaming_interval(),
+                        **self.gen_kwargs,
+                    ),
+                    label="voice_design_mlx",
+                )
```
A-Mahla: This snippet appears three times. Wouldn’t it be better to factor it out into a function?

TTS/qwen3_tts_handler.py
```diff
@@ -269,26 +537,20 @@ def _decode(item):
 
     def process(self, llm_sentence):
         if isinstance(llm_sentence, tuple) and llm_sentence[0] == MessageTag.END_OF_RESPONSE:
+            if not getattr(self, "runtime_config", None):
+                self.should_listen.set()
```
A-Mahla, line 542: A potential side effect introduced by this change is that, by removing lines 309–311 from the finally block, `should_listen.set()` might never be called `MessageTag.END_OF_RESPONSE` (e.g. LLM crashes, ...) is never sent. Not blocking for realtime mode, but for other it introduce a potential deadlock

TTS/qwen3_tts_handler.py
```diff
@@ -60,19 +73,59 @@ def setup(
...
         self.xvec_only = xvec_only
         self.parity_mode = parity_mode
-        self.streaming_chunk_size = streaming_chunk_size
+        self.mlx_quantization = self._normalize_mlx_quantization(mlx_quantization)
         self.max_new_tokens = max_new_tokens
         self.blocksize = blocksize
+        self.gen_kwargs = gen_kwargs or {}
+        self._mlx_ref_audio_cache = {}
+        self._mlx_temp_ref_audio_files = set()
+
```
A-Mahla, line 91: `self.dtype` is only initialized in `_setup_faster()`. Let’s define also it in setup(), so all attributes are initialized independently of the backend.

```suggestion
        self.dtype = None
```

TTS/qwen3_tts_handler.py
```diff
@@ -89,30 +142,219 @@ def setup(
...
+        if model_type == "custom_voice":
+            self.speaker = session_voice
+            self.ref_audio = None
+            return
+
+        if self._resolve_audio_path(session_voice) is not None:
+            self.ref_audio = session_voice
+            return
+
+        logger.debug(
```
A-Mahla: Let’s make it non-silent when it’s ignored.

```suggestion
        logger.warning(
```

A-Mahla (changes requested):
Awesome! Thanks @andimarafioti! It works really well. I added a few comments, and then we can merge it.

A-Mahla (approved):

TTS/qwen3_tts_handler.py
andimarafioti: moved it :)

TTS/qwen3_tts_handler.py
andimarafioti: addressed it

TTS/qwen3_tts_handler.py
andimarafioti: changed

TTS/qwen3_tts_handler.py
andimarafioti: done :)

PR #236, opened 2026-04-16
Fix local live transcription queueing

andimarafioti:
## Summary
This PR fixes a bug with local live-transcription queueing. It only happens in local mode. 

## Root Cause
Different modes were using different STT-to-LLM paths. In particular, websocket and realtime already routed STT output through `TranscriptionNotifier`, while local mode could bypass that shared partial/final transcript boundary.

## What Changed
- always instantiate `TranscriptionNotifier` in `s2s_pipeline.py`
- always route STT output through it before the LLM, regardless of mode
- remove the mode-specific branching so local, websocket, and realtime share the same transcript contract

## Why This Shape
`TranscriptionNotifier` is now the single place responsible for handling partial versus final transcriptions. That keeps the filtering rule in one layer instead of duplicating defensive checks in the OpenAPI LLM handler.

## Validation
- `pytest tests/test_openai_api_language_model.py tests/test_parakeet_transcription_events.py`
- `python -m py_compile s2s_pipeline.py LLM/openai_api_language_model.py STT/transcription_notifier.py tests/test_openai_api_language_model.py tests/test_parakeet_transcription_events.py`

### DIFF
--- a/s2s_pipeline.py
+++ b/s2s_pipeline.py
@@ -285,6 +285,7 @@ def build_pipeline(
     lm_response_queue = queues_and_events["lm_response_queue"]
     lm_processed_queue = queues_and_events["lm_processed_queue"]
     text_output_queue = None  # Only set for websocket/realtime modes; kept None otherwise to avoid unbounded queue growth
+    transcription_notifier = None
     if module_kwargs.mode == "local":
         from connections.local_audio_streamer import LocalAudioStreamer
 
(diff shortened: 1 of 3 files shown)


s2s_pipeline.py
```diff
@@ -410,13 +395,18 @@ def build_pipeline(
...
     )
 
-    if module_kwargs.mode in ("websocket", "realtime"):
-        # STT outputs to an intermediate queue so TranscriptionNotifier can
-        # intercept partials and emit transcription events before forwarding
-        # finals to the LLM.
-        stt_dest = stt_output_queue
-    else:
-        stt_dest = text_prompt_queue
+    from STT.transcription_notifier import TranscriptionNotifier
```
A-Mahla, line 398: nit: I’d rather have it as a top-level import.

A-Mahla (approved):

PR #238, opened 2026-04-17
Add Qwen3-TTS non-streaming mode override

andimarafioti:
## What changed

This adds a `qwen3_tts_non_streaming_mode` override to the speech-to-speech Qwen3-TTS integration.

- Adds a tri-state CLI/config flag so unset preserves each backend's existing default behavior.
- Applies explicit `true` / `false` overrides to the faster-qwen3-tts backend.
- Passes the override through the faster voice-clone path.
- Patches faster custom-voice and voice-design setup so they can use the same override without changing the upstream package.
- Warns and ignores the override on Apple Silicon because `mlx-audio` does not expose `non_streaming_mode` yet.
- Adds backend-focused tests for the new flag behavior.

## Why

The faster-qwen3-tts backend has different effective defaults across voice-clone and custom-voice flows. This makes it hard to compare output quality and latency across modes.

This change makes the text-prefill behavior explicit and user-configurable while keeping the existing defaults intact when the flag is not set.

## Impact

Users can now run:

```bash
--qwen3_tts_non_streaming_mode true
```

or:

```bash
--qwen3_tts_non_streaming_mode false
```

to force full-text prefill or token-by-token trailing-text feeding on the faster backend.

On Apple Silicon, the flag is accepted but logged as ignored until `mlx-audio` supports the same control.

## Validation

- `python -m py_compile TTS/qwen3_tts_handler.py arguments_classes/qwen3_tts_arguments.py tests/test_qwen3_tts_handler_backend.py`
- direct parser check for `None` / `true` / `false`
- direct smoke check that the faster custom-voice override patches `_prepare_generation_custom`
- direct smoke check that the faster voice-clone path forwards the explicit override

`pytest` was not run because the current `.venv` does not have `pytest` installed.

### DIFF
--- a/TTS/qwen3_tts_handler.py
+++ b/TTS/qwen3_tts_handler.py
@@ -10,7 +10,6 @@ from pathlib import Path
 from sys import platform
 import tempfile
 from time import perf_counter
-import types
 
 import numpy as np
 from rich.console import Console
@@ -127,7 +126,6 @@ class Qwen3TTSHandler(BaseHandler):
                 dtype=dtype,
                 attn_implementation=attn_implementation,
             )
-            self._install_faster_non_streaming_mode_override()
 
         logger.info(
             "Using Qwen3-TTS streaming chunk size %d (~%.0fms audio per chunk) on %s",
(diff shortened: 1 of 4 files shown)


A-Mahla (approved):
LGTM
demo

PR #239, opened 2026-04-20
Preserve VAD pre-roll audio

andimarafioti:
## Summary

This PR preserves the beginning of user speech when VAD triggers slightly after the acoustic onset.

## What changed

- add a rolling pre-speech buffer in `VADIterator`
- prepend that buffered audio when a speech segment starts
- use the padded audio view for realtime progressive transcription as well
- update the `speech_pad_ms` help text to match the actual behavior
- add targeted tests covering pre-roll capture and buffer semantics

## Why this changed

We found that transcription could miss some audio at the start of an utterance. The code exposed `speech_pad_ms`, but the iterator did not actually retain and prepend any pre-trigger audio, so soft onsets could be clipped before STT saw them.

## User impact

- fewer clipped first syllables or words at the start of speech
- better alignment between the documented `speech_pad_ms` setting and runtime behavior
- no change to the local-mode issue that was explicitly left out of scope

## Root cause

`VADIterator` only started buffering once the speech probability crossed the VAD threshold. Any audio immediately before that threshold crossing was dropped, even when `speech_pad_ms` was configured.

## Validation

- `pytest /Users/andresmarafioti/Documents/speech-to-speech/tests/test_vad_iterator.py`
- `pytest /Users/andresmarafioti/Documents/speech-to-speech/tests/test_parakeet_transcription_events.py`

### DIFF
--- a/VAD/vad_handler.py
+++ b/VAD/vad_handler.py
@@ -170,7 +170,7 @@ class VADHandler(BaseHandler):
 
             # Yield accumulated audio periodically while speaking
             if (current_time - self.last_process_time) >= self.realtime_processing_pause:
-                array = torch.cat(self.iterator.buffer_with_pad()).cpu().numpy()
+                array = torch.cat(self.iterator.speech_buffer()).cpu().numpy()
                 duration_ms = len(array) / self.sample_rate * 1000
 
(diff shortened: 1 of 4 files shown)


andimarafioti:
Good PR after a few changes :)

PR #240, opened 2026-04-20
Scale Qwen3 TTS token budgets to utterance length

andimarafioti:
## What changed
This PR replaces the fixed Qwen3-TTS `max_new_tokens` budget with an utterance-aware estimate that scales with the amount of text being synthesized.

The estimator now:
- derives a per-utterance codec-token budget from text length
- adds a conservative flat `0.5s` pause for each punctuation mark
- aligns the requested budget to the streaming chunk size
- clamps the estimate to `qwen3_tts_max_new_tokens`, which remains configurable above the default when longer utterances are needed

It also updates the Qwen3-TTS argument help text and adds backend coverage for both the MLX and `faster-qwen3-tts` paths.

## Why this changed
Long utterances could be truncated because the handler always passed the same fixed codec-token budget to Qwen3-TTS, regardless of how much text it was asked to speak.

## Impact
This makes long responses safer by giving Qwen3-TTS more room when the utterance is longer, while still keeping a ceiling in place for callers that want to bound generation.

This branch is intentionally scoped to the Qwen3-TTS fix only and does not include the separate VAD work from the other branch.

## Validation
- `python -m pytest tests/test_qwen3_tts_handler_backend.py`
- `python -m pytest tests/test_tts_input_coalescing.py`

### DIFF
--- a/TTS/qwen3_tts_handler.py
+++ b/TTS/qwen3_tts_handler.py
@@ -6,10 +6,13 @@ Qwen3 TTS Handler
 """
 
 import logging
+import math
 from pathlib import Path
+import re
 from sys import platform
 import tempfile
 from time import perf_counter
+import unicodedata
 
 import numpy as np
 from rich.console import Console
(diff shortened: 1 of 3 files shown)


andimarafioti:
Good PR

PR #241, opened 2026-04-20
Make model handlers stateless and move session/request-related config to ConnState

A-Mahla:
Pipeline handlers (VAD, STT, LLM, TTS) stored per-connection state (chat history, instructions, tools, voice config, ...) as instance attributes. This coupled them to the realtime session lifecycle and blocked support for multiple concurrent connections sharing the same handler instances.

### What

- **Typed pipeline messages**: Replaced ad-hoc tuples and dicts flowing between pipeline stages with Pydantic dataclasses (`VADAudio`, `Transcription`, `LLMResponseChunk`, `TTSInput`, `GenerateResponseRequest`, etc.). Deleted the `MessageTag` enum.

- **`BaseHandler` is now generic**: `BaseHandler[T]` — each handler declares its input type, making the pipeline contract explicit.

- **Per-connection state on `ConnState`**: `RuntimeConfig` (which now holds `Chat`) is created per-connection inside `ConnState` instead of being a shared singleton threaded through every handler's constructor.

- **Config flows through messages, not shared objects**: `GenerateResponseRequest` carries everything the LLM needs to produce a response. `LLMResponseChunk` / `TTSInput` propagate session + per-response overrides downstream so TTS can resolve voice without reaching back into shared state.

- **Handlers are stateless**: LLM handlers no longer cache instructions, tools, or chat on `self`. VAD reads turn-detection config from the incoming message. TTS resolves voice from the `TTSInput` payload.

- **Typed pipeline events**: `text_output_queue` events (`SpeechStartedEvent`, `AssistantTextEvent`, `TranscriptionCompletedEvent`, etc.) are Pydantic models instead of raw dicts. Dispatch is keyed by `type(event)` instead of string matching.

- **Conversation items write directly to chat**: `conversation.item.create` appends to the connection's `Chat` instead of putting control tuples on the LLM queue.


andimarafioti:
Comments from codex:

Findings

  - High: /tmp/speech-to-speech-pr241/api/openai_realtime/websocket_router.py:123, /tmp/speech-to-speech-pr241/api/openai_realtime/websocket_router.py:237, /tmp/speech-to-speech-pr241/api/openai_realtime/websocket_router.py:318 — the PR
    claims to unblock multiple concurrent connections, but the realtime entrypoint still hard-rejects a second websocket (session_limit_reached). More importantly, the shared send loop still fans each pipeline text/audio item out to every
    registered connection. So the new per-connection ConnState does not actually make concurrent sessions work end-to-end; if the guard is ever removed, sessions will cross-talk.
  - High: /tmp/speech-to-speech-pr241/api/openai_realtime/handlers/conversation.py:71, /tmp/speech-to-speech-pr241/LLM/language_model.py:424 — conversation.item.create / response.create.input now stores multipart content, including
    input_image, directly in runtime_config.chat, and LanguageModelHandler passes that chat verbatim into tokenizer.apply_chat_template(). The removed _add_to_context() path used to strip images for text-only models. As written, multimodal
    conversation items will break the local text-only HF/MLX LLM path instead of degrading gracefully.
  - Medium: /tmp/speech-to-speech-pr241/LLM/utils.py:37, /tmp/speech-to-speech-pr241/LLM/language_model.py:329, /tmp/speech-to-speech-pr241/LLM/openai_api_language_model.py:130 — resolve_auto_language() now returns a language name for plain
    codes like "en" as well as "en-auto", so both LLM handlers inject Please reply to my message in ... for normal transcriptions too. That changes behavior for explicit STT language settings and contradicts the helper’s own docstring,
    which says the extra instruction should only appear for -auto.

  Targeted tests in an isolated worktree passed: uv run pytest tests/openai_realtime/test_realtime_service.py tests/openai_realtime/test_websocket_router.py tests/test_openai_api_language_model.py tests/test_parakeet_transcription_events.py
  tests/test_qwen3_tts_handler_backend.py tests/test_tts_input_coalescing.py (154 passed). I did not find coverage for a second concurrent websocket session or for input_image reaching the local text-only LanguageModelHandler, which is  where the first two issues hide.
  
-----_

The first comment I don't mind so much, I think you mentioned this was precursor work for multi-session, so that's fine. The other two seem real

api/openai_realtime/service.py
```diff
@@ -163,20 +176,20 @@ def __init__(
...
-            "token_usage": self._on_token_usage,
-            "partial_transcription": self.conversation.on_partial_transcription,
-            "transcription_completed": self.conversation.on_transcription_completed,
+        self._pipeline_dispatch: dict[type[PipelineEvent], Callable[[str, PipelineEvent], list[ServerEvent]]] = {
+            SpeechStartedEvent: self.audio.on_speech_started,
+            SpeechStoppedEvent: self.audio.on_speech_stopped,
+            AssistantTextEvent: self.response.on_assistant_text,
+            TokenUsageEvent: self._on_token_usage,
+            PartialTranscriptionEvent: self.conversation.on_partial_transcription,
+            TranscriptionCompletedEvent: self._on_transcription_completed,
```
andimarafioti, line 185: love this :)

LLM/language_model.py
```diff
@@ -336,73 +301,69 @@ def _stream_tokens(
...
         if ctx.input_tokens or output_tokens:
-            yield (MessageTag.TOKEN_USAGE, ctx.input_tokens, output_tokens)
-        yield (MessageTag.END_OF_RESPONSE, None, None)
+            yield TokenUsage(input_tokens=ctx.input_tokens, output_tokens=output_tokens)
+        yield EndOfResponse()
 
     def on_session_end(self) -> None:
-        self.chat.reset()
-        self._last_instructions = None
-        self.tools = None
```
andimarafioti, line 419: Where is the reset happening now? You're just logging

STT/parakeet_tdt_handler.py
```diff
@@ -230,7 +223,7 @@ def process(self, spoken_prompt):
                     try:
                         progressive_text = self._show_progressive_transcription(audio_input)
                         if progressive_text:
-                            yield (MessageTag.PARTIAL, progressive_text)
+                            yield PartialTranscription(text=progressive_text)
```
andimarafioti, line 226: great!

STT/transcription_notifier.py
```diff
@@ -1,35 +1,46 @@
...
         if self.text_output_queue and text:
             self.text_output_queue.put(
-                {"type": "transcription_completed", "transcript": str(text)}
+                TranscriptionCompletedEvent(transcript=str(text), language_code=language_code)
             )
             logger.debug("Transcription completed: %s", str(text)[:80])
-        yield transcription
+
+        if not self.suppress_yield:
+            yield transcription
```
andimarafioti, line 46: Why would you want to suppress the yield?

andimarafioti (approved):
LGTM! Although I left some comments :)

LLM/language_model.py
A-Mahla: For now, they are only logged because `BaseHandler` still requires them. The next improvement will be to remove them completely from VAD and TTS side and then from the `BaseHandler` interface so we can support multi-session and move the reset logic into `ConnState`.

STT/transcription_notifier.py
A-Mahla: The STT only handles speech audio, but when realtime mode is enabled, I also need session, response, and other service state for LLM, which are sent directly to the `text_input_queue` by the service. for other than realtime mode, transcription is still yielded

PR #242, opened 2026-04-21
Add batch sentence streaming for LLM handlers

A-Mahla:
Instead of yielding sentences one at a time during streaming, accumulate multiple sentences before sending them to the TTS pipeline. This reduces the number of TTS invocations and can improve audio coherence.

A new CLI argument `--lm_stream_batch_sentences` / `--open_api_stream_batch_sentences` (default: 3) controls how many sentences are batched together. Set to 1 to restore the previous behavior.


andimarafioti (approved):
LGTM :)

PR #244, opened 2026-04-23
Preserve smart apostrophes in speechable text

andimarafioti:
## Summary
- normalize smart single and double quotes to ASCII before applying the speechable-character filter
- preserve contractions like `I’ll` and `here’s` instead of collapsing them to `Ill` / `heres`
- add a regression test covering smart apostrophes and emoji stripping

## Root Cause
`remove_unspeechable()` allowed straight ASCII apostrophes but dropped typographic apostrophes such as `’`. When the LLM emitted smart punctuation, the sanitizer removed the apostrophe before TTS received the text, which made contractions sound and read incorrectly.

## Validation
- `pytest -q tests/test_llm_utils.py`
- `ruff check LLM/utils.py tests/test_llm_utils.py`
- `ruff format --check LLM/utils.py tests/test_llm_utils.py`

### DIFF
--- a/LLM/utils.py
+++ b/LLM/utils.py
@@ -5,6 +5,15 @@ import re
 import requests
 from PIL import Image
 
+SMART_PUNCT_TRANSLATION = str.maketrans(
+    {
+        "\u2018": "'",
+        "\u2019": "'",
+        "\u201c": '"',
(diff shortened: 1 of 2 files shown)


A-Mahla (approved):
demo

PR #246, opened 2026-04-23
Add VAD start persistence

andimarafioti:
## Summary
- add a configurable `start_persistence_ms` gate before VAD opens a speech segment
- keep tentative above-threshold chunks in a pending buffer until the start-persistence threshold is met
- discard false starts cleanly so short bursts do not contaminate the speech buffer or pre-speech prefix
- cover the new behavior with VAD iterator tests and set the default start persistence to 128 ms

## Root Cause
The VAD iterator switched into a triggered speech state on the first above-threshold chunk. Short bursts of noise could therefore open a speech segment immediately, even though the segment might later be discarded by `min_speech_ms`. That still polluted the active speech buffer and could interfere with pre-speech padding behavior.

## Validation
- `pytest -q tests/test_vad_iterator.py`
- `ruff check VAD/vad_iterator.py VAD/vad_handler.py arguments_classes/vad_arguments.py tests/test_vad_iterator.py`
- `ruff format --check VAD/vad_iterator.py VAD/vad_handler.py arguments_classes/vad_arguments.py tests/test_vad_iterator.py`

### DIFF
--- a/VAD/vad_handler.py
+++ b/VAD/vad_handler.py
@@ -18,6 +18,7 @@ logger = logging.getLogger(__name__)
 # Optional import for audio enhancement
 try:
     from df.enhance import enhance, init_df
+
     HAS_DF = True
 except (ImportError, ModuleNotFoundError) as e:
     HAS_DF = False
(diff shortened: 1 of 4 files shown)


andimarafioti:
mega stale

PR #247, opened 2026-04-23
Add per-session VAD diagnostics reports

andimarafioti:
## Summary
- add optional per-session VAD diagnostics capture behind `--vad_diagnostics_dir`
- write `audio.wav`, `diagnostics.json`, and a self-contained `report.html` for each completed session
- expose per-frame iterator state so speech decisions, thresholds, and end-of-speech grace periods can be visualized after the run

## Why
VAD tuning is much easier if we can inspect the full recording alongside the raw frame-by-frame VAD state after a session. This adds a lightweight diagnostics path that makes it clear where speech started, where it stayed active, and where the iterator entered its end-of-speech grace period before closing the segment.

## Impact
- normal runs are unchanged unless `--vad_diagnostics_dir` is set
- debug runs produce a session directory with raw audio plus an HTML report for quick inspection
- the report highlights active speech, the end-of-speech grace period, thresholds, and VAD/session events

## Root Cause
The pipeline previously had no durable way to inspect the iterator's internal decisions after the fact, which made it hard to validate VAD behavior or debug false triggers.

## Validation
- `pytest -q tests/test_vad_iterator.py tests/test_vad_diagnostics.py`
- `ruff check VAD/vad_iterator.py VAD/vad_diagnostics.py VAD/vad_handler.py arguments_classes/vad_arguments.py tests/test_vad_diagnostics.py`
- `ruff format --check VAD/vad_iterator.py VAD/vad_diagnostics.py VAD/vad_handler.py arguments_classes/vad_arguments.py tests/test_vad_diagnostics.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,7 @@ See [VADHandlerArguments](./src/speech_to_speech/arguments_classes/vad_arguments
 - `--thresh`: Threshold value to trigger voice activity detection.
 - `--min_speech_ms`: Minimum duration of detected voice activity to be considered speech.
 - `--min_silence_ms`: Minimum length of silence intervals for segmenting speech, balancing sentence cutting and latency reduction.
+- `--vad_diagnostics_dir`: Save one directory per session with `audio.wav`, raw VAD frame data, and an HTML report that visualizes the waveform, speech state, and thresholds.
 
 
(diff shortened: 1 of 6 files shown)


andimarafioti:
debug stuff

PR #248, opened 2026-04-23
Release prep - typing, CI, folder layout, new uv entrypoint

A-Mahla:
## Summary

This is intended as the **last cleanup-focused PR** before the next release. The diff looks large, but **it does not change core speech-to-speech behavior**; it raises **project quality and packaging** toward a production baseline and sets us up for a PyPI-oriented layout.

### Entry point / deployment note

The app is now invoked as:

```bash
uv run speech-to-speech
```

instead of calling the interpreter under `python s2s_pipeline.py`. 

---

## What’s included

- **mypy** fixes **across the repository** so static checking is consistent with the current package layout.
- **Stronger typing** on functions (broader annotations to match real usage and catch regressions earlier).
- **Lightweight CI** running **mypy**, **ruff**, and **pytest**. **Merge-blocking checks are not enabled yet**; this is mainly to establish a baseline and iterate.
- **Reworked folder structure** to align with common Python packaging conventions and to **prepare for a future PyPI release** (clearer `src`/package boundaries and standard layout).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
+      - run: uv sync --group dev
+      - name: Ruff check
+        run: uv run ruff check src/ tests/
+      - name: Ruff format
+        run: uv run ruff format --check src/ tests/
+      - name: Mypy
+        run: uv run mypy src/
```
andimarafioti, line 31: Can we divide into two jobs: mypy and ruff? I find it easier to then see what's failing +. if ruff fails here then mypy doesn't even run.

archive/STT/moonshine_handler.py
```diff
@@ -61,9 +63,6 @@ def warmup(self):
     def process(self, vad_audio: VADAudio):
         logger.debug("infering moonshine...")
 
-        global pipeline_start
-        pipeline_start = perf_counter()
-
```
andimarafioti, line 66: Did you move the timings somewhere else?

archive/TTS/parler_handler.py
```diff
@@ -185,20 +187,22 @@ def process(self, tts_input: TTSInput | EndOfResponse):
         language_code = tts_input.language_code
         text = tts_input.text
 
-        voice = None
+        voice: str | None = None
```
andimarafioti: I would stick to Optional[str]. For consistency across the repo

archive/TTS/parler_handler.py
```diff
@@ -185,20 +187,22 @@ def process(self, tts_input: TTSInput | EndOfResponse):
...
         if response and response.audio and response.audio.output:
-            voice = response.audio.output.voice
+            voice = str(response.audio.output.voice) if response.audio.output.voice else None
```
andimarafioti:
```suggestion
            voice = str(response.audio.output.voice) if response.audio.output.voice is not None else None
```

andimarafioti (approved):
Good work! I stopped looking after 20 files or so because it looked mostly like non-functional changes. All good with general linting and such, the CI is great, I would split it into three CIs and mentioned in the comment but nothing blocking here, thank you!

archive/STT/moonshine_handler.py
A-Mahla: I kept the logic in the STT module, where pipeline_start was actually used, and also found some dead logic around the global utility variables. I’ll think about metrics and usage in a future PR and maybe reintroduce them as part of a rework logic.

PR #253, opened 2026-04-28
Refactor Chat into typed abstraction

A-Mahla:
## Summary

Rework the `Chat` class into a typed conversation-history abstraction backed by OpenAI SDK models (`ConversationItem` / `ResponseInputParam`), replacing the previous raw-dict approach. This gives us:

- Strongly-typed items (system, user, assistant, function_call, function_call_output) with validation on add.
- Automatic pairing of `tool_call` ↔ `function_call_output` by `call_id`, preventing orphaned entries that broke the API.
- Bounded-buffer eviction per user-turn instead of per-message.
- Dual serialisation (`to_response_api_chat()` / `to_transformers_chat()`) for respectively the Responses API and the Transformers API.

### DIFF
--- a/src/speech_to_speech/LLM/chat.py
+++ b/src/speech_to_speech/LLM/chat.py
@@ -1,4 +1,7 @@
+import logging
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 
 class Chat:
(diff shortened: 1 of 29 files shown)


src/speech_to_speech/api/openai_realtime/service.py
```diff
@@ -292,7 +292,7 @@ def _on_transcription_completed(self, conn_id: str, event: TranscriptionComplete
         cfg = st.runtime_config
         transcript = event.transcript
         if transcript:
-            cfg.chat.append({"role": "user", "content": transcript})
+            cfg.chat.add_item(make_user_message(transcript))
```
andimarafioti, line 295: nice

src/speech_to_speech/LLM/chat.py
```diff
@@ -1,58 +1,369 @@
...
+                            {
+                                "type": "function",
+                                "id": item.call_id,
+                                "function": {
+                                    "name": item.name,
+                                    "arguments": args,
+                                },
+                            }
+                        ],
+                    }
```
andimarafioti: I would prefer if these were method calls than created on the fly, that way if we need to change how messages are created, we can do it in one place. Something like:
results.append(_create_function_assistant_response(id, name, arguments))

src/speech_to_speech/LLM/chat.py
```diff
@@ -1,58 +1,369 @@
...
+                                break
+                        if name:
+                            break
+                result.append(
+                    {
+                        "role": "tool",
+                        "tool_call_id": item.call_id,
+                        "name": name,
+                        "content": item.output,
+                    }
```
andimarafioti: same here ofc

src/speech_to_speech/LLM/chat.py
```diff
@@ -1,58 +1,369 @@
...
+            result.append({"role": "system", "content": text})
+        for item in self.buffer:
+            if isinstance(item, RealtimeConversationItemUserMessage):
+                has_images = any(p.type == "input_image" for p in item.content)
+                if has_images:
+                    result.append(
+                        {
+                            "role": "user",
+                            "content": [p.model_dump(exclude_none=True) for p in item.content],
+                        }
```
andimarafioti: and here... you get the gist

src/speech_to_speech/LLM/chat.py
```diff
@@ -1,58 +1,369 @@
...
+    )
+
+
+def make_assistant_message(text: str) -> RealtimeConversationItemAssistantMessage:
+    return RealtimeConversationItemAssistantMessage(
+        type="message",
+        role="assistant",
+        content=[AssistantContent(type="output_text", text=text)],
+    )
+
```
andimarafioti, line 414: see, you almost have them xD

andimarafioti (changes requested):
I actually liked it and only left a nit around how you create the messages, but codex had this to say:


  - P1 src/speech_to_speech/LLM/language_model.py:412: local transformer/MLX tool calls now fail. The new RealtimeConversationItemFunctionCall is created without call_id, but Chat.add_item() rejects function calls missing one at src/
    speech_to_speech/LLM/chat.py:156. Any generated tool call will raise before the LLMResponseChunk is yielded. Copy call_id=t.call_id and likely id=t.id.
  - P2 src/speech_to_speech/s2s_pipeline.py:404: non-realtime modes lose the configured initial prompt. The pipeline creates RuntimeConfig(chat=Chat(...)) only, while the LLM handlers now read instructions solely from
    runtime_config.session.instructions and swallow init_chat_prompt via **_kwargs. This means local/websocket/socket runs no longer get the default concise voice system prompt from the CLI args.

src/speech_to_speech/LLM/chat.py
A-Mahla: This doesn’t serve the same purpose: one is normalized for the internal chat type, while the `to_...` methods are built to normalize things on the fly for a specific chat format. I strengthen the typing on the Transformers side to get a better handle on it, but the Transformers library doesn’t provide a very clean way to type chat formats.

src/speech_to_speech/LLM/chat.py
A-Mahla: see here https://github.com/huggingface/speech-to-speech/pull/253#discussion_r3155471426

andimarafioti:
Findings

  - P1 src/speech_to_speech/connections/socket_sender.py:52: default socket mode no longer re-enables listening after a response. TTS now emits AUDIO_RESPONSE_DONE on EndOfResponse, and local/websocket streamers handle that sentinel, but
    SocketSender treats any bytes as payload and has no should_listen reference. Since VAD clears should_listen at speech end, socket clients will be ignored after the first response until the 30s watchdog fires. Pass should_listen into
    SocketSender and handle AUDIO_RESPONSE_DONE like the other streamers.
  - P2 src/speech_to_speech/s2s_pipeline.py:407: non-realtime open_api runs still use the transformer LLM config for RuntimeConfig. This block always reads language_model_handler_kwargs, so open_api_chat_size and open_api_init_chat_prompt
    are ignored in local/socket/websocket mode. Branch on module_kwargs.llm == "open_api" here, like the realtime chat-size logic above.
  - P2 src/speech_to_speech/LLM/chat.py:137: conversation.item.create now mutates client-provided item IDs before echoing them back. ConversationHandler appends the same item object it later returns in conversation.item.created, but
    Chat.add_item() replaces IDs unless they match internal prefixes like msg_/fco_. A client-sent id="item_abc" becomes msg_*, and previous_item_id chains no longer reference the IDs clients supplied. Preserve existing IDs and only
    generate when absent.

andimarafioti:
Finding

  - P2 src/speech_to_speech/LLM/chat.py:46: Realtime client-supplied item IDs are rejected unless they use internal prefixes like msg_, sys_, fc_, or fco_. The OpenAI SDK model says item IDs may be provided by the client or generated by the server, without a prefix requirement. Valid conversation.item.create payloads like id="item_abc" now fail with invalid_conversation_item.

  Suggested Change
  Relax item ID validation so existing IDs are preserved and IDs are generated only when absent:

  def _ensure_id(value: str | None, prefix: str) -> str:
      return value if value else _generate_id(prefix)

  Then keep using it for item.id assignment in Chat.add_item(). If stricter handling is needed for function_call.call_id, validate that separately from item IDs because call_id is part of tool-call pairing, while id is just the conversation
  item identifier.

andimarafioti (approved):

PR #255, opened 2026-04-29
Add speculative turn revisions

andimarafioti:
## Summary

Adds a raw-audio-first speculative turn path for realtime mode. VAD now soft-ends turns, keeps listening, and can reopen a recent turn by emitting a newer revision that contains the combined raw audio. Downstream stages carry turn/revision metadata so stale STT, LLM, and TTS work can be ignored instead of leaking into the active response.

## Details

- Adds a shared `SpeculativeTurnTracker` for latest/committed turn revisions.
- Threads `turn_id` and `turn_revision` through VAD audio, STT transcriptions, pipeline events, LLM requests/chunks, and TTS inputs.
- Replaces the previous speculative user transcript in realtime chat when a newer raw-audio revision lands.
- Stages assistant/tool chat writes in LLM handlers and only commits them if the revision is still current.
- Drops stale LLM/TTS output before it reaches clients or audio synthesis.
- Adds `--speculative_reopen_ms` to configure the realtime reopen window.

## Validation

- `uv run ruff check src/speech_to_speech tests/openai_realtime/test_realtime_service.py`
- `uv run pytest`

`uv run pytest` passed with 357 tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,6 @@ with client.beta.realtime.connect(model="model_name") as conn:
 | `response.created` | Emitted on the first outbound audio chunk (response is `in_progress`). |
 | `response.output_audio.delta` | Base64 PCM audio chunk from TTS. |
 | `response.output_audio.done` | Audio stream complete for the current output item. |
-| `output_audio_buffer.cleared` | Server-side VAD interruption cleared already-buffered assistant audio. |
 | `response.output_audio_transcript.done` | Full assistant text transcript for the turn. |
 | `response.function_call_arguments.done` | Tool call with `call_id`, `name`, and JSON `arguments`. |
 | `response.done` | Response finished (`completed`, `cancelled` with reason `turn_detected` or `client_cancelled`). |
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ with client.beta.realtime.connect(model="model_name") as conn:
 | `response.created` | Emitted on the first outbound audio chunk (response is `in_progress`). |
 | `response.output_audio.delta` | Base64 PCM audio chunk from TTS. |
 | `response.output_audio.done` | Audio stream complete for the current output item. |
+| `output_audio_buffer.cleared` | Server-side VAD interruption cleared already-buffered assistant audio. |
 | `response.output_audio_transcript.done` | Full assistant text transcript for the turn. |
 | `response.function_call_arguments.done` | Tool call with `call_id`, `name`, and JSON `arguments`. |
 | `response.done` | Response finished (`completed`, `cancelled` with reason `turn_detected` or `client_cancelled`). |
--- a/archive/TTS/melo_handler.py
+++ b/archive/TTS/melo_handler.py
@@ -14,6 +14,7 @@ from speech_to_speech.baseHandler import BaseHandler
 from speech_to_speech.pipeline.cancel_scope import CancelScope
 from speech_to_speech.pipeline.handler_types import TTSIn, TTSOut
 from speech_to_speech.pipeline.messages import AUDIO_RESPONSE_DONE, EndOfResponse
+from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
 
 logger = logging.getLogger(__name__)
 
@@ -48,9 +49,11 @@ class MeloTTSHandler(BaseHandler[TTSIn, TTSOut]):
         gen_kwargs: dict[str, Any] = {},  # Unused
         blocksize: int = 512,
         cancel_scope: CancelScope | None = None,
+        speculative_turns: SpeculativeTurnTracker | None = None,
     ) -> None:
         self.should_listen = should_listen
         self.cancel_scope = cancel_scope
+        self.speculative_turns = speculative_turns
         self.device = device
         self.language = language
         self.model = TTS(language=WHISPER_LANGUAGE_TO_MELO_LANGUAGE[self.language], device=device)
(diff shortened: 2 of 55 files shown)


andimarafioti:
Claude's review:

 Overview                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                 
  This is a substantial feature that adds a speculative turn mechanism for the realtime mode. When VAD detects a silence gap, instead of immediately committing the turn and triggering a full STT→LLM→TTS cycle, it "soft-ends" the turn and    
  keeps listening. If the user resumes speaking within a configurable window (--speculative_reopen_ms, default 1200ms), the turn is reopened with a new revision containing the combined raw audio. Downstream stages carry turn_id/turn_revision
   metadata so stale work is dropped.                             

  What's Good

  1. Clean central abstraction — SpeculativeTurnTracker is well-designed with clear semantics: observe, is_latest, commit, begin_reopen_candidate, confirm_reopen_candidate, cancel_reopen_candidate. The pending-reopen lifecycle is explicit
  and well-guarded.
  2. Thread safety — Uses a single Condition consistently. The wait_for_pending_reopen with deadline-based timeout is solid. No busy-spinning.
  3. Metadata threading — turn_id/turn_revision are propagated through every pipeline message type (VADAudio, Transcription, LLMResponseChunk, TTSInput, EndOfResponse, all events). This is thorough.
  4. Staged chat writes — LLM handlers accumulate pending_chat_items and only commit to the original chat if the turn is still current. This prevents ghost assistant messages in the conversation history.
  5. Good test coverage — Tests for SpeculativeTurnTracker, LMOutputProcessor blocking behavior, RealtimeService speculative turn flow. The threading tests (test_pending_reopen_holds_assistant_chunk_until_cancelled,
  test_confirmed_reopen_drops_held_assistant_chunk) are particularly good.
  6. Backwards compatible — All speculative_turns params default to None, turn_id/turn_revision default to None. When disabled, everything passes through unchanged.

  Issues & Concerns

  Potential Bugs

  1. _ensure_turn_for_speech_start has two reopen paths (vad_handler.py:229-253) — There's a non-tracker path at line 241-243 that increments _current_turn_revision directly without going through SpeculativeTurnTracker.observe() ... except
  it does call observe on line 252. But this direct increment path doesn't use the begin_reopen_candidate/confirm_reopen_candidate protocol, so downstream stages won't be blocked by wait_for_pending_reopen. This seems intentional (fast path
  when no pending reopen was started earlier), but the two reopen mechanisms could lead to subtle differences in behavior.
  2. Memory leak in _latest_revision / _committed_revision dicts (speculative_turns.py:24-26) — These grow unboundedly as turns accumulate. Each turn_id string is kept forever. For long-running sessions, this could add up. Consider pruning
  old turns when a new one starts, or using an LRU-style eviction.
  3. _speculative_segments grows unboundedly (vad_handler.py:420) — Each soft-ended segment is appended to _speculative_segments and concatenated into subsequent outputs via _combined_turn_audio. If a user has many short pauses within the
  reopen window, segments accumulate. They're only cleared on _start_new_turn (line 155). For a very long utterance with many pauses, this could cause large memory allocations and slow numpy concatenation.

  Design Questions

  4. getattr(self, "speculative_turns", None) pattern — Used in responses_api_language_model.py:92, lm_output_processor.py:49, and TTS handlers. Since setup() always sets the attribute, the getattr fallback shouldn't be needed. This seems
  like defensive coding against calling process() before setup(), but in that case you'd have bigger problems. Minor, but a plain self.speculative_turns would be clearer.
  5. _turn_output_allowed called multiple times per response — In responses_api_language_model.py, _turn_output_allowed (which calls wait_for_pending_reopen) is called:
    - Once per sentence batch (line 211)
    - Once for remaining text (line 265)
    - Once for ReadTimeout fallback (line 342)
    - Once for chat commit (line 357)

  Each call can block for up to 2s. In the worst case a single response could block 4× the timeout. This is probably fine in practice (the pending reopen resolves quickly), but worth noting.
  6. No should_listen.clear() in realtime mode — In _process_realtime, after a soft-end, should_listen is never cleared (unlike _process_normal which calls self.should_listen.clear() at line 451). This is clearly intentional (keep listening
  for reopens), but it means the pipeline never stops accepting audio in realtime mode. The flow control depends entirely on the speculative turn tracker to discard stale work downstream.

  Minor

  7. assert in production code (vad_handler.py:250) — assert self._current_turn_id is not None and self._current_turn_revision is not None — This will be optimized away with -O. If this invariant matters, use an explicit check.
  8. on_session_end in VAD resets speculative_turns (vad_handler.py:497-498) — But RealtimeService.register() also calls speculative_turns.reset() (service.py:207-208). Double-reset isn't harmful but adds confusion about ownership.

  Verdict

  This is a well-thought-out feature with clean abstractions and good test coverage. The core SpeculativeTurnTracker is solid. The main concerns are the unbounded memory growth (#2, #3) for long sessions, and the dual reopen path in VAD (#1)
   which could be simplified. The rest are minor style/robustness points.

andimarafioti:
Review from Codex:
 Findings

  - High: queued assistant/TTS output can still leak during a pending reopen window. /tmp/s2s-pr255-review/src/speech_to_speech/api/openai_realtime/service.py:289 only checks is_latest(), and /tmp/s2s-pr255-review/src/speech_to_speech/api/
    openai_realtime/handlers/response.py:219 still emits assistant events even when commit() is deferred by a pending reopen. Same issue exists at the TTS boundary, e.g. /tmp/s2s-pr255-review/src/speech_to_speech/TTS/
    qwen3_tts_handler.py:640, where stale gating also only checks is_latest(). Repro: if LMOutputProcessor queues an AssistantTextEvent for turn_1/rev0, then VAD starts a reopen candidate before the send loop consumes that event, the
    service still emits response.output_audio_transcript.done while the reopen is pending. If the candidate later confirms, stale assistant text/audio may already have reached the client, and stale assistant chat items may remain. The
    dispatch/TTS paths should wait for pending reopen and re-check latest, or commit() should return whether emission is allowed.

  Validation

  - uv run --python 3.12 pytest tests/test_speculative_turns.py tests/test_lm_output_processor.py tests/openai_realtime/test_realtime_service.py -q -> 120 passed
  - uv run --python 3.12 pytest -q -> 399 passed
  - uv run --python 3.12 ruff check ... -> passed

  I did not find other blocking issues in the revised turn replacement, STT metadata propagation, tracker pruning, or chat staging paths.

PR #257, opened 2026-04-30
Fix Parakeet progressive empty windows

andimarafioti:
## Summary
- Reset Parakeet live transcription streaming state after every final transcription.
- Add regression coverage for the final transcription reset path.
- Add a representative turn-to-turn regression test: a previous long live turn has fixed sentence timing, final STT completes, and the next short progressive chunk must decode from the start instead of being sliced to an empty window.

## Root Cause
The failure was not caused by VAD sending empty audio. The logs show VAD was silent for many seconds, then yielded real progressive audio for the next utterance (`532ms`, then `852ms`). The empty `[1, 1, 0]` tensor was created later in Parakeet progressive STT windowing.

`SmartProgressiveStreamingHandler` keeps per-utterance progressive state, including `fixed_sentences` and `fixed_end_time`. That state only becomes non-zero after a long enough live transcription window fixes completed sentences, so most short turns are unaffected. This is why the bug was intermittent rather than constant.

Before this change, the nano-parakeet final transcription path cleared `processing_final` but did not reset `streaming_handler`. After a previous long/progressive utterance, stale `fixed_end_time` could survive across the silence gap. On the next turn, the first progressive chunk might be only a few hundred milliseconds, so slicing from the stale fixed boundary could produce an empty window and trigger the model error.

This patch resets the progressive handler at the utterance boundary after final STT completes. It does not clear fixed sentences during an in-progress utterance.

## Validation
- `python3 -m py_compile src/speech_to_speech/STT/parakeet_tdt_handler.py src/speech_to_speech/STT/smart_progressive_streaming.py tests/test_parakeet_transcription_events.py`
- `git diff --check main...HEAD`
- `PYTHONPATH=src uv run --no-project --with pytest --with numpy --with rich --with openai==2.28.0 --with pydantic pytest -q tests/test_parakeet_transcription_events.py` (`6 passed, 1 warning`)

### DIFF
--- a/src/speech_to_speech/STT/parakeet_tdt_handler.py
+++ b/src/speech_to_speech/STT/parakeet_tdt_handler.py
@@ -292,11 +292,12 @@ class ParakeetTDTSTTHandler(BaseHandler[STTIn, STTOut]):
             if language_code:
                 console.print(f"[dim]Language: {language_code}[/dim]")
 
-        yield Transcription(text=pred_text, language_code=language_code)
-
-        # Reset processing_final flag for next utterance
         if self.enable_live_transcription:
             self.processing_final = False
+            if self.streaming_handler is not None:
+                self.streaming_handler.reset()
+
+        yield Transcription(text=pred_text, language_code=language_code)
 
     def _detect_language_from_text(self, text: str) -> Optional[str]:
         """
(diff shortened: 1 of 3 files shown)


A-Mahla (approved):
demo

PR #259, opened 2026-04-30
Add Parakeet nano timing diagnostics

andimarafioti:
## Summary
- add debug timing probes around non-MLX Parakeet final transcription
- log nano transcribe time separately from language detection and total final processing
- add progressive timing and compute-lock wait diagnostics to correlate final outliers with live updates

## Validation
- `python3 -m py_compile src/speech_to_speech/STT/parakeet_tdt_handler.py`
- `./.venv/bin/python -m pytest tests/test_parakeet_transcription_events.py tests/test_parakeet_language_detection.py`
- `git diff --check`


andimarafioti:
This was good for debugging an issue but not needed on main.

PR #269, opened 2026-05-04
Refactor LLM argument classes and unify CLI flags

A-Mahla:
## Summary
- Extract shared LLM arguments (`model_name`, `init_chat_prompt`, `chat_size`, `stream_batch_sentences`, `enable_lang_prompt`) into a common `LanguageModelBaseArguments` base class, eliminating duplication between local and OpenAI-API backends.
- Rename CLI flags for consistency: `--llm` → `--llm_backend`, `--lm_*` → `--llm_*`, remove `open_api_` prefix from shared fields.
- Add `Literal` type hints to module arguments (`mode`, `stt`, `llm_backend`, `tts`) for validation and editor support.
- Use a pre-parse step to conditionally register only the relevant LLM argument class, avoiding field-name collisions in `HfArgumentParser`.
- Expand README with LLM backend documentation (OpenAI, HF Inference, OpenRouter, vLLM, llama.cpp), Realtime mode usage, and updated CLI examples.


andimarafioti (approved):
Let's go!!!

PR #270, opened 2026-05-04
Limit OpenAI tool calls per response

andimarafioti:
## Summary
- pass `parallel_tool_calls=False` and `max_tool_calls=1` for official OpenAI Responses API requests when tools are enabled
- keep those parameters off for custom OpenAI-compatible base URLs for compatibility
- add unit coverage for both paths

## Context
Recent realtime logs showed the model returning multiple tool calls in a single response (`play_emotion`, `stop_dance`, `stop_emotion`) even though the voice prompt asks for one tool call. This enforces that request at the API parameter level for the OpenAI path.

This does not replace the app-side fix for repeated follow-up `response.create` events after tool outputs; it only prevents one model response from producing multiple tool calls at once.

## Tests
- `uv run pytest tests/test_openai_api_language_model.py tests/openai_realtime/test_realtime_service.py`
- `uv run ruff check src/speech_to_speech/LLM/openai_api_language_model.py tests/test_openai_api_language_model.py`

### DIFF
--- a/src/speech_to_speech/LLM/openai_api_language_model.py
+++ b/src/speech_to_speech/LLM/openai_api_language_model.py
@@ -76,6 +76,7 @@ class OpenApiModelHandler(BaseHandler[LLMIn, LLMOut]):
 
         self.user_role = user_role
         self.client = OpenAI(api_key=api_key, base_url=base_url)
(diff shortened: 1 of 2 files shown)


andimarafioti:
dont like this

PR #271, opened 2026-05-04
Tighten tool-use prompting

andimarafioti:
## Summary

- Make speech the default in the voice prompt and narrow tool use to explicit or necessary actions.
- Discourage movement/emotion/stop tools for routine acknowledgments, greetings, listening, or turn-ending phrases.
- Update the local tool-call prompt to forbid multiple tool calls in a single response.
- Add focused prompt tests.

## Why

The model could treat expressive robot tools as conversational punctuation, producing tool-only responses for ordinary turns like acknowledgments or stopping phrases. The prompt now makes that behavior explicitly undesirable while keeping idle-action turns available.

## Validation

- `uv run pytest tests/test_voice_prompt.py tests/test_openai_api_language_model.py`
- `uv run pytest tests/tool_call tests/test_voice_prompt.py`
- `uv run ruff check src/speech_to_speech/LLM/voice_prompt.py src/speech_to_speech/LLM/tool_call/tool_prompt.py tests/test_voice_prompt.py`
- `uv run ruff format --check src/speech_to_speech/LLM/voice_prompt.py src/speech_to_speech/LLM/tool_call/tool_prompt.py tests/test_voice_prompt.py`

### DIFF
--- a/src/speech_to_speech/LLM/tool_call/tool_prompt.py
+++ b/src/speech_to_speech/LLM/tool_call/tool_prompt.py
@@ -39,7 +39,6 @@ When you need to use a tool, output the function call between \
 
 {{ enter_code }}
 function_name(arg_name_1=value1, arg_name_2='string_value')
-another_function(arg_name_1='hello')
 {{ end_code }}
 
 RULES:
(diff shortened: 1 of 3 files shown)


A-Mahla (approved):
LGTM Thanks!

PR #276, opened 2026-05-05
Fix Qwen3 voice override validation

andimarafioti:
## Summary

- Validate Realtime voice overrides against Qwen3 CustomVoice supported speakers before applying them.
- Ignore unsupported OpenAI voice names such as `cedar` so Qwen3 generation keeps using the configured speaker.
- Add regression tests for unsupported and supported CustomVoice overrides.

## Root Cause

The Qwen3 CustomVoice handler treated any Realtime `audio.output.voice` value as a Qwen speaker. OpenAI Realtime voice names like `cedar` are valid protocol values but are not valid Qwen3 speakers, so they reached `faster-qwen3-tts` and caused an unsupported speaker error.

## Validation

- `pytest tests/test_qwen3_tts_handler_backend.py`
- `ruff check src/speech_to_speech/TTS/qwen3_tts_handler.py tests/test_qwen3_tts_handler_backend.py`

### DIFF
--- a/src/speech_to_speech/TTS/qwen3_tts_handler.py
+++ b/src/speech_to_speech/TTS/qwen3_tts_handler.py
@@ -374,7 +374,19 @@ class Qwen3TTSHandler(BaseHandler[TTSIn, TTSOut]):
             return
 
         if model_type == "custom_voice":
-            self.speaker = session_voice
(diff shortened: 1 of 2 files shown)


A-Mahla (approved):
LGTM!
demo

PR #281, opened 2026-05-18
Chat compression: background history compaction

A-Mahla:
## Summary
- Adds optional **background chat compaction**: when the user-turn count exceeds `Chat.size`, older turns are summarized into a single user/assistant pair by an LLM running in a daemon thread, instead of being evicted.
- Splits size enforcement out of `Chat.add_item` into a new explicit `Chat.trim_if_needed(compactor)` call after each generation. With `compactor=None`, behavior is unchanged (synchronous oldest-turn eviction).
- New module `LLM/compaction_prompt.py` with the compaction system prompt, the `CompactGenerateFn` interface, and `build_compactor` factory that returns a `CompactFn` compatible with `Chat`.
- `Chat` is now thread-safe (single `_lock`, single-flight compaction gate, `_gen_counter` so `reset()` / `close()` cancel any in-flight splice). Pending function calls whose outputs land after the snapshot are preserved across compaction.
- New `compact_history: bool = False` flag on `BaseLanguageModelHandler` and `ResponsesApiModelHandler`. Each backend (mlx, transformers, VLM, Responses API) implements `_build_compaction_generate_fn` for its own model.
- Tests cover concurrency, single-flight gating, reset/close cancellation, pending-FC preservation, and the Responses API path.


andimarafioti:
Codex:

Findings

P1 language_model.py (line 582): transformers compaction always crashes because return_full_text=False is passed twice. self.gen_kwargs includes return_full_text from setup, but the compactor only filters out streamer, stopping_criteria, and max_new_tokens. I reproduced this with a stub: TypeError ... got multiple values for keyword argument 'return_full_text'. Since the worker catches the exception and leaves the chat unchanged, history stops being trimmed when compaction is enabled.

P1 language_model_base_arguments.py (line 40): compact_history defaults to True in the CLI/dataclass layer, even though both handlers default it to False and the PR summary says the flag is false by default. Because these args are passed via vars(...), normal CLI users will enable background compaction, extra model/API calls, and the transformer failure above without opting in.

P2 language_model.py (line 739): the MLX VLM compactor returns mlx_vlm_stream_generate(...) directly. The same helper is consumed as a token iterator in normal generation, but build_compactor expects a JSON string. That path will fail every compaction attempt for MLX VLM and leave history untrimmed.

P2 language_model.py (line 582): transformer/VLM transformer compaction runs in a background thread with no serialization around the shared model/pipeline. A new user turn can start normal generation while compaction is still generating, so the same model can be used concurrently. MLX has MLXLockContext; transformers needs an equivalent guard or a non-overlapping scheduling strategy.

Verification

PYTHONPATH=src python -m pytest tests/test_chat.py tests/test_responses_api_language_model.py -q passed.
PYTHONPATH=src python -m pytest tests/openai_realtime/test_realtime_service.py -q passed.
PYTHONPATH=src python -m ruff check ... passed on the changed files.
Reproduced the transformers compactor duplicate-keyword crash with a minimal stub.

A-Mahla:
I set the compact_history CLI arg to True by default because I believe it’s now a built-in feature in all agentic flows, especially conversational ones.

If users want to disable it, they should add `--compact_history False`.

andimarafioti (approved):

PR #282, opened 2026-05-20
Add --num_pipelines pool for concurrent realtime sessions

A-Mahla:
## Summary

- New `--num_pipelines N` flag (realtime mode only): one uvicorn on `--ws_port` routes each incoming websocket to one of N pre-built pipelines (own queues, events, `RealtimeService`, VAD/STT/LM/TTS handlers). Max concurrent sessions = N; further connections rejected with `1008 session_limit_reached`.
- Per-websocket state (`websocket`, `session_id`, send-loop scratch) lives on a `SessionState` attached to a `PipelineUnit` during a claim, so nothing outlives its websocket.
- `[pipeline N]` log prefixing via a `ContextVar` set in handler threads and the websocket route/send-loop tasks.
- On Apple Silicon, live transcription auto-disables when `num_pipelines > 1` to avoid MLX-lock contention warnings (final transcripts unaffected).
- Pool rejection uses a stateless `build_error_event` helper — no booking against a random unit's metrics.
- `GET /v1/pool` introspection endpoint with per-unit `state ∈ {idle, active, draining}` and `draining_for_s`.
- `scripts/synthetic_conversation_realtime_client.py` drives N parallel synthetic websockets from a WAV file.

`num_pipelines` defaults to 1, preserving current single-session behavior.

## Test plan

- [ ] `python -m speech_to_speech.s2s_pipeline --mode realtime --num_pipelines 2` boots one uvicorn on 8765
- [ ] `python scripts/synthetic_conversation_realtime_client.py --audio prompt.wav --n 2` → both clients receive distinct responses
- [ ] Same with `--n 3` → third gets `session_limit_reached`
- [ ] `curl http://127.0.0.1:8765/v1/pool` reflects live slot usage
- [ ] `--num_pipelines 1` behavior matches main
- [ ] `--num_pipelines 2 --mode local` errors at startup

## Follow-ups (not in this PR)

- Audit `on_session_end` on non-default STT/TTS handlers (whisper variants, chatTTS, pocket)
- Multi-uvicorn-worker safety: add a lock around `_claim_unit` if ever running uvicorn with workers > 1
- Deployment-layer node restart for pipelines stuck draining >10min — tracked in [andimarafioti/s2s-endpoint#25](https://github.com/andimarafioti/s2s-endpoint/issues/25)


A-Mahla:
## Design note: timeless SESSION_END drain

The release path `_release_unit_after_drain()` in `src/speech_to_speech/api/openai_realtime/websocket_router.py` intentionally has **no timeout-fallback release**. If `SESSION_END` doesn't propagate through the handler chain back to `output_queue`, the unit stays unavailable indefinitely, by design.

**Rationale:** a stuck `SESSION_END` means one of the handlers (STT, LM, or TTS) is hung on in-flight work. Releasing the unit anyway would let a new client claim it while stale output from the previous session is still being produced and dispatched that's a cross-session leak. We accept reduced pool capacity over data leakage between users.

When this happens, `GET /v1/pool` surfaces the affected unit with `state: "draining"` and `draining_for_s` showing how long it's been stuck. A warning is logged after `SESSION_END_DRAIN_TIMEOUT_S` (default 10s) for ops visibility.

**Deployment-layer follow-up:** if a pipeline has been stuck draining for more than 10 minutes, the underlying compute is broken and the node should be restarted. Tracked in [andimarafioti/s2s-endpoint#25](https://github.com/andimarafioti/s2s-endpoint/issues/25).

src/speech_to_speech/api/openai_realtime/websocket_router.py
```diff
@@ -55,284 +61,390 @@ def _keep_user_text_event(item: Any) -> bool:
...
+
+    The input queue is also drained so pending audio from a released session
+    cannot be processed by the handlers (and thus reach the next session that
+    claims this unit). SESSION_END is enqueued by the route handler *after*
+    this returns to serve as the soft reset signal for stateful handlers.
+    """
+    unit.cancel_scope.cancel()
+    _flush_queue(unit.input_queue)
+    _flush_queue(unit.output_queue, preserve=preserve)
+    _flush_queue(unit.text_output_queue, preserve=preserve)
```
andimarafioti, line 117: Why aren't you flushing the text_prompt_queue?

src/speech_to_speech/s2s_pipeline.py
```diff
@@ -747,6 +935,27 @@ def main() -> None:
 
     setup_logger(args.module_kwargs.log_level)
 
+    if args.module_kwargs.num_servers < 1:
+        raise ValueError(f"--num_servers must be >= 1, got {args.module_kwargs.num_servers}")
+    if args.module_kwargs.num_servers > 1 and args.module_kwargs.mode != "realtime":
```
andimarafioti: this validation runs before prepare_all_args(), but prepare_all_args() can mutate mode via local_mac_optimal_settings to "local". So --local_mac_optimal_settings --num_servers 2 bypasses the realtime-only guard and then ignores the pool size in local mode.

src/speech_to_speech/s2s_pipeline.py
```diff
@@ -747,6 +935,27 @@ def main() -> None:
...
+    # a flood of warnings without affecting final transcripts. With a pool, pre-emptively turn
+    # it off so logs stay readable; the final STT path is unaffected. Non-darwin platforms
+    # don't share this lock, so leave their live transcription alone.
+    if args.module_kwargs.num_servers > 1 and platform == "darwin" and args.module_kwargs.enable_live_transcription:
+        logger.info(
+            "MLX contention: --num_servers=%d > 1 on Apple Silicon → disabling live transcription "
+            "(progressive STT contends on the global MLX lock)",
+            args.module_kwargs.num_servers,
+        )
+        args.module_kwargs.enable_live_transcription = False
```
andimarafioti, line 977: Interesting, I also had issues with this in my work for the speculative VAD.

andimarafioti (approved):
This looks great, very elegant solution 🚀  I left a couple of comments, just the text_prompt_queue seems a bit larger, if it's intended, please add a comment for future reference

src/speech_to_speech/api/openai_realtime/websocket_router.py
A-Mahla: good catch, I flush now also the `text_prompt_queue` one

PR #288, opened 2026-05-26
Add Qwen3 TTS audio energy logging

andimarafioti:
## Summary

Adds audio-energy diagnostics to Qwen3-TTS streaming so we can tell whether generated chunks contain speech-like signal when macOS playback appears to stop after an early sentence.

## Details

- Logs per-source-chunk RMS, peak, and active-sample percentage at debug level.
- Adds the same aggregate energy metrics to the existing Qwen3-TTS generated-audio info log.
- Keeps the existing TTS input coalescing behavior unchanged.
- Adds a focused unit test for the energy-stat helper.

## Validation

- `uv run pytest tests/test_tts_input_coalescing.py tests/test_qwen3_tts_handler_backend.py`
- `uv run ruff check src/speech_to_speech/TTS/qwen3_tts_handler.py tests/test_qwen3_tts_handler_backend.py tests/test_tts_input_coalescing.py`


andimarafioti:
Closing in favor of #289, which uses a branch on the main repository.

PR #291, opened 2026-05-29
Fix MLX-VLM crash in the multi-threaded pipeline

sxy-trans-n:
## Problem

Running a VLM through the MLX-LM backend (`--llm_is_vlm true`) crashes on
the first turn:

```
speech-to-speech
--mode realtime
--stt parakeet-tdt
--llm_backend mlx-lm
--tts qwen3 --qwen3_tts_mlx_quantization 6bit
--model_name "mlx-community/Qwen3.5-4B-4bit"
--llm_is_vlm true

RuntimeError: There is no Stream(gpu, 1) in current thread.
File ".../mlx_vlm/generate.py", line 1322, in generate_step
mx.eval([c.state for c in prompt_cache])
```

## Cause
MLX binds both **streams** and **model arrays** to the thread that created
them. `mlx_vlm.generate` also creates a thread-local `generation_stream` at
import time (on the main thread). But `VisionLanguageModelHandler` loads the
model in `setup()` (main thread) while `_generate()` runs in a pipeline
**worker thread** — so the stream/arrays belong to the wrong thread and
`mx.eval` / `mx.synchronize` fail.
## Fix
- Defer the MLX VLM model load + stream creation to the worker thread.
  Done eagerly in `run()` (before the message loop) to avoid first-request
  lag, with an idempotent guard so `_generate()` is safe too.
- Replace `mlx_vlm.generate`'s thread-local stream with a plain stream on
  that worker thread. Safe because `MLXLockContext` already serializes all
  MLX access across handlers.
- Pass `enable_thinking` to the VLM chat template. The text-only path
  already did this; the VLM path didn't, so replies leaked `<think>` blocks.
## Testing
Verified locally on an M3 Max (`--mode local`, same flags): the VLM model
loads on the worker thread, multi-turn conversation works, and replies no
longer contain `<think>` blocks.

### DIFF
--- a/src/speech_to_speech/LLM/language_model.py
+++ b/src/speech_to_speech/LLM/language_model.py
@@ -4,7 +4,7 @@ import logging
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Iterator, Sized
 from queue import Empty
-from threading import Lock, Thread
+from threading import Lock, Thread, local
 from typing import Any, Literal, Optional, Protocol, runtime_checkable
 
 import torch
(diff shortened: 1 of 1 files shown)


andimarafioti (commented):
Findings

[P2] MLX VLM still breaks with --num_pipelines > 1: _ensure_mlx_vlm_loaded_on_worker_thread() writes a process-global mlx_vlm.generate.generation_stream, then returns early forever once that handler is loaded. With multiple realtime pipeline units, whichever handler loads last overwrites the global stream, so another handler can later call stream_generate from its own worker thread using the wrong thread’s stream. See [language_model.py:683-689](https://github.com/huggingface/speech-to-speech/blob/345bd29ccdcdd7a4ce38254406672d2c5541f891/src/speech_to_speech/LLM/language_model.py#L683-L689) and the pool construction at [s2s_pipeline.py:623-656](https://github.com/huggingface/speech-to-speech/blob/345bd29ccdcdd7a4ce38254406672d2c5541f891/src/speech_to_speech/s2s_pipeline.py#L623-L656). The stream needs to be installed per generation, under the MLX lock, or otherwise made per handler/thread.

[P2] Default chat compaction captures unloaded None model objects: _load_model() now sets self.model, self.processor, and self.tokenizer to None, but BaseLanguageModelHandler.setup() builds the compactor immediately, and _build_compaction_generate_fn() captures those values into locals. Since compact_history defaults to true, long MLX VLM conversations will fail compaction with NoneType access once the chat exceeds chat_size. See [language_model.py:173](https://github.com/huggingface/speech-to-speech/blob/345bd29ccdcdd7a4ce38254406672d2c5541f891/src/speech_to_speech/LLM/language_model.py#L173), [language_model.py:708-710](https://github.com/huggingface/speech-to-speech/blob/345bd29ccdcdd7a4ce38254406672d2c5541f891/src/speech_to_speech/LLM/language_model.py#L708-L710), and [language_model.py:787-799](https://github.com/huggingface/speech-to-speech/blob/345bd29ccdcdd7a4ce38254406672d2c5541f891/src/speech_to_speech/LLM/language_model.py#L787-L799).

sxy-trans-n:
Thanks — both issues confirmed and fixed in the latest push.

- `--num_pipelines > 1`: stream now lives on a `threading.local()` and
  is re-pointed under the MLX lock per call, not set-once.
- Compaction `None` capture: the closure reads `self.model` /
  `self.processor` lazily; by the time compaction fires, the first
  generation has already populated them.

Also moved the load into `run()` so the first turn doesn't pay it.

Smoke-tested with `--mode local` + `Qwen3.5-4B-VLM` — model loads on the
worker thread, multi-turn works.

andimarafioti:
Sorry, this became a bit stale. My fault. There was some other work that might have addressed this. Feel free to reopen if you think it's necessary/good.

PR #298, opened 2026-06-04
Add audio-input LLM STT bypass

andimarafioti:
## Summary

Adds an STT-bypass mode for audio-input LLMs through the OpenAI-compatible Chat Completions backend:

```bash
--stt none --llm_backend chat-completions --model_name YOUR_AUDIO_CAPABLE_MODEL
```

An explicitly selected audio-capable model is required. The default `gpt-5.4-mini` accepts text and image input but does not accept audio. Direct audio mode rejects other LLM backends before handler construction and warmup, avoiding mixed `/v1/responses` and `/v1/chat/completions` model compatibility failures.

The Chat Completions backend stays on `/v1/chat/completions` for warmup, ordinary text/tool turns, and audio turns, so providers do not need to implement `/v1/responses`.

## Details

- Adds `AudioInputNotifier` to route final VAD audio into the Chat Completions LLM.
- Requires `--llm_backend chat-completions` when `--stt none` is selected, with an early, actionable validation error.
- Routes realtime audio completion through `RealtimeService` before queuing generation, so the request is marked pending and can be cancelled by barge-in before model output begins.
- Announces pending implicit responses with `response.created` before emitting the top-level error and response-specific terminal events, then releases the response slot.
- Records direct input-audio duration before response item startup resets transient state, preserving per-response and global usage metrics.
- Moves audio-turn orchestration into the shared OpenAI-compatible handler, with Chat Completions serialization, request, tool, refusal, streaming, and usage hooks.
- Encodes VAD audio to WAV bytes in memory before base64 serialization.
- Supports both common OpenAI-compatible audio content representations:
  - `--responses_api_audio_content_type input_audio` (default)
  - `--responses_api_audio_content_type audio_url` (base64 WAV data URL)
- Retains a bounded number of completed audio turns with `--responses_api_audio_history_turns` (default `1`); older audio is replaced with `[User audio input]` while preserving role order.
- Treats in-band audio history transactionally: provisional user, assistant, function-call, and fast function-output state is rolled back after provider failures, cancellations, or an interrupted generator.
- Reuses Chat Completions tool conversion and event parsing for audio requests, including refusals and streaming usage chunks.
- Preserves environment-provided credentials and injects a dummy key only for unauthenticated loopback endpoints.
- Documents the backend and audio-model requirements with links to the official OpenAI model and audio-input documentation.

## Validation

- `ruff check src/ tests/`
- `mypy src/`
- `pytest -q` → `752 passed, 1 skipped`


kroaton:
Why is this not merged? Gemma 12B seems like a no-brainer for this app.

andimarafioti:
It's kind of a risky PR. I first impleemnted this when Gemma 12B came out and it wasn't working well so I didn't merge it. Now it should be working well, the demo last week with Inkling used this. I'm trying to merge it ASAP

PR #299, opened 2026-06-05
Fix Paraformer progressive transcription events

kamjin3086:
## Summary

Hi, and thank you for maintaining `speech-to-speech`.

This PR fixes Paraformer realtime/live transcription so progressive VAD chunks are emitted as `PartialTranscription` messages instead of final `Transcription` messages.

With the current behavior, a progressive Paraformer update can be treated as a completed user turn by downstream components. In a Chinese realtime setup this causes repeated LLM turns such as partial prefixes being sent as independent user messages before the final transcript arrives.

## Changes

- Return `PartialTranscription` when `VADAudio.mode == "progressive"`.
- Keep returning `Transcription` for final or legacy/non-progressive inputs.
- Preserve turn metadata on both partial and final outputs.
- Add focused Paraformer tests using a fake model, so the test does not require loading FunASR weights.

## Validation

Ran locally:

```bash
PYTHONPATH=src python3 -m pytest \
  tests/test_paraformer_transcription_events.py \
  tests/test_transcription_notifier.py -q
```

Result:

```text
7 passed, 2 warnings
```

The warnings were unrelated local environment warnings (`asyncio_mode` config without `pytest-asyncio`, and a `pydub` `audioop` deprecation warning).

### DIFF
--- a/src/speech_to_speech/STT/paraformer_handler.py
+++ b/src/speech_to_speech/STT/paraformer_handler.py
@@ -9,7 +9,7 @@ from funasr import AutoModel
 from rich.console import Console
 
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
-from speech_to_speech.pipeline.messages import Transcription
+from speech_to_speech.pipeline.messages import PartialTranscription, Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
 
 logging.basicConfig(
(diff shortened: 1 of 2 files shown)


andimarafioti (changes requested):
Please fix the test then I can merge :) Good work!

kamjin3086:
I pushed `f1be1a9` to fix the pytest collection failure. The failure was caused by importing the optional Paraformer dependency (`funasr`) at module import time, while the default CI dev environment does not install the `paraformer` extra. The handler now imports `AutoModel` inside `setup()`, so tests can import the module without `funasr`; users still get a clear error if they select Paraformer without installing the extra.

Validated locally on the latest PR merge ref `11fa772`:

```bash
uv sync --python python3.12 --group dev
uv run pytest tests/ -x -q
uv run ruff check src/ tests/
uv run mypy src/
```

Results: pytest `477 passed`, ruff check passed, ruff format check passed, mypy passed. I also validated package metadata and a wheel install smoke test on the PR branch.

andimarafioti (approved):
Thank you!!!

PR #300, opened 2026-06-05
Normalize Qwen3-TTS language aliases

kamjin3086:
## Summary

Hi, and thank you for the great work on `speech-to-speech`.

This PR adds a small language-normalization layer to the Qwen3-TTS handler. It accepts common language-code aliases such as `zh`, `zh-CN`, `en-US`, and `ja`, and maps them to the language names expected by the Qwen3-TTS backend, such as `chinese`, `english`, and `japanese`.

## Motivation

Qwen3-TTS validates requested languages against the model-supported language names. In practice, users often configure STT/LLM/TTS pipelines with BCP-47 or ISO-style language codes, e.g. `zh` for Chinese. Passing that directly to Qwen3-TTS can fail at synthesis time with an unsupported-language error, even though the intended target language is supported.

Normalizing these aliases in the handler makes the CLI more forgiving without changing the public option name or bypassing backend validation for genuinely unsupported values.

## Changes

- Add `QWEN3_LANGUAGE_ALIASES` for common aliases across the Qwen3-TTS supported languages.
- Normalize `language` during `Qwen3TTSHandler.setup()`.
- Treat `None` and empty strings as `auto`.
- Add tests for Chinese, English, and `auto` aliases.

## Validation

Ran locally:

```bash
PYTHONPATH=src python3 -m pytest tests/test_qwen3_tts_handler_backend.py -q
```

Result:

```text
43 passed, 1 warning
```

The warning was an unrelated local environment warning about `asyncio_mode` without `pytest-asyncio` installed.

### DIFF
--- a/src/speech_to_speech/TTS/qwen3_tts_handler.py
+++ b/src/speech_to_speech/TTS/qwen3_tts_handler.py
@@ -51,6 +51,28 @@ ESTIMATED_QWEN3_CHARS_PER_SECOND = 14.0
 QWEN3_TOKEN_SAFETY_MARGIN = 1.35
 QWEN3_BASE_PROMPT_SECONDS = 1.0
 QWEN3_PUNCTUATION_PAUSE_SECONDS = 0.5
+QWEN3_LANGUAGE_ALIASES = {
+    "zh": "chinese",
+    "zh-cn": "chinese",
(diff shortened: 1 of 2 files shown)


andimarafioti (approved):
Looking good!

PR #305, opened 2026-06-09
Shorten voice prompt and improve tool lead-ins

andimarafioti:
## Summary

- Shorten the voice-channel prompt wrapper and keep persona/session behavior in the session prompt.
- Shorten the local `<code>` tool-calling prompt while preserving parser-critical rules.
- Keep generic voice/tool prompts library-scoped: remove Reachy-specific mishearing aliases and avoid exact parrot-prone examples such as fixed camera/sadness lead-ins.
- Ask the model to use a brief natural utterance before tool calls, while avoiding claims about tool results before a tool result is available.
- Flush spoken lead-ins before tool-call events in both local parsed-tool and Responses API paths.
- Preserve ordered same-response text around tool calls where the current chunk model can support it: `text-only -> tool-only -> text-only` instead of dropping trailing text or bundling it with tools.
- Defensively enforce the current one-local-tool-call limit, while keeping arbitrary `text + tool + text + tool + ...` support as follow-up lifecycle work in #309.
- Add a lenient parser fallback for malformed local `<code>` blocks and cover the fallback with a test that actually forces `tokenize.TokenError`.
- Update the realtime tool-result README flow: clients send `response.create` when a tool result needs spoken follow-up; fire-and-forget robot actions can stop after the tool output is acknowledged.

## Impact

This keeps the speech-to-speech prompt surface smaller while making tool turns feel more natural and responsive. Spoken lead-ins are emitted before tool events, trailing text from the same model response is no longer silently dropped, and result-dependent speech remains tied to the explicit `function_call_output` + `response.create` follow-up flow.

The PR still intentionally preserves the current one parsed local tool-call limit. Full arbitrary interleaving across multiple text/tool parts needs a deeper response lifecycle change because `LLMResponseChunk(text, tools=[...])` and `AUDIO_RESPONSE_DONE` boundaries do not yet represent ordered assistant parts robustly; that is tracked in #309.

## Validation

- Inspected `pollen-robotics/reachy_mini_conversation_app` built-in tool definitions: `camera`, `dance`, `play_emotion`, `move_head`, `head_tracking`, `stop_dance`, `stop_emotion`, `idle_do_nothing`.
- Live GPT-5.4 checks with Reachy app-compatible tool names/schemas:
  - native OpenAI path called `camera` with required `question`
  - native path called `play_emotion` for explicit sadness performance
  - native path called `move_head`, `stop_dance`, and `head_tracking` with required arguments
  - local `<code>` prompt path produced parseable `camera(...)`, `dance(...)`, `play_emotion(...)`, and `move_head(...)` calls with spoken lead-ins
  - local parser recovered a malformed simple camera block and reduced duplicate tool blocks to one call
- Local tool prompt measurement: simple rendered prompt reduced from 205 words to 107 words before the optional-arg guard; the dance schema check is 151 words with argument docs.
- `PYTHONPATH=src python -m pytest tests/tool_call/test_function_parser.py tests/test_voice_prompt.py tests/test_responses_api_language_model.py`
- `PYTHONPATH=src python -m pytest tests/openai_realtime/test_realtime_service.py tests/openai_realtime/test_websocket_router.py`
- `PYTHONPATH=src python -m pytest tests/test_lm_output_processor.py tests/test_tts_input_coalescing.py`
- `python -m compileall -q src/speech_to_speech/LLM`
- Earlier PR validation:
  - `.venv/bin/pytest tests/test_voice_prompt.py tests/test_responses_api_language_model.py -q`
  - `.venv/bin/pytest tests/test_voice_prompt.py tests/tool_call/test_function_parser.py tests/tool_call/test_signature_from_schema.py tests/test_responses_api_language_model.py -q`
  - `.venv/bin/ruff check src/speech_to_speech/LLM/voice_prompt.py src/speech_to_speech/LLM/tool_call/tool_prompt.py src/speech_to_speech/LLM/tool_call/function_call.py src/speech_to_speech/LLM/language_model.py src/speech_to_speech/LLM/responses_api_language_model.py tests/test_voice_prompt.py tests/tool_call/test_function_parser.py tests/test_responses_api_language_model.py`
  - `git ls-files 'tests/*.py' 'tests/**/*.py' | xargs .venv/bin/pytest -q`

Note: plain `.venv/bin/pytest -q` collects an unrelated untracked `tests/test_benchmark_llm_latency.py` file in the local worktree, which fails on `ModuleNotFoundError: LLM.utils`. That untracked file is not part of this PR.


andimarafioti:
Pushed follow-up in `a312cef` to address the review comments and align the PR with the intended tool/text contract.

What changed:

- Fixed the malformed-block parser test so it actually exercises the regex fallback by forcing `tokenize.TokenError`.
- Added a parser comment clarifying that the regex fallback only recovers complete well-formed sibling calls, not the incomplete call that caused the tokenizer error.
- Dedented the local lead-in flush so pending `sentence_batch` text is emitted before a tool block even when the text immediately before `<code>` is empty.
- Changed local parsed-tool handling to stop dropping text after the tool block. It now emits parsed tools as a tools-only chunk when the block completes and leaves trailing text as normal printable text.
- Aligned the Responses API streaming and non-streaming paths around ordered chunks: `text-only -> tool-only -> text-only` where the current pipeline can represent that.
- Cleaned up the generic prompts by removing Reachy-specific mishearing aliases and replacing exact parrot-prone examples with pattern language.
- Kept the current one-local-tool-call implementation limit, but opened #309 for the deeper lifecycle work needed to support arbitrary interleaving like `text + tool + text + tool + ...` robustly.

Validation run after the changes:

- `PYTHONPATH=src python -m pytest tests/tool_call/test_function_parser.py tests/test_voice_prompt.py tests/test_responses_api_language_model.py`
- `PYTHONPATH=src python -m pytest tests/openai_realtime/test_realtime_service.py tests/openai_realtime/test_websocket_router.py`
- `PYTHONPATH=src python -m pytest tests/test_lm_output_processor.py tests/test_tts_input_coalescing.py`
- `python -m compileall -q src/speech_to_speech/LLM`

PR #310, opened 2026-06-11
Add realtime session warmup greeting

andimarafioti:
## Summary

Adds a one-time realtime session warmup response when the initial `session.update` supplies instructions.

The warmup uses the updated system prompt plus an ephemeral join prompt, so the assistant can greet or introduce itself without storing a synthetic user message in conversation history. This also exercises the configured LLM path before the user's first spoken turn, reducing first-response latency.

## Impact

- Initial realtime instructions can now produce a greeting/introduction immediately.
- The behavior applies to both local LLM and Responses API backends.
- Existing explicit `response.create` and audio/STT-triggered response flows remain unchanged.
- Docs now mention the one-time greeting/warmup behavior.

## Validation

- `.venv/bin/ruff check src/speech_to_speech/api/openai_realtime/service.py src/speech_to_speech/api/openai_realtime/websocket_router.py src/speech_to_speech/api/openai_realtime/handlers/response.py src/speech_to_speech/LLM/responses_api_language_model.py src/speech_to_speech/LLM/language_model.py src/speech_to_speech/pipeline/messages.py tests/openai_realtime/test_realtime_service.py tests/openai_realtime/test_websocket_router.py tests/test_responses_api_language_model.py`
- `.venv/bin/pytest tests/openai_realtime tests/test_responses_api_language_model.py tests/test_lm_output_processor.py tests/test_transcription_notifier.py`
- `.venv/bin/python -m compileall -q src/speech_to_speech/api/openai_realtime src/speech_to_speech/LLM src/speech_to_speech/pipeline/messages.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ with client.beta.realtime.connect(model="model_name") as conn:
 | Event | Description |
 |---|---|
 | `input_audio_buffer.append` | Stream base64 PCM audio. Decoded, resampled to 16 kHz, and chunked for the VAD. |
-| `session.update` | Deep-merge session config (instructions, tools, voice, turn detection, audio format). |
+| `session.update` | Deep-merge session config (instructions, tools, voice, turn detection, audio format). Initial instructions trigger a one-time greeting/warmup response. |
 | `conversation.item.create` | Inject `input_text` or `function_call_output` into the LLM context without triggering generation. |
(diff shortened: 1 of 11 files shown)


andimarafioti:
We decided that this shouldn't be done like this. Instead we'll handle it in the conversation app.

PR #318, opened 2026-06-15
Support text-only and out-of-band (conversation=none) responses

A-Mahla:
## What

Two related additions to `response.create` handling, both mirroring OpenAI realtime semantics.

### 1. Text-only responses via `output_modalities`

When the requested modalities are a non-empty list omitting `"audio"` (e.g. `["text"]`), the backend produces a **text-only** response: it skips TTS, emits `response.output_text.delta` / `response.output_text.done` instead of audio transcript + `response.output_audio.done`, and instructs the model with a written-channel system prompt.

- `output_modalities` absent (`None`), `[]`, or containing `"audio"` → audio (unchanged).
- a non-empty list without `"audio"` (e.g. `["text"]`) → text only.

### 2. Out-of-band responses via `conversation="none"`

When `conversation == "none"`, the response is generated against a **throwaway chat** and is never threaded into the default conversation.

- **Context** from `input`: `None` → read-only copy of the default conversation; `[]` → system prompt only (context cleared); `[...]` → system + those items.
- **Output is not committed** back to the default conversation.
- **`conversation_id` is reported as `null`** on `response.created` / `response.done`.
- Any non-`"none"` `conversation` value (`"auto"`, `None`, arbitrary id) keeps today's in-band behavior.

### Text-only (event / TTS gating)
- `response_wants_audio(response)` helper in `utils/utils.py` — single source of truth.
- `LMOutputProcessor` gates the `TTSInput` yield on it; `EndOfResponse` still flows so the lifecycle closes and the mic re-enables as before.
- `ResponseHandler.on_assistant_text` emits text delta/done when audio isn't wanted; `finish_audio_response` skips `response.output_audio.done` while still emitting `response.done`.
- New `LLM/text_prompt.py` (mirrors `voice_prompt.py`); `build_tool_system_prompt` gains a `text_only` variant. Both LM backends pick voice vs text builders via `wants_audio`.

### Out-of-band
- `is_out_of_band(response)` helper in `utils/utils.py`.
- `LLM/chat.py`: extracted `add_supported_item` (the `ConversationItem` → `SupportedItem` narrow+validate, now shared with the conversation handler) and `build_active_chat` implementing the three-way `input` rule.
- `handlers/response.py`: `handle_response_create` skips the default-chat append for out-of-band and tags the request with `turn_id=None`; `_build_response` nulls `conversation_id`.
- Both LM backends build the ephemeral chat via `build_active_chat` and suppress the commit-back to the default conversation.
- **Staleness bypass for free:** every `SpeculativeTurnTracker` method returns the permissive result for a `None` turn_id, so out-of-band responses are never dropped by a concurrent user turn — no gating-code changes needed.
- **Invalid `input`** surfaces as `response.done(status="failed")` via a new `EndOfResponse.error` → `ResponseFailedEvent` → `_on_response_failed` path (idempotent via the `in_response` guard) instead of taking down the worker.

### Scope notes
- Out-of-band reuses the single-response slot: rejected while another response is active, and barge-in / `response.cancel` still cancels it. True concurrency is out of scope.
- `input` is honored as fresh context only for `conversation="none"`; the in-band `input` path is unchanged.

## Tests

- **Text-only:** `response_wants_audio` gating in `LMOutputProcessor`; `on_assistant_text` text events; `finish_audio_response` skips audio.done; text prompt / tool-prompt content.
- **Out-of-band:** context composition (`input` present / `[]` / absent), output not committed to the default conversation, `conversation_id` null, null-turn staleness bypass, in-band contrast, invalid input → failed response, and `build_active_chat` units.

ruff, ruff format, mypy, and pytest (527 passed) all green locally.

### DIFF
--- a/src/speech_to_speech/LLM/chat.py
+++ b/src/speech_to_speech/LLM/chat.py
@@ -6,6 +6,7 @@ import threading
 from collections.abc import Callable
 from typing import Any, Literal, Union
 
+from openai.types.realtime import ConversationItem
 from openai.types.realtime.conversation_item import (
     RealtimeConversationItemAssistantMessage,
     RealtimeConversationItemFunctionCall,
@@ -18,6 +19,7 @@ from openai.types.realtime.realtime_conversation_item_assistant_message import (
 )
 from openai.types.realtime.realtime_conversation_item_system_message import Content as SystemContent
 from openai.types.realtime.realtime_conversation_item_user_message import Content as UserContent
+from openai.types.realtime.realtime_response_create_params import RealtimeResponseCreateParams
 from openai.types.responses.response_input_image_param import ResponseInputImageParam
 from openai.types.responses.response_input_message_content_list_param import (
     ResponseInputMessageContentListParam,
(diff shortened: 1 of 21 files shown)


andimarafioti (changes requested):
We chatted a bit about the system prompt. Otherwise, codex says:

Findings
[P1] Text-only responses emit response.output_text.done once per chunk.
In [response.py](https://github.com/huggingface/speech-to-speech/blob/ed3380d47c6594ec682c3584db8c0a74217b502a/src/speech_to_speech/api/openai_realtime/handlers/response.py#L272-L293), every AssistantTextEvent emits both response.output_text.delta and response.output_text.done with the same content_index=0 / output_index=0. But AssistantTextEvent is chunked; a single response can produce multiple chunks. I reproduced two chunks and got two “done” events: "Hello. " and "How are you?". Per the [Realtime text flow](https://developers.openai.com/api/docs/guides/realtime-conversations), output_text.done should represent the completed text content part, after deltas. This will make clients mark the content complete too early or overwrite partial text. Accumulate text per response, emit deltas per chunk, and emit one output_text.done at response completion with the full text.

[P2] response.done does not include the final text output.
_build_response() returns RealtimeResponse without output in [response.py](https://github.com/huggingface/speech-to-speech/blob/ed3380d47c6594ec682c3584db8c0a74217b502a/src/speech_to_speech/api/openai_realtime/handlers/response.py#L116-L129). For text-only responses, OpenAI’s docs explicitly show clients reading the final result from response.done.response.output[0]; this implementation leaves response.output as None. This is especially visible now that [ "text" ] responses are supported. The same accumulator used for the final output_text.done can populate RealtimeResponse(output=[assistant message...]).

A-Mahla:
The P1 is something I have noticed for a while. Because our delta yield is sentence-based instead of character-based, as OpenAI does, we have been forcing the output text done event for sentence chunks for latency purpose.

The P2 is a direct consequence of the first issue. If we send the accumulated text in output_done, as is the case with OpenAI, this will cause Reachy to say the same output twice.

Both are pre-existing before the PR and the reachy mini conversation app only uses the output text done event. If we sent delta events instead for chunked sentences, the result would be that it waits for the full LLM response before responding. I would keep it as is for now.

andimarafioti (approved):
Looking good ! Thank you for addressing the feedback!

PR #320, opened 2026-06-17
Add chat-completions LLM backend (OpenAI /v1/chat/completions)

### LINKED ISSUE #312: Add Chat Completions backend for providers where Responses cannot disable reasoning
## Problem

Some OpenAI-compatible inference providers do not behave the same way through the Responses API and Chat Completions API. For those providers, the current `responses-api` backend may be unable to fully disable reasoning or may delay visible text until after reasoning tokens are streamed.

This matters for low-latency voice/conversation usage: if the app ignores reasoning text for TTS, the assistant can wait a long time before any speakable output arrives, or never receive speakable text before `max_output_tokens` is exhausted.

## Concrete example: GLM-4.7 on Cerebras via Hugging Face router

Tested with:

- Model: `zai-org/GLM-4.7:cerebras`
- Base URL: `https://router.huggingface.co/v1`
- Scenario: short robot conversation/math prompts, streaming enabled

Responses API behavior:

- `reasoning={"effort": "low"}` is accepted.
- `reasoning={"effort": "none"}` is rejected with a 400 because valid values are only `low`, `medium`, and `high`.
- `extra_body={"reasoning_effort": "none"}` is accepted but appears ignored on the Responses path.
- `extra_body={"chat_template_kwargs": {"enable_thinking": false}}` also appears ignored for this model/provider path.
- Even with `reasoning.low`, the stream emits `response.reasoning_text.delta` before `response.output_text.delta`.
- In a 20-call test, `reasoning.low` produced visible text in only 15/20 calls; the other calls spent the entire 1024-token output budget on reasoning and emitted no visible answer text.

Observed latency for Responses `reasoning.low`, counting time until visible/non-reasoning text finished:

- Visible text arrived: 15/20 calls
- Median final visible text: ~1.10s
- P90 final visible text: ~1.57s
- P95 final visible text: ~1.66s
- Median total response: ~1.33s

Chat Completions behavior for the same provider/model:

```python
client.chat.completions.create(
    model="zai-org/GLM-4.7:cerebras",
    messages=messages,
    stream=True,
    extra_body={"reasoning_effort": "none"},
)
```

In the same 20-call test:

- Reasoning chunks observed: 0/20
- Visible text arrived: 20/20
- Median final visible text: ~0.38s
- P90 final visible text: ~0.54s
- P95 final visible text: ~0.61s
- Median total response: ~0.39s

This suggests that for this provider/model, Chat Completions with provider-specific `extra_body={"reasoning_effort": "none"}` is the usable low-latency path, while Responses `reasoning.low` is not sufficient for voice latency.

## Requested change

Add a Chat Completions-compatible LLM backend/path for OpenAI-compatible hosted providers where Responses is not the best abstraction.

The new path should support:

- OpenAI-compatible Chat Completions endpoint selection.
- Provider/model configuration from CLI/config, including `base_url`, `api_key`, and `model_name`.
- Passing provider-specific `extra_body`, especially:

```python
extra_body={"reasoning_effort": "none"}
```

- Streaming visible assistant text as soon as it arrives.
- Optional handling/ignoring of reasoning chunks if a provider exposes them separately.

## Why not just use Responses `reasoning.low`?

For at least the Cerebras-routed GLM-4.7 example, `low` is not a hard reasoning-token cap. It can still produce hundreds or thousands of characters of reasoning, and it can consume the full output token budget before any visible answer is emitted.

For a robot conversation app, that means poor perceived latency and occasional silent turns.

## Notes

The current Responses handler also stores generation kwargs but does not appear to forward arbitrary `gen_kwargs` into `client.responses.create()`. Even if that is fixed, the Cerebras example still needs Chat Completions because the Responses endpoint does not support `reasoning.none` as a top-level reasoning effort and appears to ignore the relevant extra-body controls for this path.

JonasZimmer1994:
## Summary

Adds a fourth `--llm_backend` value, `chat-completions`, that drives the
OpenAI-compatible `/v1/chat/completions` endpoint instead of `/v1/responses`.
It mirrors the existing `responses-api` handler (streaming, tool calls, token
usage, history, compaction, cancellation) but uses the Chat Completions
streaming protocol (`choices[].delta.tool_calls`).

Closes #312.

## Motivation

Two independent problems with the Responses path on some OpenAI-compatible
servers, both fixed by using Chat Completions:

1. **Reasoning cannot be disabled (the #312 case).** For some providers (e.g.
   `zai-org/GLM-4.7:cerebras` via the HF router) neither `reasoning.effort` nor
   `chat_template_kwargs.enable_thinking=false` reliably suppresses reasoning on
   `/v1/responses`, so speakable text is delayed or never produced within the
   output-token budget. Chat Completions with `extra_body={"reasoning_effort":
   "none"}` returns visible text immediately.

2. **Responses streaming tool calls are unreliable on some vLLM builds.** With
   auto tool choice + streaming, vLLM's `responses_stream_generator` can raise a
   pydantic `ValidationError` building `ResponseCreatedEvent` (the internal
   `json_schema` tool-calling format fails `response.text.format` validation),
   killing the stream; replayed tool-call arguments can also arrive duplicated
   and fail `json.loads`. The Chat Completions streaming tool-call path is
   unaffected.

## What changed

- `LLM/chat_completions_language_model.py` — new `ChatCompletionsApiModelHandler`.
  Serialises the conversation with `Chat.to_transformers_chat()` and re-encodes
  tool-call `arguments` from object back to a JSON string (the Chat Completions
  API requires a string, whereas `apply_chat_template` wants an object).
- `arguments_classes/chat_completions_language_model_arguments.py` — new
  `ChatCompletionsLanguageModelHandlerArguments`, subclassing the responses-api
  arguments (shared connection flags) and adding `responses_api_reasoning_effort`.
- `arguments_classes/module_arguments.py` — `chat-completions` added to the
  `llm_backend` literal.
- `s2s_pipeline.py` — backend→argument-class selection, handler factory branch,
  shared chat-size / runtime-config wiring.
- `README.md` — documents the new backend and `reasoning_effort`.
- `tests/test_chat_completions_backend.py` — mock-based unit tests (no GPU/server).

### `reasoning_effort` precedence

`extra_body` is built as: explicit `reasoning_effort` → `{"reasoning_effort": …}`;
else `disable_thinking` → `{"chat_template_kwargs": {"enable_thinking": false}}`;
else none. Never sent to the official OpenAI server (`api.openai.com`).

## Testing

- Unit tests (`tests/test_chat_completions_backend.py`, 9 tests) cover the tool/
  tool_choice converters, `arguments` string re-encoding, `extra_body` precedence,
  and the streaming + non-streaming parse via a faked client. They need no
  network (`pytest tests/test_chat_completions_backend.py`).
- Verified live against a local vLLM (`Qwen3` model, `--tool-call-parser
  qwen3_coder`): plain text, forced tool call with arguments reassembled from
  streamed deltas, and multi-turn replay of stored tool calls — no 400s/crashes,
  in both streaming and non-streaming modes.

## Design note: shared vs. dedicated arguments

`ChatCompletionsLanguageModelHandlerArguments` subclasses the responses-api
arguments, so both OpenAI-compatible backends share the `--responses_api_*` flags
and a single pipeline slot. This keeps the change small and avoids threading a new
argument set through the pipeline builders. The alternative — a fully independent
`--chat_completions_*` argument set with its own slot — is a reasonable design if
stronger separation between the two backends is preferred.

### DIFF
--- a/src/speech_to_speech/LLM/chat_completions_language_model.py
+++ b/src/speech_to_speech/LLM/chat_completions_language_model.py
@@ -1,0 +1,559 @@
+from __future__ import annotations
+
+import json
+import logging
+import time
+from collections.abc import Iterator
+from typing import Any, Optional
+
+import httpx
+from nltk import sent_tokenize
+from openai import OpenAI, Stream
+from openai.types.chat import ChatCompletionChunk
+from openai.types.realtime.conversation_item import (
+    RealtimeConversationItemAssistantMessage,
+    RealtimeConversationItemFunctionCall,
+)
+from openai.types.realtime.realtime_conversation_item_assistant_message import (
+    Content as AssistantContent,
+)
+from openai.types.responses import ResponseFunctionToolCall
+
+from speech_to_speech.baseHandler import BaseHandler
+from speech_to_speech.LLM.chat import Chat, SupportedItem, make_system_message, make_user_message
+from speech_to_speech.LLM.compaction_prompt import CompactGenerateFn, build_compactor
+from speech_to_speech.LLM.utils import remove_unspeechable, resolve_auto_language
+from speech_to_speech.LLM.voice_prompt import build_voice_system_prompt
+from speech_to_speech.pipeline.cancel_scope import CancelScope
+from speech_to_speech.pipeline.handler_types import LLMIn, LLMOut
+from speech_to_speech.pipeline.messages import (
+    EndOfResponse,
+    LLMResponseChunk,
+    TokenUsage,
+)
+from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
+from speech_to_speech.utils.utils import _generate_id
+
+logger = logging.getLogger(__name__)
+
+
+def _to_chat_tools(req_tools: Any) -> list[dict[str, Any]] | None:
+    """Convert Responses-API function tools to Chat-Completions tool format.
+
+    Responses tools are flat ``{type:"function", name, description, parameters}``;
+    Chat Completions nests them under a ``function`` key. Items already in the
+    nested form (or non-function tools) are passed through untouched.
+    """
+    if not req_tools:
+        return None
+    chat_tools: list[dict[str, Any]] = []
+    for t in req_tools:
+        d = t if isinstance(t, dict) else t.model_dump(exclude_none=True)
+        if d.get("type") == "function" and "function" not in d:
+            fn = {k: d[k] for k in ("name", "description", "parameters") if k in d}
+            chat_tools.append({"type": "function", "function": fn})
+        else:
+            chat_tools.append(d)
+    return chat_tools
+
+
+def _to_chat_tool_choice(tool_choice: Any) -> Any:
+    """Convert a Responses-API tool_choice to Chat-Completions form.
+
+    The string forms ("auto"/"required"/"none") are identical across both APIs;
+    only the forced-function object differs (flat ``name`` vs nested ``function``).
+    """
+    if isinstance(tool_choice, dict) and tool_choice.get("type") == "function" and "name" in tool_choice:
+        return {"type": "function", "function": {"name": tool_choice["name"]}}
+    if tool_choice is not None and not isinstance(tool_choice, (str, dict)):
+        d = tool_choice.model_dump(exclude_none=True)
+        if d.get("type") == "function" and "name" in d:
+            return {"type": "function", "function": {"name": d["name"]}}
+        return d
+    return tool_choice
+
+
+class ChatCompletionsApiModelHandler(BaseHandler[LLMIn, LLMOut]):
+    """LLM handler that talks to an OpenAI-compatible ``/v1/chat/completions`` server.
+
+    Functionally mirrors :class:`ResponsesApiModelHandler` but uses the mature
+    Chat Completions streaming tool-call protocol (``choices[].delta.tool_calls``)
+    instead of ``/v1/responses``. This is the robust path for vLLM + Qwen tool
+    calling. The conversation is serialised with :meth:`Chat.to_transformers_chat`,
+    which already emits OpenAI chat messages including ``tool_calls``/``tool`` roles.
+    """
+
+    def setup(
+        self,
+        model_name: str = "gpt-5.4-mini",
+        device: str = "cuda",
+        gen_kwargs: dict[str, Any] = {},
+        base_url: Optional[str] = None,
+        api_key: Optional[str] = None,
+        stream: bool = True,
+        user_role: str = "user",
+        cancel_scope: CancelScope | None = None,
+        speculative_turns: SpeculativeTurnTracker | None = None,
+        disable_thinking: bool = True,
+        reasoning_effort: Optional[str] = None,
+        request_timeout_s: float = 20.0,
+        stream_batch_sentences: int = 3,
+        enable_lang_prompt: bool = False,
+        compact_history: bool = False,
+        **_kwargs: Any,
+    ) -> None:
+        self.cancel_scope = cancel_scope
+        self.speculative_turns = speculative_turns
+        self.model_name = model_name
+        self.stream = stream
+        self.stream_batch_sentences = max(1, stream_batch_sentences)
+        self.enable_lang_prompt = enable_lang_prompt
+        self.gen_kwargs = dict(gen_kwargs)
+        self.request_timeout_s = float(request_timeout_s)
+        self.request_timeout = httpx.Timeout(
+            self.request_timeout_s,
+            connect=min(10.0, self.request_timeout_s),
+        )
+
+        self.user_role = user_role
+        self.client = OpenAI(api_key=api_key, base_url=base_url)
+        self._extra_body = self._build_extra_body(base_url, disable_thinking, reasoning_effort)
+        self.compactor = build_compactor(self._build_compaction_generate_fn()) if compact_history else None
+        self.warmup()
+
+    @staticmethod
+    def _build_extra_body(
+        base_url: Optional[str],
+        disable_thinking: bool,
+        reasoning_effort: Optional[str],
+    ) -> Optional[dict[str, Any]]:
+        """Build the provider-specific ``extra_body`` used to disable reasoning.
+
+        Providers differ in how reasoning is turned off: vLLM/Qwen honour
+        ``chat_template_kwargs.enable_thinking=false``, while others (e.g. GLM via
+        the HF router) ignore that and require ``reasoning_effort='none'``. An
+        explicit ``reasoning_effort`` therefore takes precedence; otherwise we fall
+        back to the chat-template flag. None of this applies to the official
+        OpenAI server, which rejects unknown extra_body keys.
+        """
+        if base_url is None or base_url == "https://api.openai.com/v1":
+            return None
+        if reasoning_effort is not None:
+            return {"reasoning_effort": reasoning_effort}
+        if disable_thinking:
+            return {"chat_template_kwargs": {"enable_thinking": False}}
+        return None
+
+    def _turn_is_latest(self, turn_id: str | None, turn_revision: int | None) -> bool:
+        return self.speculative_turns is None or self.speculative_turns.is_latest(turn_id, turn_revision)
+
+    def _generation_is_stale(self, gen: int | None) -> bool:
+        return gen is not None and self.cancel_scope is not None and self.cancel_scope.is_stale(gen)
+
+    def _turn_output_allowed(self, turn_id: str | None, turn_revision: int | None) -> bool:
+        if self.speculative_turns is None:
+            return True
+        return self.speculative_turns.is_latest_after_reopen_grace(turn_id, turn_revision)
+
+    def warmup(self) -> None:
+        logger.info(f"Warming up {self.__class__.__name__}")
+        start = time.time()
+        self.client.chat.completions.create(
+            model=self.model_name,
+            messages=[
+                {"role": "system", "content": "You are a helpful assistant"},
+                {"role": "user", "content": "Hello"},
+            ],
+            extra_body=self._extra_body,
+            timeout=self.request_timeout,
+        )
+        end = time.time()
+        logger.info(f"{self.__class__.__name__}:  warmed up! time: {(end - start):.3f} s")
+
+    def _build_compaction_generate_fn(self) -> CompactGenerateFn:
+        """Return a generate fn that calls Chat Completions for compaction."""
+        client = self.client
+        model_name = self.model_name
+        timeout = self.request_timeout
+        extra_body = self._extra_body
+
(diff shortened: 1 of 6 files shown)


src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+        turn_id = request.turn_id
+        turn_revision = request.turn_revision
+        speech_stopped_at_s = request.speech_stopped_at_s
+        if not self._turn_is_latest(turn_id, turn_revision):
+            logger.info("Skipping stale LLM request for turn=%s rev=%s", turn_id, turn_revision)
+            yield EndOfResponse(turn_id=turn_id, turn_revision=turn_revision)
+            return
+
+        original_chat = runtime_config.chat
+        active_chat = original_chat.copy()
```
A-Mahla, line 515: **Critical: out-of-band support is missing.**

`responses_api` checks `is_out_of_band(response)` here: when true, it calls `build_active_chat(original_chat, response)` inside a `try/except ChatItemError` that short-circuits with `EndOfResponse(error=str(exc))` on failure. When false, it falls back to `original_chat.copy()`. Without this check, out-of-band requests silently use the full conversation context.

There is a matching gap in `_generate`'s post-generation block (line 430): the `original_chat.add_item` loop runs unconditionally, but `responses_api` guards it with `if not is_out_of_band(response)`. Without that guard, out-of-band responses corrupt the default chat history.

Missing imports: `is_out_of_band`, `response_wants_audio`, `build_active_chat`, `ChatItemError`.

src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+                    if chunk.usage is not None:
+                        input_tokens = chunk.usage.prompt_tokens or 0
+                        output_tokens = chunk.usage.completion_tokens or 0
+                    if not chunk.choices:
+                        continue
+
+                    delta = chunk.choices[0].delta
+
+                    if delta.content:
+                        new_text = remove_unspeechable(delta.content)
```
A-Mahla, line 294: **Critical: text-only (`wants_audio=False`) path is missing.**

`responses_api` computes `wants_audio = response_wants_audio(response)` at the top of `_generate` and branches here:

```python
if not wants_audio:
    # forward delta verbatim: no remove_unspeechable, no sent_tokenize
    delta = raw_event.delta
    clean_text += delta
    if delta:
        yield LLMResponseChunk(text=delta, ...)
    continue
# audio path below
new_text = remove_unspeechable(raw_event.delta)
```

Without the branch, text-only responses (markdown, structured output) have symbols stripped by `remove_unspeechable` and are silently altered. The same issue applies at line 366 (non-streaming path). `_apply_config` at line 524 also needs a `wants_audio: bool` parameter to choose between `build_voice_system_prompt` and `build_text_system_prompt`.

src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+                    cancel_generation=gen,
+                )
+        finally:
+            if api_response is not None and hasattr(api_response, "close"):
+                try:
+                    api_response.close()
+                except Exception:
+                    pass
+
+        if not self._generation_is_stale(gen) and self._turn_output_allowed(turn_id, turn_revision):
```
A-Mahla, line 430: **Critical: two guards are missing from this condition.**

**1. `error_message is None`** — `responses_api` maintains an `error_message: str | None = None` variable set when the API input is empty or when an exception is caught. History write-back must be gated on `error_message is None` to prevent committing output from a failed generation. This handler has no `error_message` variable at all.

**2. `not is_out_of_band(response)`** — out-of-band responses use a throwaway chat; writing them back into `original_chat` corrupts the default conversation across turns.

Correct guard:
```python
if (
    error_message is None
    and not self._generation_is_stale(gen)
    and self._turn_output_allowed(turn_id, turn_revision)
    and not is_out_of_band(response)
):
```

src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+            original_chat.strip_images()
+            original_chat.trim_if_needed(self.compactor)
+            if input_tokens or output_tokens:
+                yield TokenUsage(
+                    input_tokens=input_tokens,
+                    output_tokens=output_tokens,
+                    turn_id=turn_id,
+                    turn_revision=turn_revision,
+                )
+        yield EndOfResponse(turn_id=turn_id, turn_revision=turn_revision, cancel_generation=gen)
```
A-Mahla, line 442: **Critical: `error=error_message` is missing.**

`responses_api` always passes it: `EndOfResponse(..., cancel_generation=gen, error=error_message)`. Callers rely on the `error` field to distinguish a clean end from a timeout or generation failure. Without `error_message` tracking in this handler (see comment on L430), the field would always be `None` even when an exception was caught, leaving the client unable to surface the error.

src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+                    if printable_text.strip():
+                        sentence_batch.append(printable_text.strip())
+                    if sentence_batch:
+                        if self._generation_is_stale(gen):
+                            logger.info("LLM generation cancelled (interruption)")
+                        else:
+                            logger.debug(f"Clean text: {clean_text}")
+                            yield from _flush_batch(sentence_batch)
+                    if clean_text.strip():
+                        pending_chat_items.append(
```
A-Mahla, line 334: **Medium: history stores `remove_unspeechable`-filtered text instead of raw text.**

`clean_text` is the TTS-ready string. Storing it in history means future turns show the model a filtered view of what it previously said, which can cause subtle drift in multi-turn conversations. Maintain a separate raw accumulator that collects `delta.content` without filtering and use that for `pending_chat_items`.

The same issue affects the non-streaming path at line 369: `pending_chat_items.append` is gated on `if message_text.strip()` where `message_text` is the filtered string. If `remove_unspeechable` strips everything, the assistant turn is silently dropped from history. The gate should use the raw content: `if message.content`.

src/speech_to_speech/LLM/chat_completions_language_model.py
```diff
@@ -0,0 +1,559 @@
...
+            response.tool_choice if response and response.tool_choice else runtime_config.session.tool_choice
+        )
+        self._apply_config(active_chat, instructions)
+        language_code, lang_name = resolve_auto_language(language_code)
+        if lang_name and self.enable_lang_prompt:
+            active_chat.add_item(make_user_message(f"Please reply to my message in {lang_name}."))
+
+        optional_kwargs: dict[str, Any] = {}
+        chat_tools = _to_chat_tools(req_tools)
+        if chat_tools is not None:
```
A-Mahla, line 531: **Medium: `tool_choice` is silently dropped when `req_tools` is falsy.**

`tool_choice` is nested inside the `chat_tools is not None` guard. `responses_api` sends the two independently:

```python
if req_tools is not None:
    optional_kwargs["tools"] = req_tools
if req_tool_choice is not None:
    optional_kwargs["tool_choice"] = req_tool_choice
```

A session setting `tool_choice="none"` (to suppress tool use) with no explicit tools list will lose the directive entirely under the current logic.

A-Mahla (changes requested):
Great work on this, really appreciate the contribution! A few gaps need to be closed before this can merge, all of them places where `responses_api_language_model.py` has grown features that are not yet reflected here.

The new `ChatCompletionsApiModelHandler` is missing three features that exist in `ResponsesApiModelHandler` today: out-of-band response handling, the text-only (`wants_audio=False`) code path, and `error_message` propagation to `EndOfResponse`. There are also two medium-severity divergences in how conversation history is written back and how `tool_choice` is sent.

Two low-priority observability gaps are noted below. They do not block merge but should be fixed in the same PR to keep the two handlers symmetric.

---

**Low: `_flush_batch` cancellation logging (L243-248)** — When `_turn_output_allowed` returns `False`, `_flush_batch` silently returns. `responses_api` logs `"LLM generation cancelled (stale speculative turn)"` before breaking. Without the log, stale-turn cancellations are invisible in the output.

**Low: `Tools:` log gated on non-empty (streaming L419, non-streaming L472)** — `logger.info(f"Tools: {tools}")` is gated on `if tools:` in both paths. `responses_api` logs this unconditionally on every successful generation. Log unconditionally for symmetry (an empty list is useful signal).

A-Mahla:
Merging this PR into the internal branch so we can work on merging it into main.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: `process()` now branches on `is_out_of_band(response)` to build a throwaway chat via `build_active_chat` (rejecting invalid input with `EndOfResponse(error=...)`), and the history write-back is guarded by `if not is_out_of_band(response)`. Imports added. Tracked in #322.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: `wants_audio = response_wants_audio(response)` drives a text-only branch in both the streaming and non-streaming paths (verbatim delta, no `remove_unspeechable`, no sentence-splitting), and `_apply_config` now takes `wants_audio` to pick `build_text_system_prompt`. Tracked in #322.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: added an `error_message` variable (empty-input guard + catch-all `except`); the write-back is now gated on `error_message is None and ... and not is_out_of_band(response)`. Tracked in #322.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: `EndOfResponse` now carries `error=error_message`, set by the empty-input guard and the generation-failure `except`. Tracked in #322.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: a separate raw accumulator stores unfiltered assistant text in history; the non-streaming write is gated on raw `message.content` so an all-symbol turn is no longer dropped. Tracked in #322.

src/speech_to_speech/LLM/chat_completions_language_model.py
A-Mahla: Done in a4e735c: `tool_choice` is now sent independently of `tools`, so `tool_choice="none"` survives with no tools list. Tracked in #322.

PR #321, opened 2026-06-19
Emit assistant transcript for fresh response when discard guard is stuck

A-Mahla:
## Problem

When speculative decoding fires, the response streams audio and `response.done` but **never sends `response.output_audio_transcript.done`**. From a live session log, the speculatively-decoded response emitted `response.created` → `output_audio.delta` → `output_audio.done` → `response.done`, with the transcript event missing entirely.

## Root cause

An asymmetry between how audio and assistant-text are discarded, combined with a `cancel_scope.discarding` flag that can get stuck `True`:

1. **`discarding` sticks `True`.** It's set by `cancel_scope.cancel()` (barge-in/reopen) and cleared only by `response_done()` — which fires when an `AUDIO_RESPONSE_DONE` sentinel reaches the send loop — or by a client `new_response()`. Server-VAD speculative responses never call `new_response()`, and when a speculative generation is superseded, every TTS handler drops its stale `EndOfResponse` **without emitting `AUDIO_RESPONSE_DONE`**. So that generation's sentinel never arrives and `discarding` stays `True`.

2. **Audio discard was generation-aware; text discard was not.** `_should_discard_audio` only drops audio whose generation is *not the current one*, so the next response's audio (current generation) streamed and completed fine. But the assistant-text drops in the send loop and in `_drain_pending_response_events` were a blanket `if discarding: drop` — so the *current, live* response's transcript was silently swallowed.

## Fix

Make assistant-text discard use the same generation-aware rule as audio:

- Add `cancel_generation` to `AssistantTextEvent` (mirroring `AudioOutput`).
- Populate it from `LLMResponseChunk.cancel_generation`, which the LLM already sets — it was just being dropped when building the event.
- Extract the shared `_generation_is_discardable(unit, generation)` helper, reuse it in `_should_discard_audio`, and use it at both assistant-text discard sites.

A current-generation transcript now survives even when `discarding` lingers. The `None`-generation path is exactly backward-compatible with the old blanket behavior.

## Tests

- Added `test_current_generation_text_survives_stuck_discarding` reproducing the exact scenario.
- Full suite green: 236 passed.

## Follow-up (out of scope)

`cancel_scope.discarding` can still get stuck `True` at its source (a superseded speculative generation never emits a sentinel to clear it). This change makes that harmless for transcripts; guaranteeing `response_done()` fires for every cancelled generation would be a deeper, separate fix.

### DIFF
--- a/src/speech_to_speech/LLM/lm_output_processor.py
+++ b/src/speech_to_speech/LLM/lm_output_processor.py
@@ -125,6 +125,7 @@ class LMOutputProcessor(BaseHandler[LLMOut, TTSIn]):
                 text=lm_output.text,
                 turn_id=lm_output.turn_id,
                 turn_revision=lm_output.turn_revision,
+                cancel_generation=lm_output.cancel_generation,
             )
             if lm_output.tools:
                 event.tools = lm_output.tools
--- a/src/speech_to_speech/api/openai_realtime/websocket_router.py
+++ b/src/speech_to_speech/api/openai_realtime/websocket_router.py
@@ -139,7 +139,7 @@ async def _drain_pending_response_events(
                 drained_usage += 1
             elif drain_assistant_events and isinstance(item, AssistantTextEvent):
                 drained_assistant += 1
-                if unit.cancel_scope.discarding:
+                if _generation_is_discardable(unit, item.cancel_generation):
                     continue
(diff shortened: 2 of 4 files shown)


andimarafioti (approved):
Great work!

PR #322, opened 2026-06-22
Add chat-completions LLM backend (OpenAI /v1/chat/completions)

### LINKED ISSUE #312: Add Chat Completions backend for providers where Responses cannot disable reasoning

A-Mahla:
## Summary

Adds a fourth `--llm_backend` value, `chat-completions`, that drives the
OpenAI-compatible `/v1/chat/completions` endpoint instead of `/v1/responses`.
It mirrors the existing `responses-api` handler (streaming, tool calls, token
usage, history, compaction, cancellation, text-only and out-of-band responses)
but uses the Chat Completions streaming protocol (`choices[].delta.tool_calls`).

Originally contributed in #320 (closes #312); merged here and brought up to
parity with the latest `responses_api_language_model.py` before merging to `main`.

## Motivation

Two independent problems with the Responses path on some OpenAI-compatible
servers, both fixed by using Chat Completions:

1. **Reasoning cannot be disabled.** For some providers (e.g. `zai-org/GLM-4.7:cerebras`
   via the HF router) neither `reasoning.effort` nor `chat_template_kwargs.enable_thinking=false`
   reliably suppresses reasoning on `/v1/responses`, so speakable text is delayed
   or never produced within the output-token budget. Chat Completions with
   `extra_body={"reasoning_effort": "none"}` returns visible text immediately.
2. **Responses streaming tool calls are unreliable on some vLLM builds.** With
   auto tool choice + streaming, vLLM's `responses_stream_generator` can raise a
   pydantic `ValidationError` and kill the stream; replayed tool-call arguments
   can also arrive duplicated. The Chat Completions streaming tool-call path is
   unaffected.

## What changed

- `LLM/chat_completions_language_model.py` — new `ChatCompletionsApiModelHandler`.
  Serialises the conversation with `Chat.to_transformers_chat()` and re-encodes
  tool-call `arguments` from object back to a JSON string.
- `arguments_classes/chat_completions_language_model_arguments.py` — new
  `ChatCompletionsLanguageModelHandlerArguments`, subclassing the responses-api
  arguments (shared connection flags) and adding `responses_api_reasoning_effort`.
- `arguments_classes/module_arguments.py` — `chat-completions` added to the
  `llm_backend` literal.
- `s2s_pipeline.py` — backend→argument-class selection, handler factory branch,
  shared chat-size / runtime-config wiring.
- `README.md` — documents the new backend and `reasoning_effort`.
- `tests/test_chat_completions_backend.py` — mock-based unit tests (no GPU/server).

## Review fixes applied on top of #320

The #320 review flagged places where `responses_api_language_model.py` had grown
features the new handler did not yet reflect. All are now closed:

- **Out-of-band** (`conversation="none"`) responses build a throwaway active chat
  via `build_active_chat` (invalid input → `EndOfResponse.error`) and are never
  written back into the default conversation.
- **Text-only** (`output_modalities=["text"]`) responses forward deltas verbatim
  (no `remove_unspeechable`, no sentence-splitting) in both streaming and
  non-streaming paths, and select `build_text_system_prompt`.
- **`error_message` tracking**: empty-input guard + catch-all `except`, history
  write-back gated on `error_message is None`, and `error=error_message`
  propagated on `EndOfResponse`.
- **History stores raw text** (not the TTS-filtered `clean_text`), preventing
  multi-turn drift; non-streaming write gated on raw `message.content`.
- **`tool_choice` sent independently of `tools`**, so `tool_choice="none"`
  survives with no tools list.
- **Observability parity**: stale-turn cancellation logged in `_flush_batch`;
  `Tools:` logged unconditionally.

## Testing

`pytest tests/test_chat_completions_backend.py` — 15 tests, no network. Covers the
tool/tool_choice converters, `arguments` string re-encoding, `extra_body`
precedence, streaming + non-streaming parse, plus the new text-only, decoupled
`tool_choice`, empty-input / generation-error, and out-of-band behaviours.
`responses_api` and `chat` suites remain green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,8 +256,9 @@ The LLM is the most compute-intensive and highest-latency component in the pipel
 - **Local inference** — `transformers` (CUDA / CPU) and `mlx-lm` (Apple Silicon) run the model entirely on your machine with no external dependency.
 - **Self-hosted servers** — `--llm_backend responses-api` can point at a local [vLLM](https://github.com/vllm-project/vllm) or [llama.cpp](https://github.com/ggerganov/llama.cpp) server, giving you control over quantization, batching, and hardware while keeping traffic on-premise.
 - **Provider APIs** — the same `responses-api` backend works with OpenAI, [HuggingFace Inference Providers](https://huggingface.co/inference-providers), [OpenRouter](https://openrouter.ai), and any other provider that implements the OpenAI Responses API.
+- **Chat Completions** — `--llm_backend chat-completions` targets the OpenAI-compatible `/v1/chat/completions` endpoint instead of `/v1/responses`. Use it for providers/servers where Responses cannot reliably disable reasoning, or whose Responses streaming tool-call path is unreliable (see [#312](https://github.com/huggingface/speech-to-speech/issues/312)).
 
-Select a backend with `--llm_backend` (`responses-api` by default) and pair it with `--model_name`. Backend-specific options (`--responses_api_base_url`, `--responses_api_api_key`, `--responses_api_stream`, etc.) are only needed for the `responses-api` backend.
+Select a backend with `--llm_backend` (`responses-api` by default) and pair it with `--model_name`. Backend-specific options (`--responses_api_base_url`, `--responses_api_api_key`, `--responses_api_stream`, etc.) are shared by the `responses-api` and `chat-completions` backends.
 
 > The examples below pair Parakeet TDT (local STT) and Qwen3-TTS (local TTS) with different LLM backends.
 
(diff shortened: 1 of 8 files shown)


andimarafioti (approved):
LGTM :D

PR #326, opened 2026-06-29
Fix two mid-generation conversation races (tool call & image)

A-Mahla:
## Problem

A client can mutate the conversation **while the LLM handler is still generating**. `conversation.item.create` (tool outputs, images, text) is applied on the asyncio/router thread, while a generation reads and writes the chat on the pipeline thread. `Chat._lock` makes each individual operation atomic, but **not** a generation's whole *serialize → generate → write-back* cycle — so a fast tool/image client reliably lands items in that window.

This surfaced as two concrete failures while testing a `camera_snapshot` round-trip against `/v1/responses`.

### 1. Tool call not in history when the client returns its output
The `function_call` was only recorded in the deferred end-of-turn write-back. A fast client returned its `function_call_output` first, so pairing failed with `No function_call with call_id ... found` and the model **re-issued the same tool call**.

`_record_tool_call` now persists the call (and any preceding assistant text, for correct history order) **before** the chunk is forwarded to the client. The call lands in `_pending_tool_calls` (not serialized until its output pairs it), so eager recording is safe; out-of-band and stale turns still record nothing.

### 2. Image stripped before the response that needed it
`strip_images()` ran in the write-back and removed **every** image in the buffer — including one a fast client injected mid-generation for the *next* turn. Token counts confirmed the model never saw the image and hallucinated a description.

We now capture the image-bearing message ids the request **actually consumed** (`Chat.image_message_ids()` at serialize time) and `strip_images(only_ids)` strips only those, sparing the next turn's image. Applied to both the OpenAI-compatible base handler and the legacy `language_model.py` handler.

## Tests
- `test_tool_call_recorded_before_chunk_is_emitted` — call is in `_pending_tool_calls` at chunk-emit time; a fast `function_call_output` pairs without raising.
- `test_image_message_ids_reports_only_image_carriers`, `test_strip_images_only_ids_spares_concurrent_image`.
- Full suite: **560 passing**; ruff + `ruff format` clean; mypy clean on changed files.

## Follow-up
A separate PR will add a defense-in-depth layer that buffers client `conversation.item.create` items while a response is active and flushes them at response-done, eliminating the whole class of mid-generation interleaving by construction. These targeted fixes remain valuable as layered safety.

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -287,22 +287,39 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
         )
 
     def _record_tool_call(self, state: _GenState, turn: _Turn, item: ResponseFunctionToolCall) -> Iterator[LLMOut]:
-        """Append a tool call to history + the tools list, and emit it (unless the
-        turn went stale, in which case it is recorded but not spoken)."""
+        """Emit a tool call, persisting it (and any assistant text seen so far)
+        to history *before* it is forwarded to the client.
+
+        The function_call must already exist in the conversation by the time the
+        client returns its ``function_call_output``; otherwise a fast client
+        races ahead of the deferred end-of-turn write-back and the output is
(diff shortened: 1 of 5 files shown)


andimarafioti (approved):
nice!

PR #327, opened 2026-06-29
Defer client conversation items during an active response

A-Mahla:
## Why

`conversation.item.create` (tool outputs, images, text) is applied on the asyncio/router thread, while a response *generates* and writes the chat on the pipeline thread. `Chat._lock` makes each operation atomic, but a generation's whole *serialize → generate → write-back* cycle is not — so a client item applied mid-generation races the write-back. This is the **root cause** behind the two races fixed in #326 (tool call not yet recorded when the output returns; image stripped before the next turn reads it).

## What

Serialize all client conversation mutations to response boundaries:

- `ConnState.deferred_items` — items that arrived while `in_response` is set.
- `ConversationHandler.handle_conversation_item_create` — if a response is active, **buffer** the item (defer its `conversation.item.created` ack too); otherwise apply immediately as before.
- `ConversationHandler.flush_deferred_items` — apply buffered items in arrival order.
- `ResponseHandler.finish_response` — flush after `_end_response` (so `in_response` is cleared and the generation's own write-back has already landed).

Because buffering, flushing, and the `in_response` flag are all managed on the single asyncio thread, this adds no new cross-thread race; the pipeline's write-back is already ordered before `finish_response` via the output queue. The flush completes synchronously before `response.done` is sent, so the next `response.create` always sees the applied items.

Result: a `function_call_output` pairs with its now-recorded `function_call`, and an image survives the just-finished response's `strip_images` — by construction, not by per-site defenses.

## Relationship to #326

Complementary. #326 makes the two known consumers robust at the source (eager tool-call recording, consumed-image-id stripping). This PR removes the *whole class* of mid-generation interleaving. Both are worth keeping as layered defense; this PR is additive and independent (touches only the protocol layer).

## Tests
- `TestDeferConversationItemsDuringResponse`: immediate apply when idle; defer while `in_response`; in-order flush on finish; `function_call_output` deferred then pairs cleanly after completion (no `invalid_conversation_item` error).
- Full suite: **561 passing**; ruff + `ruff format` clean; mypy clean.

## Note to reviewers
The `conversation.item.created` ack for a deferred item now arrives at response-done rather than immediately. Compliant clients gate `response.create` on `response.done` (so it's masked), but worth confirming for any client that blocks on the ack mid-turn.

### DIFF
--- a/src/speech_to_speech/api/openai_realtime/handlers/conversation.py
+++ b/src/speech_to_speech/api/openai_realtime/handlers/conversation.py
@@ -37,27 +37,55 @@ class ConversationHandler(RealtimeBaseHandler):
         Items are added to the LLM chat context but do NOT trigger response
         generation on their own.  A subsequent ``response.create`` event is
         required to trigger the model.
-        """
-        events: list[ServerEvent] = []
-        item = event.item
 
+        While a response is generating, the item is *deferred*: applying it now
+        would race the LLM handler's end-of-turn chat write-back, which runs on
+        the pipeline thread (e.g. a ``function_call_output`` arriving before its
+        ``function_call`` is recorded, or an image stripped before the next
+        turn reads it). Deferred items are flushed, in order, once the response
+        completes — see :meth:`flush_deferred_items`.
+        """
+        st = self._state(conn_id)
+        if st.in_response:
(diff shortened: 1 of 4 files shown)


andimarafioti (approved):
Let's gooo

PR #329, opened 2026-07-03
chore: remove deprecated melo option from tts arguments

abhimanyus1997:
Remove the deprecated and archived MeloTTS choice from the ModuleArguments class Literal option list to prevent potential runtime ValueErrors.


andimarafioti (approved):
Thank you!

PR #330, opened 2026-07-03
Refresh README docs

andimarafioti:
## Summary
- Rewrite the README around the current realtime-first default profile and OpenAI-compatible LLM backends.
- Update installation, run-mode, backend, multilingual, CLI, and citation sections to match the packaged CLI and current argument defaults.
- Update docker-compose to invoke the packaged `speech-to-speech` entrypoint in socket mode instead of the old root-level script path.

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file("docker-compose.yml")'`
- `.venv/bin/python -m pytest tests/test_cli_defaults.py`

Note: `docker compose config` was not run because Docker is not installed in this local environment.


docker-compose.yml
```diff
@@ -5,21 +5,22 @@ services:
...
+      - --mode
+      - socket
+      - --recv_host
+      - 0.0.0.0
+      - --send_host
+      - 0.0.0.0
+      - --llm_backend
+      - transformers
+      - --model_name
+      - microsoft/Phi-3-mini-4k-instruct
```
A-Mahla: Do we still want to have `microsoft/Phi-3-mini-4k-instruct` here ?

README.md
```diff
@@ -166,139 +207,139 @@ speech-to-speech \
...
    speech-to-speech --mode websocket --ws_host 0.0.0.0 --ws_port 8765
    ```
 
-2. Connect to the WebSocket server from your client application at `ws://<server-ip>:8765`. The server handles bidirectional audio streaming:
-   - Send raw audio bytes to the server (16kHz, int16, mono)
-   - Receive generated audio bytes from the server
+2. Connect from your client at `ws://<server-ip>:8765`. Send raw audio bytes as 16 kHz, int16, mono PCM and receive generated audio bytes back.
 
-### Local Approach (Mac)
+### TCP Socket
```
A-Mahla, line 255: Maybe add that using it comes with very limited features (such as no interruption, ...)

README.md
```diff
@@ -166,139 +207,139 @@ speech-to-speech \
...
+                    "input": {
+                        "turn_detection": {
+                            "type": "server_vad",
+                            "interrupt_response": True,
+                        }
+                    }
+                },
+            },
+        }
     )
```
A-Mahla, line 311: Nice :)

A-Mahla (approved):
LGTM! just let some nit

PR #333, opened 2026-07-04
Add native Keenable web search for OpenAI-compatible LLM backends

styskin:
## What

Native, server-side web search for the OpenAI-compatible LLM backends (`chat-completions` and `responses-api`), backed by [Keenable](https://docs.keenable.ai) — a low-latency web search API with a keyless free tier.

With `--keenable_web_search`, the handler advertises `web_search` / `fetch_page` tools to the model and executes them **inside the pipeline**: the tool call never reaches the realtime client — the handler runs the search, writes the `function_call` / `function_call_output` pair into the conversation, and re-requests the model in the same turn. Every realtime client (CLI, browser Space, Reachy Mini) gets live-web spoken answers with zero client-side tool handling.

```bash
speech-to-speech --mode realtime --stt parakeet-tdt --llm_backend chat-completions --tts qwen3 \
    --model_name "google/gemma-4-31B-it:cerebras" \
    --responses_api_base_url "https://router.huggingface.co/v1" \
    --responses_api_api_key "$HF_TOKEN" \
    --responses_api_reasoning_effort none --responses_api_stream \
    --keenable_web_search
```

## Why it works in a voice loop

Keenable's `realtime` search mode returns in ~200 ms consistently, and fast LLM inference (e.g. Cerebras, as in the [voice AI blog post](https://huggingface.co/blog/cerebras-gemma4-voice-ai)) makes the extra model round-trip cheap. Measured end-to-end on an L4 Inference Endpoint with the stack above: **speech-end → first spoken words of a web-grounded answer in 1.3 s**, search included:

```
Keenable web_search('current weather in Paris') -> 5 results in 235 ms
Last speech detected to first speech out: 1.296s
```

## Design notes

- **Bounded agentic loop** in `BaseOpenAICompatibleHandler._generate` (`--tool_call_max_rounds`, default 3); the final round forces `tool_choice="none"` so a turn always ends in speech. Cancellation / speculative-turn staleness is re-checked before executing tools and before each round; token usage accumulates across rounds.
- **Client tools take precedence**: a client-registered tool with the same name is neither re-advertised nor intercepted, so existing deployments (e.g. the hf-realtime-voice Space's browser-side `web_search`) keep working unchanged.
- Tool errors return as `{"error": ...}` output so the model recovers conversationally; `fetch_page` has a client-side SSRF guard and retries `live=true` when a page isn't indexed.
- Keyless by default (rate-limited free tier); `--keenable_api_key` / `KEENABLE_API_KEY` for higher limits. No new dependencies (uses `httpx`).

## Tests

22 new unit tests (`tests/test_server_tools.py` with `httpx.MockTransport`, `tests/test_keenable_tool_loop.py` with a scripted OpenAI client): the loop in streaming and non-streaming modes, name-collision precedence, error paths, max-rounds behaviour, usage accounting. `ruff check` / `ruff format` / `mypy` clean; full suite passes.

## Live demo

Identical fork of the hf-realtime-voice Space with search proxied through Keenable, pipeline on an L4 Inference Endpoint: https://huggingface.co/spaces/styskin/keenable-realtime-voice

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,6 +421,39 @@ speech-to-speech \
     --responses_api_stream
 ```
 
+### Live web search with Keenable
+
+The OpenAI-compatible backends (`chat-completions` and `responses-api`) can give the
+assistant native access to the live web via [Keenable](https://docs.keenable.ai): the
+server advertises `web_search` and `fetch_page` tools to the model and executes them
+*inside the pipeline*, feeding results back to the model in the same turn. Clients need
+no tool handling — ask about news, weather, sports or prices and the spoken answer is
+grounded in a fresh search. Fast inference (e.g. Cerebras) is what makes the extra LLM
+round-trip per search comfortable in a voice loop.
+
+```bash
+# The Cerebras voice stack, now with live web answers
+export KEENABLE_API_KEY="keen_..."   # optional: keyless free tier works too
+speech-to-speech \
+    --mode realtime \
+    --stt parakeet-tdt \
+    --llm_backend chat-completions \
+    --tts qwen3 \
+    --model_name "google/gemma-4-31B-it:cerebras" \
+    --responses_api_base_url "https://router.huggingface.co/v1" \
+    --responses_api_api_key "$HF_TOKEN" \
+    --responses_api_reasoning_effort none \
+    --responses_api_stream \
+    --keenable_web_search
+```
+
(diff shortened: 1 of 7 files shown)


andimarafioti:
Closing since it's been a month and this is still on draft.

PR #339, opened 2026-07-10
Strip Markdown syntax before TTS Fixes #338

### LINKED ISSUE #338: LLM responses containing Markdown are read literally by TTS (asterisks, headings, bullets, etc.)
## Problem

## Expected behavior

- `# Heading` → `Heading`

## Steps to reproduce

YvesLoic5:


andimarafioti:
Not sure what the motivation is here -> no description. Closing

PR #340, opened 2026-07-11
Fix Windows installation error for faster-qwen3-tts

mubashirsidiki:
When trying to install the project on Windows using `uv sync`, the installation fails because `qwentts-cpp-python` (which is pulled in by the `ggml` extra) does not have pre-built wheels for Windows.

This results in the following error:
```text
error: Distribution `qwentts-cpp-python==0.3.0` can't be installed because it doesn't have a source distribution or wheel for the current platform
```

This PR splits the `faster-qwen3-tts` dependency to only require the `ggml` extra on Linux. Windows users will now install the base package without the extra, allowing the installation to succeed out of the box while keeping the optimized backend for Linux users. The logic for macOS remains completely untouched.


andimarafioti (approved):
LGTM!

PR #342, opened 2026-07-12
feat: introduce low-latency Supertonic TTS handler

mubashirsidiki:
This PR integrates [Supertonic](https://github.com/supertone-inc/supertonic) as a new, optional real-time Text-to-Speech (TTS) backend for the pipeline. 

**About Supertonic:**
Supertonic is a lightning-fast, on-device multilingual TTS system powered by the ONNX Runtime. Despite its highly compact 99M-parameter open-weight model, it supports seamless synthesis across 31 languages natively. It provides blazingly fast inference with minimal memory footprint, making it incredibly well-suited for our local pipeline.

## Changes Made
- **New TTS Handler (`supertonic_tts_handler.py`)**: Implements the `SupertonicTTSHandler` to stream generated audio. Safely handles pipeline signals (e.g., `EndOfResponse`) and automatically resamples the native 44.1kHz Supertonic output to the pipeline's expected 16kHz format.
- **Dataclasses & Pipeline Registration**: Added `SupertonicTTSHandlerArguments` to expose `--supertonic_voice`, `--supertonic_lang`, and `--supertonic_speed`. Wired these through `s2s_pipeline.py` and `module_arguments.py`.
- **Benchmarking**: Hooked `supertonic` into `scripts/benchmark_tts.py` for latency testing.
- **Dependencies**: Added `supertonic>=1.3.1` to the optional dependencies in `pyproject.toml` so it can be cleanly installed via `pip install speech-to-speech[supertonic]`.
- **Tests & Docs**: Appended the new dataclass to `test_cli_defaults.py` to ensure CI passes, and updated `TTS/README.md` with usage instructions.


andimarafioti:
Thanks for the contribution! Supertonic is a nice fit for the local pipeline and the integration follows the repo conventions well (arguments dataclass, pipeline registration, docs, `test_cli_defaults.py`). A few things need to change before merge though, mainly around the realtime path. Grouped by severity below.

## Blocking

### 1. `speculative_turns` and `cancel_scope` are silently swallowed

`_build_realtime_pipeline_unit` injects `cancel_scope` and `speculative_turns` into every TTS handler's `setup_kwargs` (and this PR correctly adds `supertonic_tts_kw` to that tuple), but `SupertonicTTSHandler.setup()` absorbs both via `**kwargs` and never uses them. Two consequences in realtime mode:

- Stale speculative turns are synthesized and spoken instead of being dropped.
- After an interruption, the handler keeps yielding the already-synthesized buffer.

The handler needs the same guards as `Qwen3TTSHandler.process()`:

```python
def setup(self, should_listen, supertonic_voice="M1", supertonic_lang="na",
          supertonic_speed=1.0, blocksize=512,
          cancel_scope=None, speculative_turns=None, **kwargs):
    self.cancel_scope = cancel_scope
    self.speculative_turns = speculative_turns
    self.blocksize = blocksize
    ...

def process(self, tts_input):
    speculative_turns = getattr(self, "speculative_turns", None)
    if isinstance(tts_input, EndOfResponse):
        if speculative_turns and not speculative_turns.is_latest_after_reopen_grace(
            tts_input.turn_id, tts_input.turn_revision
        ):
            return
        yield AUDIO_RESPONSE_DONE
        return

    if speculative_turns and not speculative_turns.is_latest_after_reopen_grace(
        tts_input.turn_id, tts_input.turn_revision
    ):
        logger.debug("Dropping stale TTS input for turn=%s rev=%s",
                     tts_input.turn_id, tts_input.turn_revision)
        return
    if speculative_turns:
        speculative_turns.commit(tts_input.turn_id, tts_input.turn_revision)

    cancel_gen = self.cancel_scope.generation if self.cancel_scope else None
    # ... synthesize ...
    for i in range(0, n, self.blocksize):
        if cancel_gen is not None and self.cancel_scope.is_stale(cancel_gen):
            logger.info("Supertonic TTS output cancelled (interruption)")
            return
        yield audio_int16[i : i + self.blocksize]
```

Capturing `cancel_gen` before the blocking `synthesize()` call matters: if the user interrupts during synthesis, the whole buffer gets dropped instead of played.

### 2. Final chunk is not block-aligned, so the tail of every utterance is silenced in local mode

`supertonic_tts_handler.py:81` yields the raw tail chunk. `local_audio_streamer.py` runs a fixed-blocksize `sd.Stream` and does `outdata[:] = audio_chunk[:, np.newaxis]` in the callback. A short chunk raises a broadcast `ValueError`, which the bare `except Exception` converts to silence. So the last <32 ms of every sentence is dropped, on every sentence. Pad the tail like the qwen3 handler does:

```python
n = (len(audio_int16) // self.blocksize) * self.blocksize
for i in range(0, n, self.blocksize):
    yield audio_int16[i : i + self.blocksize]
if n < len(audio_int16):
    yield np.pad(audio_int16[n:], (0, self.blocksize - (len(audio_int16) - n)))
```

Related: please take `blocksize` as a setup argument instead of hardcoding `chunk_size = 512`, so it stays in sync with the streamer's block size.

### 3. Int16 conversion can wrap around

`(audio_16k_float * 32767).astype(np.int16)` has no clipping. `scipy.signal.resample_poly` overshoots past ±1.0 near transients, and the unclipped cast wraps, producing loud pops. Match the qwen3 handler:

```python
audio_int16 = np.clip(audio_16k_float * 32768, -32768, 32767).astype(np.int16)
```

## Should fix

### 4. `tts_input.language_code` is ignored

The pipeline runs language detection and populates `TTSInput.language_code`, which the qwen3 path consumes for per-utterance language switching. Here the language is frozen at the CLI value of `--supertonic_lang`. Supertonic's `"na"` mode may cope, but the handler should prefer `tts_input.language_code` when set (mapped to Supertonic's codes if needed) and fall back to the CLI arg. Otherwise multilingual conversations regress relative to qwen3, which would be a shame given 31-language support is Supertonic's headline feature.

### 5. Unrelated `faster-qwen3-tts` change in `pyproject.toml`, and it breaks Windows defaults

The marker split (ggml extra excluded on Windows, plain wheel added) has nothing to do with Supertonic. It also leaves the defaults inconsistent: `qwen3_tts_backend` defaults to `"ggml"` on non-macOS, so a default Windows install after this change gets `faster-qwen3-tts` without the ggml extra and the default pipeline fails at setup. Please drop this change from the PR. If it's fixing a real Windows install issue, open it separately with a Windows-aware backend fallback so we can discuss it on its own merits.

### 6. Dead device wiring in `s2s_pipeline.py`

`overwrite_device_argument` gains `if hasattr(kwargs, "supertonic_device")`, but `SupertonicTTSHandlerArguments` has no such field, so the branch never fires. Either add a device argument (ONNX Runtime execution provider selection would actually be useful here) or remove the block.

## Minor

### 7. Benchmark default handlers

Adding `supertonic` to the default `--handlers` list in `scripts/benchmark_tts.py` makes the benchmark fail out of the box for anyone without the optional dependency installed. Keep it in the help text and choices, but out of the default list.

### 8. Empty-text path

`yield np.zeros(0, dtype=np.int16)` pushes a zero-length chunk downstream; a plain `return` is cleaner (and with the fix in point 2, a zero-length chunk would otherwise get padded into a full block of silence).

### 9. TTFA scales with sentence length

The full utterance is synthesized before the first chunk is yielded. Probably fine given how fast the model is, but if `supertonic` exposes a streaming synthesis API it would be worth using here. If not, no action needed, just noting it.

### 10. Small doc/dependency checks

- Please double check the README claim that models download to `.cache/supertonic3`.


---

Happy to merge once 1 to 3 are addressed; 4 is the one I'd most like to see land with the initial version. Thanks again, this is a solid base.

mubashirsidiki:
Thanks for the thorough review, @andimarafioti. I’ve addressed all of the points in `c51be57`. Here’s a cleaner summary of what changed:

## Fixed

### 1. Cancellation and speculative turns
- `cancel_scope` and `speculative_turns` are now accepted in `setup()` and handled in `process()`.
- Stale speculative turns are dropped.
- `cancel_gen` is captured before the blocking `synthesize()` call, so interrupted generations now drop the full buffer correctly.

### 2. Block-aligned chunking
- Chunked output now uses `self.blocksize` from setup.
- Tail padding now matches the `qwen3` pattern.

### 3. Safe int16 conversion
- Audio conversion now uses:
  ```python
  np.clip(audio * 32768, -32768, 32767).astype(np.int16)
  ```
- This avoids wraparound when values fall outside `[-1, 1]`.

## Also addressed

### 4. Language code handling
- The handler now prefers `tts_input.language_code` when present.
- It falls back to the CLI argument otherwise.

### 5. Windows packaging
- I kept the platform markers because `qwentts-cpp-python` only ships manylinux wheels and does not support Windows.
- I added the base `faster-qwen3-tts` package for Windows so the rest of the pipeline still works there.
- Users can pass `--qwen3_tts_backend torch` to use Qwen3 TTS on Windows.
- If you’d rather, I can split this into a separate PR.

### 6. Dead device wiring
- Removed the unused device override path.

## Minor cleanups

### 7. Benchmark defaults
- Supertonic was removed from the default `--handlers` list.

### 8. Empty text handling
- The empty-text path now returns early instead of yielding a zero-length chunk.

### 9. `rename_args` mismatch
- Fixed the prefix mismatch where the dataclass fields used `supertonic_*` but `rename_args` was using `"supertonic_tts"`.

### 10. Type annotations
- Added type annotations throughout for consistency and better type-checking.

## Notes

- Streaming synthesis is still not available right now because `supertonic.synthesize()` is blocking. No streaming API is available upstream.
- I also confirmed the model download path is `~/.cache/supertonic3/` and updated the README accordingly.
- The voice count is now corrected from `4` to `10`:
  - `M1`-`M5`
  - `F1`-`F5`

Thanks again for the review. Happy to make any further adjustments.

mubashirsidiki:
@andimarafioti  could you please take another look when you get a chance?

andimarafioti (approved):
Thank you for this! Supertonic is a great model and I wanted to add it, you saved me some work :) 
Codex still found some things to complain about, and because the PR has been open for long it diverged from main. I'll fix those issues before merging it but I'll keep you in the git history. Again, thank you !

PR #344, opened 2026-07-13
Strip Markdown syntax before TTS Fixes #338

### LINKED ISSUE #338: LLM responses containing Markdown are read literally by TTS (asterisks, headings, bullets, etc.)

YvesLoic5:
## Summary
LLM responses often contain Markdown (bold, italic, code, headings, bullet lists). Since the TTS engine reads text literally, symbols like `**` or `#` get voiced instead of being stripped.

This adds `remove_markdown` to strip common Markdown syntax while keeping the enclosed text, called from `remove_unspeechable` before the speechable-character filter.

Fixes #338

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -29,7 +29,7 @@ from speech_to_speech.LLM.chat import (
 )
 from speech_to_speech.LLM.compaction_prompt import CompactGenerateFn, build_compactor
 from speech_to_speech.LLM.text_prompt import build_text_system_prompt
-from speech_to_speech.LLM.utils import remove_unspeechable, resolve_auto_language
+from speech_to_speech.LLM.utils import remove_markdown, remove_unspeechable, resolve_auto_language
 from speech_to_speech.LLM.voice_prompt import build_voice_system_prompt
 from speech_to_speech.pipeline.cancel_scope import CancelScope
 from speech_to_speech.pipeline.handler_types import LLMIn, LLMOut
@@ -353,7 +353,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             elif isinstance(event, ToolCall):
                 # Flush any pending spoken text before emitting the tool call.
                 if printable_text.strip():
-                    sentence_batch.append(printable_text.strip())
+                    sentence_batch.append(remove_markdown(printable_text.strip()))
                     printable_text = ""
                 if sentence_batch:
                     if not self._turn_output_allowed(turn.turn_id, turn.turn_revision):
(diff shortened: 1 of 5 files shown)


andimarafioti (changes requested):
Thanks for tackling this! The problem is real, but the fix is at the wrong layer and the regex has some sharp edges.

**Main issue: this runs per streamed delta.** `remove_unspeechable` is called on each `TextDelta` (base_openai_compatible_language_model.py:379), and deltas are token-sized. Paired delimiters split across deltas, so the pattern often has nothing to match. `*italic*` arriving as `*ita` / `lic*` still leaks asterisks to TTS, which is exactly what #338 reports. The right hook is where complete sentences exist, e.g. in `_flush` after `sent_tokenize` (plus the equivalent spots in language_model.py and the responses API path).

**Regex issues** (all reproduced locally):
- `_(.*?)_` mangles snake_case: `function_call_output` -> `functioncalloutput`. Intraword underscores aren't emphasis in CommonMark. Drop it or require word boundaries.
- DOTALL + alternation order breaks bullet lists: in `* first\n* second\n* third` the italic branch pairs asterisks across lines before `^[-*+]\s+` ever fires. Put line-anchored branches first and drop DOTALL.
- Math gets eaten: `2 * 3 * 4` -> `2  3  4`.
- Single pass, so nested markdown leaks: `**bold with `code`**` keeps the backticks.
- Links aren't handled: `[text](url)` passes through and TTS reads the whole URL out loud. Probably the worst offender in practice, would be great to cover it here too.

Also: please add tests. This is pure text-in/text-out and the cases above (snake_case, bullets, math, split deltas, nesting, links) are the exact test list.

One alternative worth considering: for the audio path we don't need semantic markdown parsing. Deleting delimiter characters is streaming-safe because it's character-level instead of pair-level, and with a link rewrite at sentence level it covers everything with much less regex surface.

YvesLoic5:
Fixed all of it:

* Wrong layer: remove_markdown no longer runs per-delta. It now runs only on complete sentences (after sent_tokenize) and on the trailing leftover, never on raw token fragments. remove_unspeechable stays per-delta but now only does character filtering (no pairing risk). Same fix applied to language_model.py, which had the identical bug.
* snake_case: dropped _..._ emphasis entirely, not worth it, essentially never used in LLM prose.
* Bullets/DOTALL: rewrote as 5 independent single-purpose passes (links, headings, bullets, fences, delimiters), no DOTALL, no alternation ordering.
* Math eaten: delimiter is now "touches non-space on at least one side", 2 * 3 * 4 survives.
* Nesting: delimiters deleted per-run, not per-pair, so **bold with code** resolves in one pass.
* Links: [text] (url) → text.
* Tests: added unit tests for every case you listed + 2 integration tests that split a delimiter pair across two real streamed deltas.

Went with your suggested alternative (delete delimiters, don't parse pairs).

andimarafioti (approved):
Sorry for the delay, I'll merge this to another branch to fix a couple of things, but will get it to main today. Thank you!

PR #352, opened 2026-07-15
Add WebRTC transport for the OpenAI Realtime API

A-Mahla:
Supersedes #279 — that branch predates the pipeline-pool rework of the realtime router, so this reimplements the transport on top of it rather than rebasing.

## What this adds

A WebRTC transport alongside the WebSocket endpoint, following the OpenAI GA Realtime handshake:

```
POST /v1/realtime/calls        Content-Type: application/sdp
```

The client POSTs an SDP offer and receives an SDP answer (`201`, `Location: /v1/realtime/calls/{call_id}`). Audio flows over RTP media tracks (Opus 48 kHz, resampled to/from the 16 kHz pipeline rate with a stateful `av.AudioResampler` — no per-chunk boundary artifacts, stereo downmixed correctly); all JSON events use the same protocol as WebSocket mode, carried on the `oai-events` data channel. Requires the new optional `webrtc` extra (aiortc); the endpoint answers `501` with install instructions when it is missing.

## Design

- **The send loop stays the sole consumer of the pipeline output queues.** `SessionState` now holds a `SessionTransport` (`WebSocketTransport` or `WebRTCSession`) instead of a raw websocket; the per-unit send loop hands events and PCM to the transport. All generation-staleness discards, sentinel handling, and SESSION_END drain/release logic remain in one place and work identically for both transports.
- **WebRTC sessions claim pipeline units from the same pool** as WebSocket clients (`503` + `session_limit_reached` when full); `/v1/pool` stays accurate, and disconnect goes through the same drain-and-release path (with a connect watchdog so an abandoned handshake can't hold a unit).
- **Shared client-event dispatch** for both transports. Transport-gated events: `input_audio_buffer.append` is rejected over WebRTC (`invalid_event_for_transport` — audio arrives on the media track), and `output_audio_buffer.clear` is WebRTC-only.
- **Interruption flushes server-side audio.** Over WebRTC the unplayed audio buffers in the server's paced track (20 ms frames), not in the client — so barge-in, `response.cancel`, and `output_audio_buffer.clear` all clear that buffer and playback stops within ~one frame.
- **Shared inbound path**: `append_pcm` factored out of `handle_audio_append`, so WebRTC audio keeps the 512-sample chunk alignment and `input_audio_buffer.commit` bookkeeping. `begin_audio_response` factored out of `encode_audio_chunk`, so the implicit VAD path still emits `response.created` over the data channel without encoding base64 deltas nobody reads.
- **ICE configuration** via `SPEECH_TO_SPEECH_ICE_SERVERS` (JSON list of STUN/TURN entries); aiortc defaults otherwise. NAT-restricted deployments need their own TURN — documented in the realtime README.

## Drive-by fix

`input_audio_buffer.commit` was missing from `_EVENT_TYPE_TO_MODEL`, so the WebSocket route's commit branch was unreachable (commits errored as unknown events). Added to the parse map.

## Tests

- Unit tests: stateful resampler (stereo 48 kHz → mono 16 kHz downmix), paced audio track (wall-clock pacing, silence fill, `clear()`), shared dispatch transport gating, `append_pcm` chunk/remainder/commit behavior.
- Loopback integration test with a real aiortc peer against the uvicorn-served app: SDP handshake, `session.created` + error events over the data channel, mic audio landing on `input_queue` as 512-sample chunks, pipeline audio arriving as 48 kHz RTP frames, `response.created`/`response.done` sequencing, and unit release on hangup.
- Module skips cleanly when aiortc isn't installed; aiortc added to the dev group so CI exercises it. Full suite: 582 passed; ruff and mypy clean.

### DIFF
--- a/demo/rtc/s2s-rtc-client.js
+++ b/demo/rtc/s2s-rtc-client.js
@@ -1,0 +1,809 @@
+// @ts-check
+/**
+ * Minimal WebRTC client for a speech-to-speech realtime endpoint.
+ *
+ * Sibling of `ws/s2s-ws-client.js` with the SAME public surface (constructor
+ * options subset, methods, dispatched events), so `main.js` can pick either
+ * class from the transport setting and wire it identically. Direct mode only:
+ * there is no load-balancer/queue path over WebRTC yet.
+ *
+ * Handshake (OpenAI Realtime GA "calls" endpoint, via the same-origin proxy):
+ *
+ *   1. getUserMedia -> add the mic track + create the `oai-events` data
+ *      channel on an RTCPeerConnection, `createOffer`, wait for ICE gathering.
+ *   2. POST the offer SDP (Content-Type: application/sdp) to `callsUrl`
+ *      (usually the same-origin `api/calls` proxy, which forwards it to the
+ *      s2s server's /v1/realtime/calls) -> 201 + answer SDP.
+ *   3. `setRemoteDescription(answer)`; once the data channel opens the server
+ *      pushes `session.created` and we reply with `session.update`.
+ *
+ * After that the JSON protocol on the data channel is the one the WebSocket
+ * transport speaks, with the audio moved out of it:
+ *
+ *   - Mic audio rides the RTP media track (Opus), NOT
+ *     `input_audio_buffer.append` — the server rejects `append` over WebRTC.
+ *   - Assistant audio arrives as a remote media track, NOT as
+ *     `response.output_audio.delta` events. There is no client playback
+ *     buffer: barge-in flushing happens server-side, so `speech_started`
+ *     only needs to flip the UI state here.
+ *
+ * Because no audio events exist, "the assistant is audibly speaking" is
+ * detected from the output analyser's RMS (with a short hang time), and
+ * `response.done` / `speech_started` remain the authoritative exits — the
+ * same status contract the WS client exposes.
+ *
+ * @typedef {"idle" | "connecting" | "connected" | "user-speaking" |
+ *           "processing" | "ai-speaking" | "closed" | "error"
+ * } RtcStatus
+ *
+ * @typedef {Object} RtcClientOptions
+ * @property {string} callsUrl URL to POST the SDP offer to (same-origin proxy
+ *   like `api/calls`, or a direct `/v1/realtime/calls` URL when CORS allows).
+ * @property {RTCIceServer[]} [iceServers] STUN/TURN servers for the peer
+ *   connection (from /api/config). Empty/absent -> browser defaults (host
+ *   candidates only — fine locally, may not traverse NATs).
+ * @property {string} voice
+ * @property {string} instructions
+ * @property {MediaStream} [micStream] Live mic stream. Provide this OR `acquireMic`.
+ * @property {() => Promise<MediaStream>} [acquireMic] Lazily obtain the mic
+ *   stream once connect() actually runs (same contract as the WS client).
+ * @property {AudioContext} [audioContext] Pre-created (and resumed) context —
+ *   created inside the tap gesture so iOS lets it start.
+ * @property {import("../ws/s2s-ws-client.js").ToolDef[]} [tools] Function tools
+ *   declared in the initial `session.update`.
+ */
+
+import { extractResponseTranscript } from "../ws/codec.js";
+import { OrbVisualiser, VIS_FFT_SIZE } from "../ws/orb-visualizer.js";
+
+// The server's data channel label (aiortc side ignores any other label).
+const DATA_CHANNEL_LABEL = "oai-events";
+// Give up on the handshake if the data channel hasn't opened this long after
+// the SDP answer was applied (ICE failed silently, e.g. NAT without STUN).
+// The server holds its slot behind a 30 s watchdog; stay under it.
+const DC_OPEN_TIMEOUT_MS = 20_000;
+// Don't wait forever for ICE gathering before POSTing the offer — host
+// candidates land near-instantly; STUN answers within a couple of seconds.
+const ICE_GATHERING_TIMEOUT_MS = 3_000;
+// Output-RMS gate for "the assistant is audibly speaking": open above the
+// threshold, and hang on briefly so inter-word gaps don't flap the status.
+const SPEAKING_OPEN_DB = -50;
+const SPEAKING_HANG_MS = 250;
+const LEVEL_POLL_MS = 50;
+
+/** Build an Error carrying a `code` so callers can branch on the failure kind.
+ *  @param {string} message @param {string} code */
+function _codedError(message, code) {
+  const err = /** @type {Error & { code?: string }} */ (new Error(message));
+  err.code = code;
+  return err;
+}
+
+export class S2sRtcRealtimeClient extends EventTarget {
+  /** @param {RtcClientOptions} options */
+  constructor(options) {
+    super();
+    /** @type {RtcClientOptions} */
+    this.options = options;
+    /** @type {import("../ws/s2s-ws-client.js").ToolDef[]} */
+    this._tools = options.tools ?? [];
+    /** @type {(() => Promise<MediaStream>) | null} */
+    this._acquireMic = options.acquireMic ?? null;
+    /** @type {boolean} Set by close() so late async steps unwind quietly. */
+    this._closed = false;
+    /** @type {RTCPeerConnection | null} */
+    this._pc = null;
+    /** @type {RTCDataChannel | null} */
+    this._dc = null;
+    /** @type {AudioContext | null} */
+    this._ctx = null;
+    /** @type {MediaStreamAudioSourceNode | null} */
+    this._micSrc = null;
+    /** @type {MediaStreamAudioSourceNode | null} */
+    this._remoteSrc = null;
+    /** @type {HTMLAudioElement | null} Muted sink for the Chrome quirk: a
+     * remote WebRTC track stays silent in WebAudio unless the stream is ALSO
+     * attached to an <audio> element. Never added to the DOM. */
+    this._quirkAudio = null;
+    /** @type {AnalyserNode | null} */
+    this._micAnalyser = null;
+    /** @type {AnalyserNode | null} */
+    this._outAnalyser = null;
+    /** @type {OrbVisualiser | null} */
+    this._visualiser = null;
+    /** @type {number} Level/status poll timer (setInterval id). */
+    this._levelTimer = 0;
+    /** @type {Uint8Array} Scratch buffer for time-domain RMS reads. */
+    this._levelBuf = new Uint8Array(VIS_FFT_SIZE);
+    /** @type {number} Last time the output RMS was above the speaking gate. */
+    this._lastAudibleAt = 0;
+    /** @type {RtcStatus} */
+    this._status = "idle";
+    this._aiSpeaking = false;
+    /** @type {string} The response currently holding the backend slot (from
+     * response.created), so RMS-detected audio can be attributed to it. */
+    this._activeResponseId = "";
+    /** @type {Set<string>} response_ids that audibly played, so the UI can
+     * tell a barge-in cut (keep it) from a never-heard speculative response
+     * (drop it). Attribution is by "RMS opened while this response was
+     * active" — sound can't be tied to a response id without audio events. */
+    this._audibleResponses = new Set();
+    /** @type {Map<string, string>} CURRENT assistant transcript segment per
+     * response, accumulated from streamed deltas (reset on segment done). */
+    this._asstTranscriptByResp = new Map();
+    /** @type {Map<string, string>} Completed segments per response, joined. */
+    this._asstFullByResp = new Map();
+    this._muted = false;
+    // ── Response lock ────────────────────────────────────────────────────
+    // Same scheme as the WS client: the backend allows ONE response in
+    // flight, so response.create is serialized. `_openResponses` counts
+    // confirmed-but-unfinished responses; `_createInFlight` covers the window
+    // between our create and its response.created echo; extra creates queue.
+    this._openResponses = 0;
+    this._createInFlight = false;
+    /** @type {{ image?: string }[]} */
+    this._createQueue = [];
+    this._sessionConfigured = false;
+    this._debug = (() => { try { return localStorage.getItem("s2s.debug") === "1"; } catch { return false; } })();
+  }
+
+  get status() {
+    return this._status;
+  }
+
+  /** @param {RtcStatus} status */
+  _setStatus(status) {
+    if (this._status === status) return;
+    this._status = status;
+    this.dispatchEvent(new CustomEvent("status", { detail: { status } }));
+  }
+
+  /** Full assistant transcript so far for a response: completed segments plus
+   *  the in-progress one. @param {string} rid @returns {string} */
+  _asstDisplay(rid) {
+    const full = this._asstFullByResp.get(rid) || "";
+    const seg = this._asstTranscriptByResp.get(rid) || "";
+    if (!seg) return full;
+    return full ? `${full} ${seg}` : seg;
+  }
+
+  _markAudible() {
+    if (this._status === "ai-speaking") return;
+    if (this._status === "closed" || this._status === "error") return;
+    this._setStatus("ai-speaking");
+  }
+
+  /**
+   * Full handshake. Resolves once the data channel is open AND the audio
+   * graph is wired (mic track sending, remote track feeding the analysers).
+   * @returns {Promise<void>}
+   */
+  async connect() {
+    if (this._pc) throw new Error("Already connected");
+    this._setStatus("connecting");
+
+    if (!this.options.micStream && this._acquireMic) {
+      this.options.micStream = await this._acquireMic();
+    }
+    if (this._closed) throw _codedError("connect aborted", "aborted");
+
+    this._setupAudioGraph();
+
+    const pc = new RTCPeerConnection(
+      this.options.iceServers?.length ? { iceServers: this.options.iceServers } : {},
+    );
+    this._pc = pc;
+
+    pc.addEventListener("track", (e) => this._onRemoteTrack(e));
+    pc.addEventListener("connectionstatechange", () => this._onConnectionState());
+
+    const micTrack = this.options.micStream?.getAudioTracks()[0];
+    if (!micTrack) throw new Error("No microphone track available");
+    micTrack.enabled = !this._muted;
+    pc.addTrack(micTrack, /** @type {MediaStream} */ (this.options.micStream));
+
+    // The client opens the events channel (OpenAI convention); the server
+    // waits for it and ignores channels with any other label.
+    const dc = pc.createDataChannel(DATA_CHANNEL_LABEL, { ordered: true });
+    this._dc = dc;
+    dc.addEventListener("message", (e) => this._onDcMessage(e.data));
+    dc.addEventListener("close", () => this._onDcClose());
+
+    await pc.setLocalDescription(await pc.createOffer());
+    // No trickle ICE over the one-shot HTTP handshake: the offer must carry
+    // our candidates, so wait for gathering (bounded — host candidates are
+    // near-instant, and a STUN timeout shouldn't stall the connect).
+    await this._waitIceGathering(pc);
+    if (this._closed) throw _codedError("connect aborted", "aborted");
+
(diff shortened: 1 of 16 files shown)


src/speech_to_speech/api/openai_realtime/handlers/audio.py
```diff
@@ -60,7 +60,19 @@ def handle_audio_append(self, conn_id: str, event: InputAudioBufferAppendEvent)
...
+        """Resample raw PCM16 to the pipeline rate and split into 512-sample chunks for the VAD.
+
+        Shared by both transports: the WebSocket route feeds it decoded
+        ``input_audio_buffer.append`` payloads, the WebRTC transport feeds it
+        PCM decoded from inbound media-track frames. Keeps the sub-chunk
+        remainder and the commit bookkeeping (``audio_buffer_has_data``) in
+        one place regardless of how audio arrives.
+        """
+        st = self._state(conn_id)
+        pcm_bytes = resample(pcm_bytes, src_rate, PIPELINE_SAMPLE_RATE)
```
andimarafioti, line 76: Why are you adding this function? it seems to me like functionally this is exactly the same as what we had before.

src/speech_to_speech/api/openai_realtime/transports.py
```diff
@@ -0,0 +1,95 @@
...
+    def discard_pending_audio(self) -> None:
+        """Drop transport-buffered audio that has not reached the client yet.
+
+        WebSocket clients buffer audio on their side, so this is a no-op there;
+        the WebRTC transport paces playback server-side and must flush its
+        track buffer for barge-in to actually silence the assistant.
+        """
+        ...
+
+    async def close(self) -> None: ...
```
andimarafioti: How does this work? It's never instantiated in this PR afai can see.

src/speech_to_speech/api/openai_realtime/transports.py
```diff
@@ -0,0 +1,95 @@
...
+    def discard_pending_audio(self) -> None:
+        # Unplayed audio lives client-side over WebSocket; truncation is the
+        # client's responsibility.
+        pass
+
+    async def close(self) -> None:
+        try:
+            await self.websocket.close()
+        except Exception:  # noqa: BLE001
+            pass
```
andimarafioti, line 100: Is this related to WebRTC?

src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -0,0 +1,349 @@
...
+from aiortc import RTCConfiguration, RTCIceServer, RTCPeerConnection, RTCSessionDescription
+from aiortc.mediastreams import MediaStreamError, MediaStreamTrack
+
+if TYPE_CHECKING:
+    from speech_to_speech.api.openai_realtime.service import RealtimeService, ServerEvent
+
+logger = logging.getLogger(__name__)
+
+PIPELINE_SAMPLE_RATE = 16_000
+WEBRTC_SAMPLE_RATE = 48_000
```
andimarafioti, line 38: why's webrtc by default 48k? Can't we have it match the pipeline sr? We're using 3x the bandwidth otherwise.

andimarafioti (commented):
leaving some questions

andimarafioti:
Couple of things found by codex and claude, they obviously found many things, but these are the only that seem real:

[P1] Barge-in fails after generation finishes but buffered RTP audio is still playing. finish_response() clears response state immediately, so later speech skips the interruption branch and never flushes PipelineAudioTrack. Fast TTS can leave seconds of audible buffered audio.

1. Unit leak between claim and negotiate (the one I'd block on). In webrtc_calls_endpoint, the span from _claim_unit(None) to the try: await session.negotiate(...) is unprotected. If RTCPeerConnection(configuration=...), session.setup() (which calls pc.addTrack), or anything else in that window raises, FastAPI returns a 500 and the unit stays claimed forever with transport=None. The connect watchdog doesn't exist yet at that point, since it's spawned inside negotiate(). Wrap the whole span in try/except and call _release_session on failure, or claim the unit only after the PC and session are constructed.

src/speech_to_speech/api/openai_realtime/handlers/audio.py
A-Mahla: It's `handle_audio_append` split in two, not new logic: the entry point keeps the base64/event handling for the WebSocket route, and append_pcm is the raw-PCM core the WebRTC transport calls with audio decoded from media-track frames (there's no `input_audio_buffer.append` event on that path)

src/speech_to_speech/api/openai_realtime/transports.py
A-Mahla: It's a `typing.Protocol`, a structural interface, so it's never instantiated by design. The send loop and `_dispatch_client_event` are typed against it, and WebSocketTransport (below) and WebRTCSession both satisfy it. mypy checks conformance at the call sites.

A-Mahla: After further thought, to enforce stricter safeguards around the protocol classes, I’ll change them to fully abstract classes.

src/speech_to_speech/api/openai_realtime/transports.py
A-Mahla: Indirectly: lifespan shutdown now closes whatever transport is attached (`sess.transport.close()`) instead of reaching into a raw websocket, since the session may hold either transport kind. This is the same ws-close-and-swallow behavior that used to live inline in the lifespan, just moved behind the interface.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
A-Mahla: 48 kHz is Opus's fixed RTP clock rate (RFC 7587 mandates opus/48000/2 in the SDP regardless of content), and aiortc's Opus codec is hardcoded to 48 kHz stereo (`aiortc/codecs/opus.py`). So feeding it 16 kHz frames would not remove the resample, it would just happen inside aiortc instead/

A-Mahla: Browsers give JS no control over the encoded audio path, and on the Python side aiortc is the only WebRTC stack, even the official openai SDK only handles the SDP exchange (`client.realtime.calls.create(sdp=...)`) and expects the caller to run `aiortc` for the actual connection.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -0,0 +1,349 @@
...
+                self._spawn(self._consume_inbound_audio(track))
+
+        @self._pc.on("connectionstatechange")
+        async def on_connection_state_change() -> None:
+            state = self._pc.connectionState
+            logger.info(f"[WebRTC] Connection state: {state}")
+            if state in ("failed", "closed"):
+                await self.close()
+
+    async def negotiate(self, offer_sdp: str) -> str:
```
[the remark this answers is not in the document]
A-Mahla: It's the standard WebRTC term for what that method does. Establishing a WebRTC connection starts with an SDP negotiation, the offer/answer exchange defined in RFC 3264

src/speech_to_speech/api/openai_realtime/websocket_router.py
```diff
@@ -445,22 +482,145 @@ def _state(u: PipelineUnit) -> dict[str, Any]:
...
+            logger.error(f"WebRTC negotiation failed (session {session_id}): {type(e).__name__}: {e}")
+            await session.close()
+            return Response(content="Invalid SDP offer", status_code=400, media_type="text/plain")
+
+        logger.info(f"WebRTC SDP answer returned (session {session_id})")
+        return Response(
+            content=answer_sdp,
+            status_code=201,
+            media_type="application/sdp",
+            headers={"Location": f"/v1/realtime/calls/{session_id}"},
```
A-Mahla, line 680: **[confirmed] The advertised `Location: /v1/realtime/calls/{id}` URL has no route — spec-conformant hangups leak the pool unit.**

The 201 response advertises the OpenAI GA calls contract, but there is no `DELETE`/`GET` handler for `/v1/realtime/calls/{session_id}`. A client that hangs up by DELETE-ing the Location URL (as the GA spec allows) gets a 404, the call isn't terminated, and the pipeline unit stays claimed — blocking a pool slot (503s for other users) until the client also closes its `RTCPeerConnection` or ICE failure is noticed. Either add the DELETE route (close the pc + release the unit) or drop the Location header so no client is invited to rely on it.

src/speech_to_speech/api/openai_realtime/service.py
```diff
@@ -66,6 +68,8 @@
 
 _EVENT_TYPE_TO_MODEL: dict[str, type[BaseModel]] = {
     "input_audio_buffer.append": InputAudioBufferAppendEvent,
+    "input_audio_buffer.commit": InputAudioBufferCommitEvent,
```
A-Mahla, line 71: **[confirmed] WS-visible behavior change: `input_audio_buffer.commit` and `output_audio_buffer.clear` are newly parseable.**

On main, `parse_client_event` returned `None` for both types, so every WS client got an `unknown_or_invalid_event` error and the router's commit branch was dead code. On this branch the same WS client gets silent success or `input_audio_buffer_commit_empty` for commit (and `st.audio_buffer_has_data` is now actually mutated), and `invalid_event_for_transport` instead of `unknown_or_invalid_event` for clear. These are the only WS behavior deltas found in the whole refactor — everything else in `websocket_router.py` verified behaviorally equivalent. If the WS contract is meant to be untouched, gate these two entries to the webrtc transport (or explicitly call out the WS contract change in the PR description).

demo/rtc/s2s-rtc-client.js
```diff
@@ -0,0 +1,809 @@
...
+      new CustomEvent("error", { detail: { error: new Error("Connection closed by the server") } }),
+    );
+    this._setStatus("error");
+  }
+
+  _onConnectionState() {
+    const state = this._pc?.connectionState;
+    if (this._debug) console.debug(`[rtc] connection state: ${state}`);
+    if (this._closed) return;
+    if (state === "failed" || state === "disconnected") {
```
A-Mahla: **[confirmed] Transient `disconnected` peer state is treated as fatal.**

Browsers routinely fire `connectionState === "disconnected"` on brief packet loss / interface switches and recover to `connected` seconds later on their own; only `failed` is terminal. Treating `disconnected` as an error means a WiFi blip dispatches a fatal error event, main.js tears the session down and releases the backend unit — killing a call the transport would have survived. Suggest: treat only `failed` (and `closed`) as fatal, optionally with a grace timer for `disconnected`.

demo/server.py
```diff
@@ -77,6 +79,46 @@
...
+    raw = raw.strip()
+    if not raw:
+        return []
+    try:
+        data = json.loads(raw)
+        if isinstance(data, list):
+            return data
+    except ValueError:
+        pass
+    return [{"urls": u.strip()} for u in raw.split(",") if u.strip()]
```
A-Mahla, line 103: **[confirmed] `_parse_ice_servers`: a single JSON object falls through to the comma-split fallback and produces garbage ICE URLs.**

`RTC_ICE_SERVERS='{"urls": "turn:...", "username": ...}'` is valid JSON but not a list, so it falls into the comma-split path and each JSON fragment gets wrapped as `{"urls": "{\"urls\": \"turn:..."}`. In the browser `new RTCPeerConnection({ iceServers })` then throws `SyntaxError: not a valid URL` on every connect. Fix: accept a lone JSON object by wrapping it in a list (`[parsed]`), and only comma-split input that doesn't parse as JSON at all.

src/speech_to_speech/api/openai_realtime/websocket_router.py
```diff
@@ -445,22 +482,145 @@ def _state(u: PipelineUnit) -> dict[str, Any]:
...
+                content=build_error_event(
+                    f"All {len(pool)} session slots are in use. Disconnect an existing client first.",
+                    error_type="session_limit_reached",
+                ).model_dump_json(),
+                status_code=503,
+                media_type="application/json",
+            )
+
+        pipeline_log_ctx.set(unit.index)
+        session_id = unit.service.register()
```
A-Mahla: **[plausible] Unprotected window between `_claim_unit()` and the try/finally that releases the unit.**

`unit.service.register()`, the `unit.session` assignment, and `_clean_unit(unit)` run after the claim but outside any exception handler (the WS route protects the equivalent window with try/finally). If any of them raises, the request 500s and the unit stays claimed forever: `unit.session` set, `released_at` None, no SESSION_END, no drain task — `/v1/pool` reports a phantom active session and the pool permanently loses a slot (and the s2s-endpoint LB's stuck/draining detection never fires because the unit never enters draining). Wrap this window in the same try/except-release pattern as the call-setup failure path below.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -0,0 +1,351 @@
...
+        try:
+            await self._pc.close()
+        except Exception as e:  # noqa: BLE001
+            logger.warning(f"[WebRTC] Error closing peer connection: {e}")
+        self._on_closed()
+        logger.info("[WebRTC] Session closed")
+
+    # ── SessionTransport interface ────────────────
+
+    async def send_events(self, events: list[ServerEvent]) -> None:
```
A-Mahla, line 290: **[plausible] Server events are silently dropped while the data channel isn't open, but mic audio can already be flowing.**

RTP can start as soon as ICE connects, before the SCTP data-channel handshake completes. If the user speaks in that window, `speech_started/stopped`, transcription deltas, and response bookkeeping events all hit this early-return and are dropped rather than queued — the first turn's transcript never reaches the client and its state machine desyncs (orb stuck in listening while reply audio plays). Consider buffering events until channel-open and flushing them in order.

src/speech_to_speech/api/openai_realtime/websocket_router.py
```diff
@@ -445,22 +482,145 @@ def _state(u: PipelineUnit) -> dict[str, Any]:
...
+    async def webrtc_calls_endpoint(request: Request) -> Response:
+        """WebRTC SDP handshake (OpenAI GA Realtime 'calls' endpoint).
+
+        The client POSTs an SDP offer with Content-Type: application/sdp and
+        receives an SDP answer. Audio then flows over WebRTC media tracks;
+        events flow over the 'oai-events' data channel using the same JSON
+        protocol as the WebSocket transport.
+        """
+        try:
+            from aiortc import RTCPeerConnection
```
A-Mahla: **[confirmed] First `/v1/realtime/calls` request runs the heavy `aiortc` import synchronously on the shared event loop.**

The first WebRTC handshake pays the aiortc/av/cryptography C-extension import cost (hundreds of ms to >1 s cold) inside the async handler, freezing every unit's send loop and every live WS session for that window — an audible dropout for all connected users. Cheaper: attempt the import once at module import / `create_app` time (keeping the try/except to gate the 501), or run it via `asyncio.to_thread` / `run_in_executor`.

demo/server.py
```diff
@@ -127,6 +169,11 @@ def config():
         # Deploy-pinned direct s2s URL (empty when unset). Not a secret: the
         # browser dials it itself, and Settings shows it locked.
         "s2sUrl": SPEECH_TO_SPEECH_URL,
+        # WebRTC transport availability: the /api/calls proxy only forwards to
+        # the env-pinned URL (never a client-supplied one), so the toggle is
+        # offered exactly when that URL exists.
+        "rtc": bool(SPEECH_TO_SPEECH_URL),
```
A-Mahla, line 177: **[confirmed] `rtc: true` is advertised purely from `SPEECH_TO_SPEECH_URL` being pinned, without checking the backend has the `webrtc` extra.**

A deploy that pins the URL but installs the base package (no aiortc) still shows the WebRTC toggle; selecting it makes every conversation attempt fail with the raw 501 body surfaced through `_postOffer`'s generic non-2xx path, until the user manually switches back to WebSocket. Consider probing the backend once (e.g. an OPTIONS/HEAD on `/v1/realtime/calls`, or caching the first 501) and flipping `rtc` off, or at least mapping the 501 to a friendly "backend has no WebRTC support" message client-side.

demo/rtc/s2s-rtc-client.js
```diff
@@ -0,0 +1,809 @@
...
+    if (this._debug) {
+      const extra = type.startsWith("conversation.item.input_audio_transcription")
+        ? ` item=${event.item_id} ${event.delta ?? event.transcript ?? ""}`
+        : type.startsWith("response.")
+          ? ` resp=${event.response_id ?? event.response?.id ?? ""} status=${event.response?.status ?? ""}`
+          : "";
+      console.debug(`[rtc] ${type}${extra}`);
+    }
+
+    switch (type) {
```
A-Mahla, line 421: **[confirmed] ~250 lines of transport-agnostic protocol machinery are copy-pasted from `ws/s2s-ws-client.js`.**

The server-event switch, `_sendSessionUpdate`, the create queue (`_createInFlight`/`_openResponses`/`_flushQueuedCreate`), the transcript accumulators, and the `conversation_already_has_active_response` recovery are character-identical apart from `[rtc]` log prefixes — and one copy has already diverged (`_codedError` here dropped the `extra` param the WS original has). Every protocol fix now has to land twice or the transports silently diverge. The sharing pattern already exists (`extractResponseTranscript` is imported from `../ws/codec.js`); a shared protocol/session module would leave only mic/playback/dial code per transport.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -0,0 +1,351 @@
...
+from aiortc.mediastreams import MediaStreamError, MediaStreamTrack
+
+from speech_to_speech.api.openai_realtime.transports import SessionTransport
+
+if TYPE_CHECKING:
+    from speech_to_speech.api.openai_realtime.service import RealtimeService, ServerEvent
+
+logger = logging.getLogger(__name__)
+
+PIPELINE_SAMPLE_RATE = 16_000
```
A-Mahla: **[confirmed] Third copy of `PIPELINE_SAMPLE_RATE` — import it from `service.py` instead.**

This re-declares the constant that already exists in `service.py:61` (and `handlers/audio.py:25`), with nothing tying the copies together. The WebRTC transport resamples inbound audio to this local copy while `websocket_router._on_audio` labels the same PCM with `service.PIPELINE_SAMPLE_RATE` — if the pipeline rate ever changes, WebRTC silently keeps 16 kHz: wrong-rate audio into VAD/STT, pitch-shifted TTS out. Only the aiortc import needs to stay lazy, not this constant: `from speech_to_speech.api.openai_realtime.service import PIPELINE_SAMPLE_RATE`.

src/speech_to_speech/api/openai_realtime/websocket_router.py
A-Mahla: Fixed: added `DELETE /v1/realtime/calls/{call_id}` — closes the transport, which releases the unit through the existing idempotent close path; unknown/released ids answer 404. Covered by the new `test_delete_location_hangs_up` loopback test.

demo/rtc/s2s-rtc-client.js
A-Mahla: Fixed: `disconnected` is now treated as transient (logged, ignored); only `failed` dispatches the fatal error.

demo/server.py
A-Mahla: Fixed: a single JSON dict now parses as a one-element list; the comma-split fallback is reached only for non-JSON input.

src/speech_to_speech/api/openai_realtime/websocket_router.py
A-Mahla: Fixed: `register()`/`_clean_unit()` now run inside a try/except that undoes the claim (`unit.session = None`) and returns 500.

src/speech_to_speech/api/openai_realtime/websocket_router.py
A-Mahla: Fixed: the aiortc/webrtc_session import moved to module load behind a `WEBRTC_AVAILABLE` flag; the handler returns 501 off the flag, so the event loop is never blocked mid-conversation.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
A-Mahla: Fixed: `PIPELINE_SAMPLE_RATE` is now imported from `service.py` (no import cycle; only the aiortc import needs to stay gated).

PR #358, opened 2026-07-16
Fix stuck pipeline units in session teardown

A-Mahla:
## Problem

A pipeline unit is only released for new claims after the `SESSION_END` control message travels the whole handler chain back to `output_queue`, where the send loop observes it and sets `session.drained`. A review of this path found several conditions where that never happens, leaving the unit claimed forever — and once every unit in the pool is stuck, all new connections are rejected with `session_limit_reached`:

1. **Barge-in flush could eat SESSION_END.** The send loop's speech-started interrupt flushes `output_queue` preserving only `AUDIO_RESPONSE_DONE`. A trailing `SpeechStartedEvent` (the VAD writes them straight onto `text_output_queue`, possibly mid-teardown) processed while `SESSION_END` was already on `output_queue` would silently discard it — the release task then polls `drained` forever.
2. **Slot leak on setup failure.** `register()` and `_clean_unit()` ran after the claim but *outside* the route's `try/finally`; an exception there left `unit.session` set with no release task ever spawned.
3. **Wedged or dead handler = invisible permanent capacity loss.** The release path waits indefinitely with only a one-shot log warning at 10s; a handler thread that hung (or died — `BaseHandler.run` doesn't emit `PIPELINE_END` on an uncaught error) removed the unit from the pool until restart, kept the dead session registered, and was hard to distinguish from a normal slow drain in `/v1/pool`.
4. **Fire-and-forget release task.** The `asyncio.create_task` result wasn't referenced (asyncio holds tasks weakly), and an exception in `unregister()` would kill the task *after* draining but *before* `unit.session = None`.

## Changes

- `_keep_audio_sentinel` now also preserves `SESSION_END` control messages, so barge-in flushes can't discard the drain signal.
- Session setup runs inside the route's `try`, so the `finally` always enqueues `SESSION_END` and spawns the release task, even when setup fails.
- New `SESSION_END_QUARANTINE_TIMEOUT_S = 180s`: past it, the unit is **quarantined**, not released. Its service session is unregistered immediately — the chat is closed, so late handler output can't mutate or bill the dead conversation — but the unit stays unclaimable until `SESSION_END` actually drains, which is proof the handler chain is clean. This deliberately avoids force-releasing: pipeline output such as transcripts carries no session identity, so handing the unit to a new client while an old handler is still running would leak the previous client's words into the new conversation, and a unit with a dead handler thread must not re-enter the pool as healthy. A permanently wedged unit is surfaced by `/v1/pool` as state `"stuck"` with `stuck_for_s` so operators can act on it. (Per review feedback — an earlier revision force-released after the timeout.)
- `PipelineControlMessage` gained an optional `session_id` tag as defense in depth: the route tags the `SESSION_END` it enqueues and the send loop only sets `drained` on a matching tag (untagged accepted for compatibility), so a stale `SESSION_END` can never satisfy another session's drain wait.
- Release tasks are held in a strong-reference set, and the release runs in a `try/finally` — `unregister()` errors are logged and the unit is still freed.

## Tests

`TestDrainRelease` covers: the flush preserves `SESSION_END`, quarantine keeps the unit unclaimable and rejects new clients while unregistering the dead session, a late drain returns a quarantined unit to the pool, a stale-tagged `SESSION_END` doesn't drain the next session, and a `register()` failure still releases the unit. Full suite: 578 passed, mypy clean.

### DIFF
--- a/src/speech_to_speech/api/openai_realtime/pipeline_unit.py
+++ b/src/speech_to_speech/api/openai_realtime/pipeline_unit.py
@@ -33,6 +33,11 @@ class SessionState(BaseModel):
     # `None` while the client is still active. Used by /v1/pool to surface stuck
     # units (handlers haven't finished propagating SESSION_END).
     released_at: Optional[float] = None
+    # Wall-clock time when the drain wait gave up and quarantined the unit
+    # (SESSION_END_QUARANTINE_TIMEOUT_S elapsed). The unit stays unclaimable —
+    # its handlers may still emit this session's output — until SESSION_END
+    # actually drains. Reported as "stuck" by /v1/pool.
+    quarantined_at: Optional[float] = None
 
 
 class PipelineUnit(BaseModel):
--- a/src/speech_to_speech/api/openai_realtime/websocket_router.py
+++ b/src/speech_to_speech/api/openai_realtime/websocket_router.py
@@ -37,10 +37,16 @@ from speech_to_speech.pipeline.messages import AUDIO_RESPONSE_DONE, PIPELINE_END
 logger = logging.getLogger(__name__)
 MAX_AUDIO_BATCH_BYTES = 6400
 # How long the release path waits for SESSION_END to propagate through the
-# handler chain back to output_queue before clearing unit.session. Tests
-# monkeypatch this to a small value since their fixtures usually skip the
-# real handler chain.
+# handler chain back to output_queue before warning that the unit is stuck.
+# Tests monkeypatch this to a small value since their fixtures usually skip
+# the real handler chain.
 SESSION_END_DRAIN_TIMEOUT_S = 10.0
+# Hard ceiling on the drain wait: past this, the unit is force-released even
+# though SESSION_END never came back (e.g. a wedged handler thread), trading
+# a possible leak of stale output into the next session for restored pool
+# capacity. Stale audio tagged with an old cancel generation is still dropped
+# by the send loop's discard checks.
+SESSION_END_FORCE_RELEASE_TIMEOUT_S = 180.0
 QItem = TypeVar("QItem")
 
 
@@ -73,7 +79,9 @@ async def _send_events(ws: WebSocket, events: list[ServerEvent]) -> None:
 
 
 def _keep_audio_sentinel(item: Any) -> bool:
-    return _is_audio_done(item)
+    # SESSION_END must survive barge-in flushes of output_queue: dropping it
+    # would leave the release path waiting forever for the drain signal.
+    return _is_audio_done(item) or is_control_message(item, SESSION_END.kind)
 
 
 def _keep_user_text_event(item: Any) -> bool:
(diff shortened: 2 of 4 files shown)


andimarafioti (approved):
LGTM, Of the things that GPT found, I think these might be something to address, but feel free to ignore if you think it's too niche:

1. P1: Force release can feed one client’s transcript into the next client’s conversation
After 180 seconds, the release task sets unit.session = None even though a handler may still be processing the previous session. A new client can immediately claim that same unit.
The session_id protection applies only to SESSION_END. It does not protect ordinary pipeline output:
The actual pipeline has eight queues, including several intermediate queues that are not stored on PipelineUnit.
_clean_unit() flushes only four exposed queues, and cannot remove output produced later by a handler that is still running.
A late TranscriptionCompletedEvent carries no session identity.
The send loop dispatches that event using whichever session currently owns the unit.
The service then appends the old transcript to the new session’s chat and enqueues an LLM request.
A realistic sequence is:
Session A disconnects while STT is wedged.
A’s SESSION_END waits behind the in-progress STT operation.
The timeout force-releases the unit.
Session B claims it.
STT finishes A’s audio and emits A’s transcript.
The send loop processes that transcript under B’s session_id.
That is a cross-client data leak, not merely some harmless stale audio.
A late tagged SESSION_END is also unsafe before it reaches the send loop: every BaseHandler invokes on_session_end() without checking the tag, then forwards the message. The send-loop validation therefore happens only after the stale marker has reset downstream handler state. For example, the Kokoro handler resets its voice and rebuilds its pipeline on every SESSION_END.
Recommended direction: on timeout, quarantine and rebuild/restart the entire PipelineUnit rather than returning it to the claimable pool. Alternatively, every work item and side-channel event needs an immutable session epoch, and every handler must reject mismatched items before processing or resetting state. Tagging only the final sentinel is insufficient.
2. P1: A dead handler is returned to the pool without being restarted
The PR explicitly lists dead handler threads as something the timeout handles. It does not actually recover them. The timeout merely clears unit.session, after which _claim_unit() considers the unit healthy and assigns another client.
Handler threads are started once by ThreadManager; there is no monitoring or restart path. A unit containing a dead handler will therefore repeatedly accept clients whose requests cannot traverse the pipeline, then force-release them again.
The timeout should mark the unit unhealthy and trigger replacement or a process restart. It should not advertise the unit as idle.

andimarafioti (approved):
sorry, forgot the dance moves :D

A-Mahla:
Both points were valid, addressed by replacing the force release with a quarantine (08a37e9): past the timeout the session is unregistered (chat closed, no more billing/mutation from late output) but the unit stays unclaimable until SESSION_END actually drains, and `/v1/pool` reports it as `stuck`.

Since `stuck` units no longer report `draining`, the load balancer's drain-restart check would stop seeing them — andimarafioti/s2s-endpoint#48 mirrors the logic there: a `stuck` unit now triggers the endpoint force-restart immediately, which is also the restart/rebuild recovery you suggested. Merge order between the two PRs is safe either way.

PR #368, opened 2026-07-23
Add an LLM proxy for remote LLM backends

A-Mahla:
## What this adds

A client can use the LLM this deployment is already configured with as a plain OpenAI compatible endpoint, fully concurrent with the voice conversation and never interrupted by new speech. With `--enable_llm_proxy` on a remote backend the server mounts the matching path as a passthrough proxy to the configured upstream:

* `POST /v1/chat/completions` under `--llm_backend chat-completions`
* `POST /v1/responses` under `--llm_backend responses-api`

**The auth story changed in the last commit, read this twice: this server no longer authenticates these paths at all.** Access control belongs to the gateway deployed in front of it. Behind s2s-endpoint (andimarafioti/s2s-endpoint#67) the compute replica opens the paths only to the HF token a connected session was created with, so the API key clients pass to the SDK is the session's HF token. It is not the realtime session id anymore, and it was never the id returned by the session creation call.

## Design in brief

* **No auth and no throttling in the engine, by design.** The first version gated on the realtime session id and rate limited per session. Review feedback moved both concerns to s2s-endpoint, where user identity actually lives: the compute replica opens the routes per user (HF token) and throttles per user. Standalone deployments that turn the flag on run an open proxy and should do so only on a trusted network or behind their own gateway; the README says so plainly.
* **Passthrough with normalization, no protocol translation.** The forwarded body is the client body with three server side mutations: `model` is overwritten with the configured model, `store` is forced to false on Responses requests, and `stream_options.include_usage` is injected on streamed chat completions for token accounting. Everything else, including unknown fields, passes through untouched, and the client receives every upstream frame verbatim.
* **Remote backends only.** Local generation (`transformers`, `mlx-lm`) is serialized behind locks, so the proxy answers 501 naming the two backends that work. Off by default; the flag off, a local backend, and the wrong protocol path all answer 501 with the reason. The proxy paths answer 200, 400, 501, or 502 and nothing else.
* **The proxy never touches pipeline units, queues, or cancel scopes**, which is what guarantees the speech response path cannot regress. A request in flight completes even if its session closes mid generation.
* **Observability.** The usage endpoint gains an additive `llm_proxy` section: request totals, status classes with 429 in its own bucket, and input and output token totals parsed from bodies, streamed usage frames, and Responses completion events.
* **Security posture, stated plainly.** The upstream API key never reaches clients. With the flag on and nothing in front, the proxy is open and unthrottled; the supported gated deployment is behind the s2s-endpoint compute replica.

## Commits

One commit per ticket, each green on its own, plus the rework:

1. Flag, session gate, non streaming chat completions passthrough, 401/501 contract, README
2. Streaming SSE forwarded verbatim as it arrives, connect timeout, in flight survives session close
3. Responses path with `store` forced false
4. Per session rate limit
5. `llm_proxy` usage section
6. Remove auth from the LLM proxy: access control moves to the deployment gateway (this is the rework commit; it removes the session gate and the rate limiter from 1 and 4, `session.created` keeps carrying `session.id` as plain protocol correctness)
7. Read the LLM proxy upstream settings from the renamed argument fields (bug found in manual testing, also review P1: `rename_args` strips the `responses_api_` prefix before the proxy config is built, so the config silently read the dataclass defaults and hit the OpenAI default upstream with no key; the proxy now reads the same renamed fields the pipeline LM reads, with a regression test through the rename)
8. Decode compressed upstream bytes before forwarding LLM proxy streams (review P2: the forwarded response carries no `Content-Encoding`, so a gzip upstream stream reached clients as unreadable compressed bytes and broke token accounting; the stream is now decoded once at the proxy, with a gzip regression test)
9. Drop the bearer token mention from the session id comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,27 @@ with client.realtime.connect(model="local") as conn:
 
 The server implements the core Realtime event set: `input_audio_buffer.append`, `session.update`, `conversation.item.create`, `response.create`, and `response.cancel` inbound; speech start/stop, streaming transcription, audio deltas, tool calls, and `response.done` outbound. The full event reference, architecture, and design details live in the [Realtime Engine README](./src/speech_to_speech/api/openai_realtime/README.md).
 
+### LLM Proxy
+
+With `--enable_llm_proxy`, the realtime server also exposes the remote LLM it is configured with as a plain OpenAI compatible endpoint, so a client holding an open realtime session can run side tasks (summaries, titles, background agents) with tools and streaming, fully concurrent with the voice conversation and never interrupted by new speech:
+
+* `POST /v1/chat/completions` when running `--llm_backend chat-completions`
+* `POST /v1/responses` when running `--llm_backend responses-api`
+
+The bearer token is the realtime session id. **Read this twice: the session id arrives in the `session.created` event on the realtime connection** (behind a load balancer it is not the session id returned by the session creation call). Point the stock OpenAI SDK at the server and pass the session id as the API key:
+
+```python
+from openai import OpenAI
+
+llm = OpenAI(base_url="http://localhost:8765/v1", api_key=session_id)
+completion = llm.chat.completions.create(
+    model="anything",  # ignored: the server forces its configured --model_name
+    messages=[{"role": "user", "content": "Summarize the conversation so far: ..."}],
+)
+```
+
+Requests are stateless (send the full message list each time) and are proxied to the configured upstream with the key held by the server, which never reaches clients. The `model` field is always overwritten with the server configured `--model_name`. The proxy is off by default, requires a remote backend (`chat-completions` or `responses-api`), and answers 501 with the reason otherwise; a missing or unknown session id answers 401.
+
 ## LLM Backends
 
 The LLM is the most compute-intensive and highest-latency component in the pipeline. A single forward pass through a large model can dominate end-to-end response time, so choosing the right backend for your hardware and latency budget matters. The pipeline supports:
--- a/README.md
+++ b/README.md
@@ -319,24 +319,24 @@ The server implements the core Realtime event set: `input_audio_buffer.append`, 
 
 ### LLM Proxy
 
-With `--enable_llm_proxy`, the realtime server also exposes the remote LLM it is configured with as a plain OpenAI compatible endpoint, so a client holding an open realtime session can run side tasks (summaries, titles, background agents) with tools and streaming, fully concurrent with the voice conversation and never interrupted by new speech:
+With `--enable_llm_proxy`, the realtime server also exposes the remote LLM it is configured with as a plain OpenAI compatible endpoint, so a client can run side tasks (summaries, titles, background agents) with tools and streaming, fully concurrent with the voice conversation and never interrupted by new speech:
 
 * `POST /v1/chat/completions` when running `--llm_backend chat-completions`
 * `POST /v1/responses` when running `--llm_backend responses-api`
 
-The bearer token is the realtime session id. **Read this twice: the session id arrives in the `session.created` event on the realtime connection** (behind a load balancer it is not the session id returned by the session creation call). Point the stock OpenAI SDK at the server and pass the session id as the API key:
+The server performs no authentication and no throttling of its own. Enable the proxy only on a trusted network, or deploy the server behind a gateway that owns access control. The s2s-endpoint compute replica is such a gateway: it opens these paths only to clients that created their session with an HF token, checks the API key against that token, and applies a rate limit per user. Point the stock OpenAI SDK at whichever host you talk to; this server ignores the API key (a gateway in front decides what it must be):
 
 ```python
 from openai import OpenAI
 
-llm = OpenAI(base_url="http://localhost:8765/v1", api_key=session_id)
+llm = OpenAI(base_url="http://localhost:8765/v1", api_key="unused")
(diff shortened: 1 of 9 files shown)


andimarafioti:
Couple of questions here:
Can we set `--enable_llm_proxy` independetly per pipeline when we connect a user to it, is it set at startup and done? It would be interesting to only enable this for logged in users for example, but then I imagine that we would need a way to set/unset it.
Is this endpoint tracked on the `usage` we had? I would be more keen on tracking how much this endpoint is pinged, which after being pinged by logged in users, we would be able to track clearly who's using what and how much. Maybe we could then also set some limits.

A-Mahla:
Good questions, and they ended up reshaping the design. The gate is no longer in this repo at all: this server now does no auth and no throttling on the proxy paths, and s2s-endpoint owns access control where user identity actually lives (andimarafioti/s2s-endpoint#67).

On enabling per user: it now works that way by construction. The load balancer fingerprints the HF token a session is created with (HMAC, no raw token stored) and embeds the fingerprint in the signed session token. The compute replica opens the LLM routes only to an API key matching a currently connected session's fingerprint, so logged in sessions get access and anonymous ones get 401, with nothing to set or unset at runtime. The flag here stays a startup switch for whether the deployment offers the feature at all.

On usage: the engine already tracks this endpoint in `/v1/usage` under an additive `llm_proxy` section since commit 5 (request totals, status classes, input and output tokens). Throttling per user lives at the replica (`LLM_PROXY_REQUESTS_PER_MINUTE`, default 20). Attributing usage to individual users on the dashboard is a follow up: the LB can compute the dashboard fingerprint and the LLM fingerprint side by side at session creation and join them.

andimarafioti:
I’d request changes. Three actionable issues:

- **[P1] Proxy credentials and base URL are always discarded.** [`prepare_all_args()` renames](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/s2s_pipeline.py#L320-L327) `responses_api_base_url/api_key` to `base_url/api_key`, but the new configuration reads the old attributes [[here](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/s2s_pipeline.py#L669-L675)](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/s2s_pipeline.py#L669-L675). They resolve to their class defaults, `None`. Consequently the proxy uses `api.openai.com` and `Bearer None`, ignoring the configured provider and potentially sending prompts to the wrong host. Use the renamed/effective values and preserve the SDK’s environment-variable fallback.

- **[P2] Compressed streaming responses become unreadable.** The client negotiates gzip, `_forward_and_account()` uses [`[aiter_raw()](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/api/openai_realtime/llm_proxy.py#L276-L285)`](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/api/openai_realtime/llm_proxy.py#L276-L285), but the downstream [`[StreamingResponse](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/api/openai_realtime/llm_proxy.py#L264-L268)`](https://github.com/huggingface/speech-to-speech/blob/2971c1b8022c1adb3e161b24eb4c97384ad89c88/src/speech_to_speech/api/openai_realtime/llm_proxy.py#L264-L268) drops `Content-Encoding`. My reproduction received gzip bytes without a gzip header. Decode with `aiter_bytes()` or correctly propagate encoding while keeping accounting functional.

Validation: all **652 tests passed** against the PR merged with current `main`, and all six GitHub checks are green; the failures above are uncovered integration/edge cases. The PR title and `session.py` bearer-token comment should also be updated now that session-gating was removed. I did not post a GitHub review or modify the worktree.

A-Mahla:
All three are addressed, thanks for the thorough pass.

* **P1, fixed in 2720768.** Manual testing against the HF router hit the same thing: the config read the raw `responses_api_*` names after `rename_args` had popped them, fell back to the dataclass class defaults, and the proxy went to `api.openai.com` with `Bearer None`. The config is now built by `build_llm_proxy_config()` from the same renamed fields the pipeline LM reads, with a regression test that runs real argument objects through the rename. Verified live against `router.huggingface.co` on both backends. One deliberate deviation from the suggestion: the proxy does not mirror the SDK's environment variable fallback. It uses exactly the values the backend flags carry, so a deployment that relies on ambient `OPENAI_*` variables fails loudly at the provider instead of the proxy silently picking up credentials the operator never pointed it at.
* **P2, fixed in a3a4e62.** The stream is now read with `aiter_bytes()` instead of `aiter_raw()`: the forwarded response carries no `Content-Encoding`, so the bytes are decoded once at the proxy. This also fixes token accounting for compressed upstreams, which was silently parsing gzip bytes as SSE. An uncompressed stream still passes through byte identical. Regression test with a gzip upstream included.
* **Title and comment, done in aeff493.** The PR title no longer says session gated and the `session.py` comment no longer calls the session id a bearer token.

PR #370, opened 2026-07-24
fix: RuntimeError: Cannot execute emptyCache() without MPS backend.

### LINKED ISSUE #145: RuntimeError: Cannot execute emptyCache() without MPS backend.
2024-11-29 11:40:37,152 - STT.paraformer_handler - DEBUG - infering paraformer...
  0%|                                                                                                                                                                       | 0/1 [00:00<?, ?it/s]/root/anaconda3/envs/speechtospeech/lib/python3.10/site-packages/funasr/models/paraformer/model.py:251: FutureWarning: `torch.cuda.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cuda', args...)` instead.
  with autocast(False):
rtf_avg: 0.280: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  2.26it/s]
Exception in thread Thread-5 (run):
  File "/root/anaconda3/envs/speechtospeech/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/root/anaconda3/envs/speechtospeech/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/ai/app/voicechat/speech-to-speech/baseHandler.py", line 37, in run
    for output in self.process(input):
  File "/ai/app/voicechat/speech-to-speech/STT/paraformer_handler.py", line 56, in process
    torch.mps.empty_cache()
  File "/root/anaconda3/envs/speechtospeech/lib/python3.10/site-packages/torch/mps/__init__.py", line 85, in empty_cache
    torch._C._mps_emptyCache()
RuntimeError: Cannot execute emptyCache() without MPS backend.
^CException ignored in: <module 'threading' from '/root/anaconda3/envs/speechtospeech/lib/python3.10/threading.py'>
  File "/root/anaconda3/envs/speechtospeech/lib/python3.10/threading.py", line 1567, in _shutdown
    lock.acquire()
KeyboardInterrupt:

thiagobarbosa:
## Summary

FIx #145 

File `torch.mps.empty_cache()` in `src/speech_to_speech/STT/paraformer_handler.py:62` was called unconditionally after every STT inference, but MPS is only available on Apple Silicon Macs — on CUDA/CPU systems this raised `RuntimeError: Cannot execute emptyCache() without MPS backend.`

**Change:** guarded the call with `if self.device == "mps":`, matching the exact pattern already used for the same purpose in `src/speech_to_speech/LLM/language_model.py:719`.

**Tests:** updated `tests/test_paraformer_transcription_events.py` to set `handler.device` in the test fixture (previously unset, since `torch.mps.empty_cache` was fully monkeypatched to a no-op) and added `test_process_does_not_touch_mps_cache_on_non_mps_device`, which patches `empty_cache` to raise the exact `RuntimeError` from the bug report and asserts `process()` with `device="cuda"` doesn't trigger it. I verified this test fails against the pre-fix code (reproducing the reported traceback) and passes with the fix. Full suite: 598 passed.

### DIFF
--- a/src/speech_to_speech/STT/paraformer_handler.py
+++ b/src/speech_to_speech/STT/paraformer_handler.py
@@ -59,7 +59,8 @@ class ParaformerSTTHandler(BaseSTTHandler):
         logger.debug("infering paraformer...")
 
         pred_text = self.model.generate(vad_audio.audio)[0]["text"].strip().replace(" ", "")
-        torch.mps.empty_cache()
+        if self.device == "mps":
+            torch.mps.empty_cache()
 
         logger.debug("finished paraformer inference")
         console.print(f"[yellow]USER: {pred_text}")
--- a/tests/test_paraformer_transcription_events.py
+++ b/tests/test_paraformer_transcription_events.py
@@ -10,9 +10,10 @@ class _FakeParaformerModel:
         return [{"text": " 今 天 天 气 不 错 "}]
 
 
-def _handler():
+def _handler(device="mps"):
     handler = object.__new__(ParaformerSTTHandler)
     handler.model = _FakeParaformerModel()
+    handler.device = device
     return handler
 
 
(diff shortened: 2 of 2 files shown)


Hotragn:
Drive-by note from another contributor, not a maintainer — this looks superseded, so flagging it rather than leaving it to age.

`044dca3` ("Fix(stt): Guard `torch.mps.empty_cache()` when not on MPS", #513) landed the same guard on `main`:

```python
# src/speech_to_speech/STT/paraformer_handler.py:61
if self.device == "mps":
    torch.mps.empty_cache()
```

It also added an equivalent regression test, so the coverage half of this PR is in too:

```
tests/test_paraformer_transcription_events.py:113
    test_paraformer_skips_mps_cache_clear_on_cpu
```

Verified on `main` at `16d7f98` — 9 passed for that file.

Your diagnosis was right and it predates #513 by about a month; it just wasn't the branch that landed. Two things that are *not* covered, if you wanted to carry this forward:

- `LLM/language_model.py:891` and `:1076` still call `torch.mps.empty_cache()` unguarded. Both sit in the MLX branch, which is selected by `backend`, not `device` — so `--llm_backend mlx-lm` with a non-MPS `--llm_device` reaches them. Their sibling transformers branches (`:907`, `:1102`) do guard, so it reads as an oversight rather than intent.
- I looked at that myself and didn't file it, because on Apple Silicon torch ships MPS so the call succeeds regardless of `self.device` — it's an inconsistency more than a live crash. Worth confirming a real failing configuration before spending effort on it.

Either way this PR as-is is a no-op against current `main`.

PR #374, opened 2026-07-28
Fix faster-whisper adapt_gen_kwargs consuming its caller's config dict

LHMQ878:
### The bug

`FasterWhisperSTTHandler.adapt_gen_kwargs` translated the pipeline's `return_timestamps` into faster-whisper's inverse `without_timestamps` by popping from its argument:

```python
def adapt_gen_kwargs(self, gen_kwargs):
    gen_kwargs["without_timestamps"] = not gen_kwargs.pop("return_timestamps", True)
    return gen_kwargs
```

The `pop` consumes the source key out of the caller's dict, which makes the translation single-use. Three consequences, all silent:

1. **The second read flips the setting.** Pass one config dict to two handlers — the first call consumes `return_timestamps`, so the second call falls back to the `True` default and produces `without_timestamps=False`. A user who asked for `return_timestamps=False` gets timestamps enabled on the second handler.

   ```python
   config = {"return_timestamps": False}
   h.adapt_gen_kwargs(config)   # -> {'without_timestamps': True}   correct
   h.adapt_gen_kwargs(config)   # -> {'without_timestamps': False}  inverted
   ```

2. **The caller's dict is consumed**, so any other code that reads `config["return_timestamps"]` afterwards sees it gone.

3. **The shared `gen_kwargs = {}` default of `setup` is polluted.** `setup` passes its default straight in, so after one no-argument `setup` the class default is permanently `{"without_timestamps": False}`, and every later handler inherits it.

### The fix

Build a new dict instead of mutating the argument:

```python
adapted = dict(gen_kwargs)
adapted["without_timestamps"] = not adapted.pop("return_timestamps", True)
return adapted
```

The `pop` is now on the copy, so it still strips the key faster-whisper does not accept from what is forwarded to `transcribe`, while leaving the caller's dict alone. Also added a docstring, since the inversion is not obvious from the code.

### Scope

Part of a two-site sweep: I grepped every in-place write to a `gen_kwargs` argument across `src/` and found exactly two, this one and `whisper_stt_handler.py` (separate PR, different shape — aliasing rather than popping). No third instance exists.

### Tests

New `tests/test_faster_whisper_gen_kwargs_adaptation.py`, 7 tests behind `pytest.importorskip("faster_whisper")`, covering the inversion in both directions, the default, that the caller's dict survives, that other kwargs pass through, that the shared `setup` default stays empty, and the sharpest one:

```python
def test_one_config_reused_gives_both_handlers_the_same_setting(handler):
    config = {"return_timestamps": False}
    first = handler.adapt_gen_kwargs(config)
    second = handler.adapt_gen_kwargs(config)
    assert first == second == {"without_timestamps": True}
```

Control experiment: with the fix reverted, **3 failed / 4 passed**, one failure per consequence above.

`ruff check` + `ruff format --check` clean at the pinned 0.15.20, `mypy src/` clean. Full suite: this branch 15 failed / 565 passed vs. 16 failed / 557 passed on `main`. The one differing test, `test_speculative_turns.py::test_try_is_latest_after_reopen_grace_reports_pending_without_blocking`, passes 3/3 in isolation — it is a timing flake, unrelated to this change.

### DIFF
--- a/src/speech_to_speech/STT/faster_whisper_handler.py
+++ b/src/speech_to_speech/STT/faster_whisper_handler.py
@@ -63,6 +63,15 @@ class FasterWhisperSTTHandler(BaseSTTHandler):
         del self.model
 
     def adapt_gen_kwargs(self, gen_kwargs: dict[str, Any]) -> dict[str, Any]:
-        gen_kwargs["without_timestamps"] = not gen_kwargs.pop("return_timestamps", True)
+        """Translate the pipeline's `return_timestamps` into faster-whisper's inverse.
 
-        return gen_kwargs
+        Builds a new dict rather than popping from the argument. Popping consumes
+        the source key, so passing the same config to a second handler would read
+        the default instead of the caller's value and flip the setting; it also
+        mutates the shared `gen_kwargs = {}` default of `setup`, leaving
+        `without_timestamps` behind for every later handler.
+        """
+        adapted = dict(gen_kwargs)
+        adapted["without_timestamps"] = not adapted.pop("return_timestamps", True)
+
+        return adapted
--- a/tests/test_faster_whisper_gen_kwargs_adaptation.py
+++ b/tests/test_faster_whisper_gen_kwargs_adaptation.py
@@ -1,0 +1,72 @@
+"""`adapt_gen_kwargs` must translate its argument without consuming it.
+
+It maps the pipeline's `return_timestamps` onto faster-whisper's inverse
+`without_timestamps`. Popping the source key out of the caller's dict makes the
+translation single-use: the second handler built from the same config reads the
+default instead of the supplied value, and the setting silently inverts.
+"""
+
+from typing import Any
+
+import pytest
+
+faster_whisper = pytest.importorskip("faster_whisper")
+
+from speech_to_speech.STT.faster_whisper_handler import FasterWhisperSTTHandler  # noqa: E402
+
+
+@pytest.fixture
+def handler() -> FasterWhisperSTTHandler:
+    """A handler instance without running `setup`, which loads a model."""
+    return object.__new__(FasterWhisperSTTHandler)
+
+
+@pytest.mark.parametrize(
+    "return_timestamps,expected",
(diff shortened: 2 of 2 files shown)


Hotragn:
Drive-by review from another contributor, not a maintainer. I touched the caller of this function in #424, so flagging the overlap — the short version is that the reported symptom is already fixed on `main`, but this PR still does something #424 didn't.

**What's already fixed.** `setup()` now normalizes a copy before calling the helper:

```python
# src/speech_to_speech/STT/faster_whisper_handler.py:52
# adapt_gen_kwargs() mutates and returns what it is given, so normalize a copy:
gen_kwargs = self.adapt_gen_kwargs(dict(gen_kwargs))
```

So the user-visible bug in your description — passing the same config to a second handler and having `return_timestamps` silently invert — no longer reproduces through `setup()`, which is the only production caller. There's a regression test for it (`test_faster_whisper_language_reporting.py::test_setup_does_not_mutate_the_caller_gen_kwargs`).

**What this PR still adds.** `adapt_gen_kwargs` itself remains destructive on `main`:

```python
def adapt_gen_kwargs(self, gen_kwargs: dict[str, Any]) -> dict[str, Any]:
    gen_kwargs["without_timestamps"] = not gen_kwargs.pop("return_timestamps", True)
    return gen_kwargs
```

It's a public method on the handler, and the fix is currently at the call site rather than in the function — so the next caller re-introduces the bug for free. Your version makes the translation pure, which is the more robust place for it, and the docstring explains *why* the pop is wrong. That reads to me as worth keeping on its own merits.

**Suggestion if you want to revive it:** rebase and reframe it as "make `adapt_gen_kwargs` non-mutating" rather than a bug fix, since the caller-consumption symptom is gone. The `dict(gen_kwargs)` in `setup()` then becomes redundant and could be dropped in the same PR, leaving one guarantee in one place instead of two overlapping ones. Worth noting it'll conflict with `setup()` as it stands now.

Apologies that #424 landed across this — I didn't spot your PR until after. Maintainer's call on whether the remaining hardening is worth a rebase.

PR #376, opened 2026-07-28
Add mic/speaker device selectors to the demo Settings UI.

dcolley:
Lets users pick capture and playback devices (persisted in localStorage); speaker routing uses AudioContext.setSinkId where supported.


andimarafioti (approved):
Great work, thank you !

PR #378, opened 2026-07-29
Fix Whisper STT language detection reading a text token

### LINKED ISSUE #375: Whisper STT: language detection reads a text token, silently mistranscribing non-English speech (German → Italian)
## Summary

`WhisperSTTHandler.process()` infers the detected language by positionally slicing the second
generated token:

```python
pred_ids = self.model.generate(input_features, **self.gen_kwargs)
language_code = self.processor.tokenizer.decode(pred_ids[0, 1])[2:-2]  # remove "<|" and "|>"
```

This assumes `generate()` returns the forced decoder prefix
(`<|startoftranscript|><|de|><|transcribe|>…`). **Recent `transformers` no longer includes that
prefix in the returned sequence**, so `pred_ids[0, 1]` is an ordinary *text* token. Slicing it with
`[2:-2]` produces a fragment of a word, which is never in `SUPPORTED_LANGUAGES`, so the handler
takes the "unsupported language" branch and **re-generates with `language=self.last_language`**.

When `last_language` is `None` (the default unless `--language` is set to a non-`auto` value), the
re-generation runs with **no language forced**, Whisper auto-detects, and the *correct* first-pass
transcription is **discarded and replaced** by the mis-detected one.

Observed effect: German speech is returned as Italian, including degenerate loops.

| spoken (de) | returned |
| --- | --- |
| "Der blaue Ball liegt auf dem Tisch." | "Il bambino è il bambino sul tischo." |
| "Ich heiße Max. Bitte merke dir das." | "Io chiamo Max, per te lo che ti è il nostro." |
| "Ich habe ein bisschen Angst vor morgen." | "Mi ha un po' di un po' di un po' di …" (loops) |
| "Wie heiße ich?" | "come si?" |

This is **not** flaky: the parse fails on *every* call with this `transformers` version, and it
fails in **both** forced-language and auto-detect mode, so language detection is effectively broken
for any non-English user of the transformers-Whisper STT backend. Possibly the same root cause as
#79 ("Chinese is not recognized with the default model").

## Reproduction

Minimal, no `speech-to-speech` needed — it isolates the parsing step:

```python
import soundfile as sf, torch
from transformers import AutoProcessor, AutoModelForSpeechSeq2Seq

MID = "openai/whisper-large-v3-turbo"
proc = AutoProcessor.from_pretrained(MID)
model = AutoModelForSpeechSeq2Seq.from_pretrained(MID)
audio, sr = sf.read("german_utterance.wav", dtype="float32")   # any de utterance, 16 kHz mono
feats = proc(audio, sampling_rate=16000, return_tensors="pt").input_features

ids = model.generate(feats, language="de", task="transcribe", max_new_tokens=25)
print("text          :", proc.batch_decode(ids, skip_special_tokens=True)[0])
print("ids[0,1] raw  :", repr(proc.tokenizer.decode(ids[0, 1])))
print("handler parse :", repr(proc.tokenizer.decode(ids[0, 1])[2:-2]))
```

Actual output (transformers 4.57.3, whisper-large-v3-turbo):

```
text          :  Wie heiße ich?
ids[0,1] raw  : ' heiß'
handler parse : 'e'          # <- fed to `language_code not in SUPPORTED_LANGUAGES`
```

`ids[0, 1]` is the word "heiß", not `<|de|>`. Identical result with and without `language="de"`.

## Suggested fix

Stop inferring the language from a token *position*. Scan the leading tokens for an actual language
special token, and fall back to the explicitly requested language:

```python
import re

_LANG_TOKEN = re.compile(r"<\|([a-z]{2,3})\|>")

def _detected_language(self, pred_ids) -> str | None:
    """Language token is only present when generate() returns the forced decoder prefix, which
    newer transformers strips — so scan for it instead of assuming index 1."""
    tok = self.processor.tokenizer
    for tid in pred_ids[0][:4].tolist():
        m = _LANG_TOKEN.fullmatch(tok.convert_ids_to_tokens(tid) or "")
        if m:
            return m.group(1)
    return None
```

and in `process()`:

```python
language_code = self._detected_language(pred_ids) or self.gen_kwargs.get("language") or self.last_language
if language_code not in SUPPORTED_LANGUAGES and not self.gen_kwargs.get("language"):
    ...  # existing re-generate path, only when nothing was forced
```

The important behavioural guarantee: **when a language is explicitly forced, never re-generate
without it** — the first pass is authoritative, and today's fallback throws away a correct
transcription.

Happy to open a PR along these lines if the approach looks right.

## Environment

- `speech-to-speech` 0.2.10 (also present in 0.2.11 — `STT/whisper_stt_handler.py` is byte-identical between the two tags)
- `transformers` 4.57.3, `torch` 2.11.0+cu128
- `openai/whisper-large-v3-turbo`, CUDA (NVIDIA GB10, aarch64) — also reproduces on CPU
- Backend: `--stt whisper` (the transformers path, not faster-whisper)

Hotragn:
Fixes #375

## Problem

`WhisperSTTHandler.process()` inferred the detected language by slicing the second generated token:

```python
pred_ids = self.model.generate(input_features, **self.gen_kwargs)
language_code = self.processor.tokenizer.decode(pred_ids[0, 1])[2:-2]  # remove "<|" and "|>"
```

That assumes `generate()` echoes the forced decoder prefix (`<|startoftranscript|><|de|><|transcribe|>`). Recent `transformers` releases do not include that prefix in the returned sequence, so `pred_ids[0, 1]` is an ordinary **text** token and the `[2:-2]` slice yields a word fragment — `" heiß"` becomes `"e"`.

That fragment is never in `SUPPORTED_LANGUAGES`, so **every** call took the "unsupported language" branch and re-generated with `language=self.last_language`. When `last_language` is unset (the default unless `--language` is given a non-`auto` value) the re-generation runs with no language forced, so the correct first-pass transcription is discarded and replaced by an auto-detected one.

Observed effect: German speech returned as Italian, including degenerate loops.

| spoken (de) | returned |
| --- | --- |
| "Der blaue Ball liegt auf dem Tisch." | "Il bambino è il bambino sul tischo." |
| "Ich heiße Max. Bitte merke dir das." | "Io chiamo Max, per te lo che ti è il nostro." |
| "Wie heiße ich?" | "come si?" |

This is not flaky — the parse fails on every call, in both forced-language and auto-detect mode, so language detection is effectively broken for any non-English user of the transformers-Whisper backend. The bad code also propagates: `language_code` flows through `resolve_auto_language()` into the LLM prompt, so a mis-parse also tells the model to answer in the wrong language.

## Fix

- **Detect the language by scanning for an actual language token** instead of assuming a position. The candidate ids are resolved from the tokenizer's own special-token vocabulary, so this is correct whether or not `generate()` echoes the decoder prefix, and it cannot be fooled by a text token. Only Whisper language tokens are 2–3 letters inside `<|…|>`, so `<|transcribe|>`, `<|nospeech|>`, `<|notimestamps|>` and timestamp tokens can't match.
- **An explicitly requested language is authoritative.** The first pass already ran with it, so never re-generate — that is what threw away a correct transcription.
- **Removed the second positional decode** after the fallback branch, which overwrote the language the fallback had just chosen with another mis-parse.
- **Stop re-generating with `language=None`** when there is no known-good language to retry with. That produced an identical result at twice the cost.
- **Keep a real-but-unsupported detected language** (e.g. `ru`) rather than discarding a correct transcription, and don't let it become the sticky `last_language` fallback.
- Build the retry kwargs without mutating the handler's shared `gen_kwargs`.

## Tests

`tests/test_whisper_language_detection.py` drives the handler with a fake processor/model, covering both the legacy prefix-echoing behaviour and the current prefix-stripping behaviour, so no model download is needed.

The fake tokenizer implements `decode()` with realistic word pieces, so the suite reproduces the actual reported symptom against the previous code rather than erroring out. On `main`, the core case fails with:

```
assert result.language_code == "de"
E   AssertionError: assert 'e' == 'de'
```

All 14 tests fail before the change and pass after. Full suite: 611 passed. `ruff check`, `ruff format --check` and `mypy src/` are clean.

## Note on overlap

This does not touch the mutable-default `gen_kwargs` handling in `setup()`, which is what #373 and #374 address — those should merge cleanly alongside this.

### DIFF
--- a/src/speech_to_speech/STT/whisper_stt_handler.py
+++ b/src/speech_to_speech/STT/whisper_stt_handler.py
@@ -143,14 +143,60 @@ class WhisperSTTHandler(BaseSTTHandler):
         self._language_token_id_map = mapping
         return mapping
 
-    def _detected_language(self, pred_ids: Any) -> Optional[str]:
-        """Read the language Whisper decoded, or ``None`` if it did not report one.
+    def _code_from_language_tokens(self, token_ids: Any) -> Optional[str]:
+        """Resolve the first Whisper language special token in ``token_ids`` to its code."""
+        language_tokens = self._language_token_ids()
+        if not language_tokens:
+            return None
 
-        The language token is only in the returned sequence when ``generate()`` echoes the
-        forced decoder prefix. Recent ``transformers`` strips that prefix, so the leading ids
-        are ordinary text. Scanning for a real language token instead of slicing a fixed
-        position keeps this correct on both behaviours; guessing from position turns a text
-        token into a bogus "language" and silently discards a good transcription.
+        try:
+            flat = token_ids.flatten().tolist() if hasattr(token_ids, "flatten") else list(token_ids)
+        except (IndexError, TypeError):
+            return None
+
+        for token_id in flat:
+            try:
+                code = language_tokens.get(int(token_id))
+            except (TypeError, ValueError):
+                continue
+            if code is not None:
+                return code
+        return None
+
+    def _detect_language(self, input_features: Any) -> tuple[Any, Optional[str]]:
+        """Detect the spoken language up front, returning ``(encoder_outputs, code)``.
+
+        ``generate()`` excludes the decoder input ids from its output, and the language token
+        is one of them, so the detected language is never observable in the generated
+        sequence on the ``transformers`` versions supported here. Whisper exposes
+        ``detect_language()`` for exactly this, so ask it directly.
+
+        The encoder output is computed once here and handed back so ``generate()`` can reuse
+        it instead of re-encoding. Forcing the detected language also stops ``generate()``
+        from running its own internal detection, so this is cheaper than the plain auto path
+        rather than an extra cost.
+        """
+        detect_language = getattr(self.model, "detect_language", None)
+        if detect_language is None:
+            return None, None
+
+        try:
+            encoder_outputs = self.model.get_encoder()(input_features)
+            token_ids = detect_language(encoder_outputs=encoder_outputs)
+        except Exception as e:  # pragma: no cover - depends on the installed transformers
+            logger.warning("Whisper language detection failed, falling back: %s", e)
+            return None, None
+
+        return encoder_outputs, self._code_from_language_tokens(token_ids)
+
+    def _language_from_prefix(self, pred_ids: Any) -> Optional[str]:
+        """Read a language token off the generated sequence, if this version emits one.
+
+        Older ``transformers`` echoed the forced decoder prefix
+        (``<|startoftranscript|><|de|><|transcribe|>``) in ``generate()`` output. Current
+        versions strip it, so this is only a fallback for when ``detect_language()`` is
+        unavailable. Scanning for a real language token rather than slicing a fixed position
+        is what keeps a *text* token from being mistaken for a language.
         """
         language_tokens = self._language_token_ids()
         if not language_tokens:
@@ -162,14 +208,7 @@ class WhisperSTTHandler(BaseSTTHandler):
         except (IndexError, TypeError):
             return None
 
-        for token_id in token_ids[:_LANGUAGE_TOKEN_SCAN_DEPTH]:
-            try:
-                code = language_tokens.get(int(token_id))
-            except (TypeError, ValueError):
-                continue
-            if code is not None:
-                return code
-        return None
+        return self._code_from_language_tokens(token_ids[:_LANGUAGE_TOKEN_SCAN_DEPTH])
 
     def _forced_language(self) -> Optional[str]:
         """The language explicitly requested for generation, if any."""
@@ -180,33 +219,35 @@ class WhisperSTTHandler(BaseSTTHandler):
         logger.debug("infering whisper...")
 
         input_features = self.prepare_model_inputs(vad_audio.audio)
-        pred_ids = self.model.generate(input_features, **self.gen_kwargs)
-
         forced_language = self._forced_language()
-        detected_language = self._detected_language(pred_ids) or forced_language
 
-        if detected_language in SUPPORTED_LANGUAGES:
-            assert detected_language is not None
-            language_code = detected_language
+        gen_kwargs: dict[str, Any] = dict(self.gen_kwargs)
+        language_code = forced_language
+        if forced_language is None:
+            # Auto-detect mode: ask Whisper which language this is, then force it so the
+            # transcription and the reported code cannot disagree.
+            encoder_outputs, detected = self._detect_language(input_features)
+            if encoder_outputs is not None:
+                gen_kwargs["encoder_outputs"] = encoder_outputs
+            if detected is not None:
+                gen_kwargs["language"] = detected
+                language_code = detected
+
+        pred_ids = self.model.generate(input_features, **gen_kwargs)
+
+        if language_code is None:
+            # detect_language() was unavailable. Fall back to a prefix token if this version
+            # emits one, then to the last known language, and only then to the default.
+            language_code = self._language_from_prefix(pred_ids) or self.last_language or DEFAULT_LANGUAGE
+
+        # Report whatever language was actually transcribed, even if it is outside
+        # SUPPORTED_LANGUAGES -- discarding a correct transcription because its language is
+        # not on a downstream allowlist is the bug this handler had. Only remember supported
+        # languages, so an unsupported one never becomes the sticky fallback.
+        if language_code in SUPPORTED_LANGUAGES:
             self.last_language = language_code
-        elif forced_language is not None:
-            # The first pass already ran with the requested language, so it is authoritative.
-            # Re-generating here would throw away a correct transcription.
-            language_code = forced_language
         else:
-            if detected_language is not None:
-                logger.warning("Whisper detected unsupported language: %s", detected_language)
-            last_language = self.last_language
-            if last_language in SUPPORTED_LANGUAGES:
-                assert last_language is not None
-                # Auto-detection is unusable and we have a known-good language: retry with it.
-                logger.debug("Reprocessing with the last known language: %s", last_language)
-                pred_ids = self.model.generate(input_features, **{**self.gen_kwargs, "language": last_language})
-                language_code = last_language
-            else:
-                # Nothing better to fall back to. Keep this pass rather than re-generating
-                # with language=None, which would produce an identical result at double cost.
-                language_code = detected_language or DEFAULT_LANGUAGE
+            logger.warning("Whisper detected unsupported language: %s", language_code)
 
         pred_text = self.processor.batch_decode(pred_ids, skip_special_tokens=True, decode_with_timestamps=False)[0]
 
--- a/src/speech_to_speech/STT/whisper_stt_handler.py
+++ b/src/speech_to_speech/STT/whisper_stt_handler.py
@@ -175,14 +175,21 @@ class WhisperSTTHandler(BaseSTTHandler):
         it instead of re-encoding. Forcing the detected language also stops ``generate()``
         from running its own internal detection, so this is cheaper than the plain auto path
         rather than an extra cost.
+
+        Both calls run under ``torch.no_grad()``. Calling the encoder directly does not
+        inherit the no-grad context that ``generate()`` applies internally, so without this
+        the encoder output would keep its autograd graph alive for the whole of decoding --
+        wasted memory for a forward-only pipeline, and enough to risk OOM with a large
+        Whisper checkpoint.
         """
         detect_language = getattr(self.model, "detect_language", None)
         if detect_language is None:
             return None, None
 
         try:
-            encoder_outputs = self.model.get_encoder()(input_features)
-            token_ids = detect_language(encoder_outputs=encoder_outputs)
+            with torch.no_grad():
+                encoder_outputs = self.model.get_encoder()(input_features)
+                token_ids = detect_language(encoder_outputs=encoder_outputs)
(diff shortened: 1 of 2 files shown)


andimarafioti (changes requested):
LGTM but codex has some requests that seem valid, specially the first one, can you address it?

## Verdict: **Request changes**

The implementation fixes explicitly forced languages such as `language="de"`, but the default/auto-detection path remains broken. CI is green, though the tests currently encode the incorrect auto-mode behavior.

### P1: Auto mode labels every prefix-stripped transcription as English

The patch tries to recover the language by scanning `pred_ids`, then falls back to `DEFAULT_LANGUAGE = "en"` when no token is found.

But ordinary Whisper `generate()` output explicitly **excludes decoder-input IDs**. The language token is one of those decoder-input IDs, so it cannot appear in `pred_ids` under the normal API used here.  Transformers detects the language internally and inserts it into `init_tokens`, before stripping those tokens from the returned output.

Consequently, for German audio in auto mode:

1. Whisper correctly detects German and generates German text.
2. `_detected_language(pred_ids)` returns `None`.
3. The handler reports `en-auto`.
4. With language prompting enabled, the LLM is explicitly told to reply in English.

The new test actually codifies this bug by feeding German text and asserting `result.language_code == "en-auto"`.

**Suggested fix:** in auto mode, call Whisper’s public `model.detect_language(input_features)` method, map its returned ID to the language code, and pass that code into `generate(language=...)`. That does not add redundant detection overall: forcing the detected language prevents `generate()` from running its own internal detection again.

A suitable inline comment would be:

> Normal `Whisper.generate()` tensors exclude the decoder input IDs, including the detected language token, so `_detected_language(pred_ids)` will always return `None` on the Transformers versions supported here. This makes auto mode label non-English transcriptions as `en-auto`, which can then force the LLM and TTS into English. Could we use `model.detect_language(input_features)` before generation and pass the detected code back through `language=...`?

### P2: A valid unsupported language is still discarded when `last_language` exists

When a real language token such as Russian is detected and `last_language == "de"`, the patch deliberately retranscribes the Russian audio as German and discards the original result.  The test explicitly expects this destructive fallback.

That contradicts the PR’s stated goal of keeping a “real-but-unsupported detected language.” It also repeats the underlying failure pattern: a valid transcription is thrown away because its language is outside a downstream allowlist.

The fallback should only apply when the language is genuinely unknown. When Whisper confidently reports `ru`, preserve the transcription and return `ru-auto`; simply avoid making it the sticky fallback if downstream support is limited.

A suitable inline comment would be:

> This still discards a valid transcription whenever Whisper detects a real language outside `SUPPORTED_LANGUAGES` and a previous supported language exists. For example, Russian after German is retranscribed as German. Could we retain any actual detected language and reserve `last_language` fallback for cases where no language could be determined?

Other than these two behavioral issues, the changes are well structured, avoid mutating shared generation kwargs during retries, and have good regression-test scaffolding. I have not posted anything to the PR.

Hotragn:
Thanks — both points were valid, and P1 was worse than "the auto path remains broken": I verified your reasoning against transformers 4.57.3 with `openai/whisper-tiny` and the language token is absent from `generate()` output in **both** modes, so `_detected_language(pred_ids)` returned `None` unconditionally and auto mode reported `en-auto` for every language. Pushed in 869f7e4.

## P1 — auto mode now detects explicitly

Mechanism, measured on a real model and real audio rather than argued:

```
prefix scan on real generate() output : None          # so the old path always defaulted
detect_language()                     : en            # correct
auto-mode generate() first 4 tokens   : ['ĠI', "'m", 'Ġconfused', 'Ġwhy']
```

Auto mode now calls `model.detect_language()` and forces the result into `generate(language=...)`, so the transcription and the reported code can't disagree.

Same 6 s clip through the actual handler, all three revisions:

| revision | `--language auto` | `--language de` | `--language en` |
| --- | --- | --- | --- |
| `main` | `'-auto'` | `'a'` | `''` |
| first revision of this PR | `'en-auto'` (always, any language) | `'de'` | `'en'` |
| now | `'en-auto'` (actually detected) | `'de'` | `'en'` |

Note `main` mislabels even English — the codes it produces are slice artifacts.

**On the cost**, since this is a latency-sensitive pipeline: I run the encoder once and pass `encoder_outputs` to both `detect_language()` and `generate()`. I checked that this is bit-identical to letting `generate()` re-encode, at 2 s / 7 s / 13 s / 40 s inputs — `prepare_model_inputs` always pads or truncates to 3000 frames, so there's no longform path for the precomputed encoder output to break. Combined with forcing the detected language (which skips `generate()`'s internal detection), auto mode ends up **faster** than before:

```
new auto-mode process()        2099 ms
previous single generate()     3081 ms   (that call alone, same input)
```

And it costs nothing in the default configuration, where `--language` defaults to `en` and the forced path runs with no detection at all.

I kept the token scan, demoted to `_language_from_prefix()` and documented as a fallback for versions without `detect_language()`. Happy to drop it entirely if you'd rather not carry the dead-on-current-versions branch.

## P2 — no transcription is discarded any more

You were right that this contradicted the PR's own stated goal. There is now **no re-generation path at all**. The reported language is whatever was actually transcribed, and `SUPPORTED_LANGUAGES` only governs whether it becomes the sticky `last_language`. Russian after German now returns `ru-auto` with the Russian text, instead of being re-transcribed as German.

The two tests that encoded the old behaviour are gone. New cases cover the forced-detection path, the shared encoder pass, an unsupported detected language being preserved, and `detect_language` being absent or raising.

Full suite 616 passed; `ruff check`, `ruff format --check`, `mypy src/` clean.

src/speech_to_speech/STT/whisper_stt_handler.py
```diff
@@ -112,24 +125,131 @@ def warmup(self) -> None:
...
+        it instead of re-encoding. Forcing the detected language also stops ``generate()``
+        from running its own internal detection, so this is cheaper than the plain auto path
+        rather than an extra cost.
+        """
+        detect_language = getattr(self.model, "detect_language", None)
+        if detect_language is None:
+            return None, None
+
+        try:
+            encoder_outputs = self.model.get_encoder()(input_features)
```
andimarafioti: This direct encoder call runs with gradients enabled, unlike `generate()`. The returned encoder output therefore retains its full autograd graph throughout decoding, increasing memory use for the default large Whisper model and potentially causing GPU OOMs in auto-language mode. Could we run the encoder and detection inside `torch.no_grad()` or `torch.inference_mode()`, and add a test asserting that the encoder executes with gradients disabled?

andimarafioti (changes requested):
The earlier behavioral issues are addressed, and the focused regression tests pass. One GPU-memory issue remains in the new manual encoder path.

src/speech_to_speech/STT/whisper_stt_handler.py
Hotragn: Good catch — confirmed and fixed in 6a38b01.

You're right that the direct encoder call doesn't inherit the no-grad context `generate()` applies internally. Measured on transformers 4.57.3 / `openai/whisper-tiny`:

```
before:  requires_grad=True   grad_fn=NativeLayerNormBackward0
after:   requires_grad=False  grad_fn=None
```

Since that encoder output is then handed to `generate()` and held for the whole of decoding, the entire graph stayed alive with it — exactly the OOM risk you describe, and worse on a large checkpoint.

Both the encoder forward and `detect_language()` now run inside `torch.no_grad()`. Transcription output is unchanged (`en-auto`, identical text) and the ambient grad mode is restored afterwards.

I used `no_grad()` rather than `inference_mode()` deliberately: the encoder output crosses into `generate()`, and inference tensors carry extra restrictions there for no benefit in this path.

Three tests added, asserting the grad state observed *inside* the encoder and `detect_language()` calls rather than just around them:

- `test_detection_runs_with_gradients_disabled` — grad is off in both calls
- `test_detection_restores_the_ambient_grad_mode` — `no_grad` is scoped, doesn't leak
- `test_detection_runs_with_gradients_disabled_even_under_enabled_grad` — still holds when the caller wraps the handler in `torch.enable_grad()`

The first and third fail against the previous commit. Full suite 637 passed; ruff, format and mypy clean.

Hotragn:
Status note while this waits on a re-review — no code change.

I verified this branch against current `main` (`0739839`, i.e. 0.2.12 with Smart Turn VAD and the dependency-extras cleanup) by rebasing locally: **808 passed, 1 skipped**, ruff/format/mypy clean. `base_stt_handler.py` changed in the Smart Turn work and these handlers subclass it, so that seemed worth confirming rather than assuming.

I've deliberately **not** pushed that rebase. Your inline comment on `whisper_stt_handler.py:184` is the active thread here, and force-pushing would risk marking it outdated mid-review for no real benefit — GitHub still reports this `mergeable=true` against `main`, and `whisper_stt_handler.py` hasn't been touched upstream since this branch's base. Happy to rebase the moment you'd prefer it.

Recap of where the two review rounds landed, so nothing is buried:

- **P1** (auto mode labelled everything `en`) — auto mode now calls `model.detect_language()` and forces the result into `generate(language=...)`. Verified on transformers 4.57.3 that the language token is genuinely absent from `generate()` output, so the old scan returned `None` unconditionally.
- **P2** (valid transcription discarded) — no re-generation path remains; `SUPPORTED_LANGUAGES` governs only the sticky fallback.
- **GPU memory** — encoder + detection run under `torch.no_grad()`; confirmed `requires_grad` goes `True` → `False` on a real model, with three tests asserting grad state observed *inside* the calls.

Still open from my side: I offered to drop `_language_from_prefix()` entirely, since it's dead on every `transformers` version this supports and only exists as a fallback for versions without `detect_language()`. Your call — happy either way.

andimarafioti (approved):
Thank you! AWesome work! Specially thank you for working through the reviews!

PR #379, opened 2026-07-29
Acquire the global MLX lock in MLXAudioWhisperSTTHandler

### LINKED ISSUE #377: MLXAudioWhisperSTTHandler bypasses the global MLX lock → Metal assertion crash (SIGABRT)
## Summary

`STT/mlx_audio_whisper_handler.py` calls `self.model.generate()` without acquiring the
global MLX lock that `utils/mlx_lock.py` documents as mandatory. When Whisper inference
overlaps with Qwen3-TTS generation, both touch the same Metal command buffer and the
process aborts:

```
-[_MTLCommandBuffer addCompletedHandler:]:1011: failed assertion `Completed handler provided after commit call'
```

The process dies with SIGABRT and **no Python traceback**, so it looks like a silent
disappearance rather than a crash.

## Why this looks like a clear oversight

`utils/mlx_lock.py` states the requirement explicitly in its module docstring:

> MLX models (STT, LLM, TTS) cannot be used concurrently from multiple threads on Apple
> Silicon due to Metal command buffer limitations. This module provides a global lock
> that all MLX handlers should acquire before using their models.

Every other MLX inference path honors it — `STT/parakeet_tdt_handler.py`,
`TTS/qwen3_tts_handler.py`, `TTS/kokoro_handler.py`, `LLM/language_model.py` all import
and use `MLXLockContext`. `STT/mlx_audio_whisper_handler.py` is the only one that does
not import it at all.

On v0.2.11 the two unguarded call sites are:

- `mlx_audio_whisper_handler.py:93` — in `warmup()`
- `mlx_audio_whisper_handler.py:113` — in `process()`

## Reproduction

macOS 26.5.1, Apple M5 (24 GB), Python 3.12.13,
`speech-to-speech` 0.2.11, `mlx` 0.31.1, `mlx-audio` 0.4.2.

```bash
speech-to-speech \
  --mode local \
  --device mps \
  --stt mlx-audio-whisper \
  --mlx_audio_whisper_model_name mlx-community/whisper-large-v3-turbo \
  --language auto \
  --llm_backend chat-completions \
  --model_name <any OpenAI-compatible model> \
  --responses_api_base_url <endpoint> \
  --tts qwen3 \
  --qwen3_tts_model_name Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign \
  --qwen3_tts_mlx_quantization 6bit
```

Hold a multi-turn conversation. The crash needs STT and TTS to overlap, so it is timing
dependent — in the run captured below it hit on turn 9. Turn churn increases the odds:
barge-in / speculative turn reopens (`VAD: reopened speculative turn turn_N revision 1`)
are a reliable way to get an STT request issued while TTS is still generating.

Last log lines before the abort in one run:

```
... - STT.transcription_notifier - INFO - Transcription completed (language=en-auto): ...
... - baseHandler - INFO - ChatCompletionsApiModelHandler: 0.377 s
ASSISTANT: I'm not sure what you're asking. Could you please clarify?
  0%|          | 0/356 [00:00<?, ?frames/s]        <-- Whisper still running
-[_MTLCommandBuffer addCompletedHandler:]:1011: failed assertion `Completed handler provided after commit call'
```

Note the Whisper progress bar is still active while the TTS handler starts generating.

## Fix

Wrap both `self.model.generate()` calls in `MLXLockContext`, matching what
`parakeet_tdt_handler` already does:

```python
from speech_to_speech.utils.mlx_lock import MLXLockContext

# in warmup()
with MLXLockContext(handler_name=self.__class__.__name__):
    _ = self.model.generate(dummy_audio, verbose=False)

# in process()
with MLXLockContext(handler_name=self.__class__.__name__):
    result = self.model.generate(audio_input, verbose=False, **gen_kwargs)
```

## Result after patching

12-turn stress run on the same setup: 12/12 turns completed, process survived, no
Metal assertions. End-to-end latency (`Last speech detected to first speech out`) was
unchanged — min 1.16 s, median 1.50 s, max 3.20 s (first turn, cold). The lock's own
contention logging never crossed its 0.25 s reporting threshold, so serializing STT
against TTS cost nothing measurable in this configuration.

Happy to open a PR if useful.

Hotragn:
Fixes #377

## Problem

`utils/mlx_lock.py` documents the global lock as mandatory for every MLX handler:

> MLX models (STT, LLM, TTS) cannot be used concurrently from multiple threads on Apple Silicon due to Metal command buffer limitations. This module provides a global lock that all MLX handlers should acquire before using their models.

`STT/mlx_audio_whisper_handler.py` was the only MLX inference path that did not import or acquire it. `parakeet_tdt_handler.py`, `TTS/qwen3_tts_handler.py`, `TTS/kokoro_handler.py` and `LLM/language_model.py` all already honor it.

When Whisper STT inference overlaps Qwen3-TTS generation, both touch the same Metal command buffer and the process aborts:

```
-[_MTLCommandBuffer addCompletedHandler:]:1011: failed assertion `Completed handler provided after commit call'
```

It dies with SIGABRT and **no Python traceback**, so it looks like the server silently disappeared rather than crashed. Barge-in and speculative turn reopens (`VAD: reopened speculative turn …`) make this easy to hit, since they issue an STT request while TTS is still generating.

## Fix

Wrap the `generate()` call in both `warmup()` and `process()` in `MLXLockContext`, so this handler serializes against every other MLX consumer. `MLXLockContext.__exit__` releases on the exception path too, so a failed transcription can no longer strand the lock and wedge the whole pipeline.

While in the same language-resolution block, two related defects in the code the lock now wraps:

- `hasattr(result, "language")` was true even when the value was `None`, so the handler logged `Detected unsupported language: None` and skipped the language the user had explicitly requested. Absent and `None` are now both treated as "no language reported", and an explicitly requested language wins outright since `generate()` already ran with it.
- `setup()` stored `--language auto` verbatim in `last_language`, where it then failed every `SUPPORTED_LANGUAGES` check it was consulted for. It is now normalized to `None`, matching what `WhisperSTTHandler` already does.

## Tests

`tests/test_mlx_audio_whisper_handler.py` fakes the model, so it needs no MLX/mlx-audio install and runs on Linux CI.

The lock tests assert the lock is actually **held** while `generate()` runs — probed with a non-blocking acquire from a second thread — rather than just checking the import. One test drives two handlers concurrently and asserts they are never inside `generate()` at the same time, which is the condition that produced the Metal assertion. Others cover release on the exception path for both `process()` and `warmup()`.

7 of the 18 tests fail against `main`. Full suite: 615 passed. `ruff check`, `ruff format --check` and `mypy src/` are clean.

## Verification on device

Reported in #377 from a 12-turn stress run on macOS 26.5.1 / Apple M5, `mlx` 0.31.1, `mlx-audio` 0.4.2, with `--stt mlx-audio-whisper --tts qwen3`: 12/12 turns completed, no Metal assertions. End-to-end latency (last speech detected → first speech out) unchanged — min 1.16 s, median 1.50 s, max 3.20 s. The lock's own contention logging never crossed its 0.25 s reporting threshold, so serializing STT against TTS cost nothing measurable in that configuration.

### DIFF
--- a/src/speech_to_speech/STT/mlx_audio_whisper_handler.py
+++ b/src/speech_to_speech/STT/mlx_audio_whisper_handler.py
@@ -9,10 +9,13 @@ from rich.console import Console
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
 from speech_to_speech.pipeline.messages import Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
+from speech_to_speech.utils.mlx_lock import MLXLockContext
 
 logger = logging.getLogger(__name__)
 
 console = Console()
+
+DEFAULT_LANGUAGE = "en"
 
 SUPPORTED_LANGUAGES = [
     "en",
@@ -47,7 +50,9 @@ class MLXAudioWhisperSTTHandler(BaseSTTHandler):
 
         self.model_name = model_name
         self.start_language = language
-        self.last_language = language
+        # "auto" is a request to detect, not a language code, so it must never leak into
+        # last_language -- it would fail every SUPPORTED_LANGUAGES check downstream.
+        self.last_language = language if language != "auto" else None
         self.gen_kwargs = gen_kwargs
 
         # Load the model directly
@@ -90,10 +95,37 @@ class MLXAudioWhisperSTTHandler(BaseSTTHandler):
 
         try:
             # Pre-warm the model by running a transcription
-            _ = self.model.generate(dummy_audio, verbose=False)
+            with MLXLockContext(handler_name=self.__class__.__name__):
+                _ = self.model.generate(dummy_audio, verbose=False)
             logger.info("Model warmed up and ready")
         except Exception as e:
             logger.warning(f"Warmup failed: {e}")
+
+    def _forced_language(self) -> Optional[str]:
+        """The language explicitly requested by the user, if any."""
+        if self.start_language and self.start_language != "auto":
+            return self.start_language
+        return None
+
+    def _resolve_language(self, result: Any) -> str:
+        """Pick the language code to report, updating the sticky fallback on success."""
+        forced = self._forced_language()
+        if forced is not None:
+            # generate() ran with this language, so it is authoritative.
+            self.last_language = forced
+            return forced
+
+        detected = getattr(result, "language", None)
+        if isinstance(detected, str) and detected:
+            if detected in SUPPORTED_LANGUAGES:
+                self.last_language = detected
+                return detected
+            logger.warning("Detected unsupported language: %s", detected)
+
+        last_language = self.last_language
+        if last_language is not None and last_language in SUPPORTED_LANGUAGES:
+            return last_language
+        return DEFAULT_LANGUAGE
 
     def process(self, vad_audio: STTIn) -> Iterator[STTOut]:
         logger.debug("inferring mlx-audio whisper...")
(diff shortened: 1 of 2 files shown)


andimarafioti (approved):
Great work! Thank you !

PR #382, opened 2026-07-29
Add Qwen3-ASR STT backend

### LINKED ISSUE #337: Add Qwen3-ASR support as an STT backend
## Motivation

Qwen3-ASR now has a small 0.6B checkpoint (`Qwen/Qwen3-ASR-0.6B-hf`) that looks like a good fit for the local STT slot in this project, especially on consumer hardware.

It would be useful to support it via a dedicated STT backend, for example:

```bash
speech-to-speech --stt qwen3-asr --qwen3_asr_model_name Qwen/Qwen3-ASR-0.6B-hf
```

## Notes from local validation

- `transformers==5.6.2` does not recognize `model_type: qwen3_asr`.
- Testing intermediate releases showed `qwen3_asr` is recognized starting at `transformers==5.13.0`.
- With `transformers==5.13.0`, `AutoConfig` resolves to `Qwen3ASRConfig` and `AutoProcessor` resolves to `Qwen3ASRProcessor`.
- A local smoke test with the 0.6B model downloaded from ModelScope was able to run inference and produce a transcript.

## Proposed scope

- Add `qwen3-asr` to the `--stt` choices.
- Add a `Qwen3ASRSTTHandler` using `AutoModelForMultimodalLM` and `AutoProcessor`.
- Add CLI arguments for model name, device, dtype, language, and generation options.
- Keep the initial implementation final-transcript only. Progressive/partial transcription can be considered separately if the model/API supports it well enough.
- Document that Qwen3-ASR requires a compatible Transformers version, currently `>=5.13.0`, instead of silently failing on older installs.

## Dependency concern

Raising the global Transformers requirement may conflict with existing backends, so this might be better handled as an optional/extra path or with a clear runtime error when the installed Transformers version is too old.

poorvajasathasivam:
Closes #337

- Add `qwen3-asr` as an STT backend option.
- Add `Qwen3ASRSTTHandler` using Transformers `AutoModelForMultimodalLM` and `AutoProcessor`.
- Add Qwen3-ASR CLI arguments for model, device, dtype, language, prompt, compile mode, and generation options.
- Document the optional `qwen3-asr` extra and Transformers `>=5.13.0` requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ pip install "speech-to-speech[facebook-mms]"    # MMS TTS
 pip install "speech-to-speech[faster-whisper]"  # Faster Whisper STT
 pip install "speech-to-speech[whisper-mlx]"     # Lightning Whisper MLX STT on macOS
 pip install "speech-to-speech[paraformer]"      # Paraformer STT through FunASR
+pip install "speech-to-speech[qwen3-asr]"       # Qwen3-ASR STT through Transformers >=5.13.0
 pip install "speech-to-speech[mlx-lm]"          # mlx-vlm support for vision models on macOS
 ```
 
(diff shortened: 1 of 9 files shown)


poorvajasathasivam (commented):
Closes #337

dcrey7:
Hi @poorvajasathasivam, thanks for the first version. Main moved the backend wiring into `backend_registry.py` after you opened this, so the branch no longer merges. I rebased the same approach onto current main in #544, with credit to you, plus:

+ Registry entry - no `module_arguments.py` or `s2s_pipeline.py` edits needed now
+ Warmup - like the Whisper and Paraformer handlers
+ Partials - progressive mode, reusing the last final language for short windows
+ Language table - from the Transformers processor, not a copy
+ Tests and docs

If you would rather pull the commits into this PR, say so and I will close mine.

poorvajasathasivam:
Thanks for handling this, @dcrey7! Appreciate the shoutout and glad #544 got merged. I'll close this one out.

PR #387, opened 2026-07-30
fix: acquire MLX lock in LightningWhisperSTTHandler to prevent SIGSEGV

### LINKED ISSUE #385: LightningWhisperSTTHandler does not acquire the global MLX lock — SIGSEGV in mlx::core::metal::CommandEncoder (the second path missed by #377/#379)
## Summary

#379 fixed `STT/mlx_audio_whisper_handler.py`, but its description states that handler was
*"the only MLX inference path that did not acquire the lock"*. That is not the case:
**`STT/lightning_whisper_mlx_handler.py` — the handler behind `--stt whisper-mlx` — still has
zero references to `mlx_lock` on `main`** (`cc37fe8`). It runs `LightningWhisperMLX.transcribe()`
(an MLX model) with no serialization against `Qwen3TTSHandler`, `KokoroHandler` or
`LLM/language_model`, all of which do hold the lock.

I hit this as a **SIGSEGV** (not the SIGABRT from #377): the process died with an invalid read
inside `mlx::core::metal::CommandEncoder`'s constructor while a second Python thread was
concurrently in `mlx::core::array::eval()`. Same root cause, different crash signature.

```
$ git grep -l mlx_lock cc37fe8 -- src/
src/speech_to_speech/LLM/language_model.py
src/speech_to_speech/STT/mlx_audio_whisper_handler.py     <- fixed by #379
src/speech_to_speech/STT/parakeet_tdt_handler.py
src/speech_to_speech/TTS/kokoro_handler.py
src/speech_to_speech/TTS/qwen3_tts_handler.py
src/speech_to_speech/s2s_pipeline.py
src/speech_to_speech/utils/mlx_lock.py
#  STT/lightning_whisper_mlx_handler.py is absent
```

## Environment

| | |
|---|---|
| `speech-to-speech` | 0.2.11 (installed wheel; behavior on `main` is unchanged for this file) |
| Hardware / OS | Apple M4 Max, macOS 26.5.2 (25F84) |
| `mlx` / `mlx-metal` / `mlx-lm` | 0.31.1 (as pinned by `pyproject.toml`) |
| `lightning-whisper-mlx` | 0.0.10 |
| Pipeline | `--mode realtime --stt whisper-mlx --tts qwen3-tts` (Qwen3-TTS 1.7B CustomVoice 6bit, MLX) |
| LLM | **out of process** (OpenAI-compatible endpoint), so the only two in-process MLX consumers were TTS and STT |

## The invariant, and who violates it

`utils/mlx_lock.py` states it as mandatory:

> MLX models (STT, LLM, TTS) cannot be used concurrently from multiple threads on Apple Silicon
> due to Metal command buffer limitations. This module provides a global lock that **all MLX
> handlers should acquire** before using their models.

and `s2s_pipeline.py:1017` repeats it:

> `# On Apple Silicon, all MLX inference serializes through a global lock (utils/mlx_lock.py).`

`STT/lightning_whisper_mlx_handler.py` has four unguarded MLX inference calls
(line numbers from the installed 0.2.11; identical on `main`):

- `warmup()` — `self.model.transcribe(dummy_input)` at **line 66**
- `process()` — `self.model.transcribe(...)` at **lines 73, 75 and 80** (the 80 one is the
  unsupported-language re-transcribe)

## Crash evidence

`~/Library/Logs/DiagnosticReports/Python-2026-07-29-174903.ips`,
`EXC_BAD_ACCESS (SIGSEGV) / KERN_INVALID_ADDRESS at 0x00579525aebc76c8`.

**Crashing thread** (a Python thread, `thread_run` → … → `array::eval()`):

```
objc_msgSend                                                          <- invalid read
-[NSObject isMemberOfClass:]
-[AGXG16XFamilyCommandBuffer tryCoalescingPreviousComputeCommandEncoderWithConfig:nextEncoderClass:]
-[AGXG16XFamilyCommandBuffer computeCommandEncoderWithConfig:]
-[AGXG16XFamilyCommandBuffer computeCommandEncoderWithDispatchType:]
mlx::core::metal::CommandEncoder::CommandEncoder(mlx::core::metal::DeviceStream&)
mlx::core::metal::Device::get_command_encoder(int)
mlx::core::steel_matmul_regular_axpby<true>(...)
mlx::core::steel_matmul_axpby<true>(...)
mlx::core::AddMM::eval_gpu(...)
mlx::core::gpu::eval(mlx::core::array&)
mlx::core::eval_impl(...)  ->  mlx::core::eval(...)  ->  mlx::core::array::eval()
core.cpython-312-darwin.so (?)  ->  PyObject_Vectorcall  ->  _PyEval_EvalFrameDefault
```

`tryCoalescingPreviousComputeCommandEncoderWithConfig:` is the tell: it dereferences the
**previous** encoder on that command buffer, and that object had been mutated/released by
another thread.

**A second Python thread was inside MLX at the same moment** — this is the part that makes it
concurrency rather than a one-off Metal glitch:

```
thread 5303023 (crashed)  ... mlx::core::array::eval() -> AddMM::eval_gpu -> CommandEncoder ctor -> SIGSEGV
thread 5303027            ... mlx::core::array::eval() -> array::wait() -> mlx::core::Event::wait()
                              (reached from Python via slot_nb_int / PyNumber_Long, i.e. int(mlx_array))
thread 5302877            mlx::core::scheduler::StreamThread::thread_fn() -> binary_op<Multiply>
```

Two independent `array::eval()` stacks on two Python threads, with the LLM out of process —
so those two threads can only be Qwen3-TTS and lightning-whisper STT.
(Weaker, supporting detail: the crashing op is `AddMM`, which `mlx.nn.Linear` only emits when
the layer **has a bias** — Whisper's linears do, Qwen3's are bias-free. Consistent with the
crashing thread being STT, but I would not lean on it.)

## Timeline from the run log

```
17:48:58,333  lm_output_processor  Forwarding to TTS: '这都什么跟什么呀我实在猜不出来麻烦你重新说一次吧'
17:48:58,333  mlx_lock             Qwen3TTS: MLX lock acquired after 0.000s (previous_owner=none, depth=1)
17:48:58,423  qwen3_tts_handler    Qwen3-TTS TTFA: 0.09s (custom_voice_mlx)
17:48:58,506 … 17:48:59,001        Qwen3TTSHandler: 0.07 s   (audio chunk yielded every ~70ms — still holding the lock)
17:48:59,013  vad_handler          Speech soft-ended (segment=3200ms, active=1792ms, turn=turn_4 rev=1)
17:48:59,013  lightning_whisper…   infering whisper...        <- MLX inference starts, no lock
              <last line in the log; crash report timestamped 17:49:03>
```

Lock accounting for the whole session: **24 acquires / 23 releases, all 24 by `Qwen3TTS`**.
The missing release is the 17:48:58,333 one — the process died while TTS still held the lock,
and STT walked straight into MLX anyway because it never asks for it.

Trigger, as in #377: **barge-in and speculative turn reopen**. The user spoke while TTS was
generating, so a VAD segment ended (`turn_4 rev=1`) and STT was dispatched mid-generation.

## Suggested fix

Same shape as `parakeet_tdt_handler.py` / #379 — wrap the inference in `MLXLockContext`,
in both `warmup()` and `process()`. In `process()` one context around the whole
language-resolution block covers all three `transcribe()` calls:

```python
from speech_to_speech.utils.mlx_lock import MLXLockContext
...
    with MLXLockContext(handler_name=self.__class__.__name__):
        if self.start_language != "auto":
            transcription_dict = self.model.transcribe(audio, language=self.start_language)
        else:
            ...
```

I'm happy to open a PR if that's useful — say the word. (I'd follow #379's testing approach:
fake the model so the lock tests assert the lock is actually *held* during `transcribe()`
and run on Linux CI.)

## Side effects worth stating up front

- **STT now queues behind TTS's lock**, and `qwen3_tts_handler.py:772` holds it for a whole
  utterance (`yield from` is inside the `with`, `timeout=10.0`). So the added latency is
  bounded by the remaining generation time of the current utterance.
- **Barge-in does not regress**: interruption is driven by VAD (`speech_started`), which never
  touches MLX and does not wait on the transcript. A cancellation closes the TTS generator,
  which releases the lock immediately.
- With sentence-level streaming working normally, utterances are short and the queueing cost is
  small. (Separately, on CJK text the sentence splitter's punctuation whitelist is ASCII-only, so
  a whole reply becomes one utterance and the hold gets long — I'll file that as its own issue;
  the two fixes complement each other but shouldn't be bundled.)
- `MLXLockContext.__exit__` releases on the exception path, so a failed transcription cannot
  strand the lock.

## Reproduction — honest status

**n = 1, not reproduced on demand.** It is a probabilistic race: I deliberately re-entered the
same window (long replies, repeated barge-in during playback) at least twice more without a
crash. What is solid is the mechanism: the invariant is documented, this handler provably
violates it, and the crash stack plus the two concurrent `array::eval()` threads match that
mechanism exactly. Frequency I can't quantify; the fix is one `with` statement either way.

## Relation to #377 / #379

Same root cause, second code path. Please treat this as the remainder of #377 rather than a new
class of bug — and note that #379's "the only MLX inference path that did not acquire it" needs
correcting, since `--stt whisper-mlx` is still unguarded. It isn't an exotic option either: it's
the MLX whisper path users land on when `parakeet-tdt`'s language coverage isn't enough — in my
case Chinese, where parakeet is not usable.

That also probably explains why this went unnoticed: `--local_mac_optimal_settings` forces
`stt = "parakeet-tdt"` (`s2s_pipeline.py:239`), i.e. the recommended Apple Silicon path happens
to pick the one STT handler that does hold the lock. The exposed population is users who need a
language parakeet can't do and therefore have to pass `--stt whisper-mlx` explicitly.

⚠️ Also worth flagging: switching `--stt mlx-audio-whisper` is **not** a workaround on the
released 0.2.11 — that's exactly the handler #379 fixed, and the fix isn't in a release yet
(`main`'s `pyproject.toml` still says `0.2.11`, no `v0.2.12` tag). On 0.2.11 both whisper
handlers are unguarded, so today the only lock-honoring STT is `parakeet-tdt`.

varunsahni18:
## Summary

Fixes #385

The `LightningWhisperSTTHandler` (the handler behind `--stt whisper-mlx`) was the only remaining MLX inference path that did not serialize through the global MLX lock. When STT and TTS ran concurrently on Apple Silicon, the shared Metal command buffer was corrupted, causing a SIGSEGV in `mlx::core::metal::CommandEncoder`.

PR #379 fixed the same bug in `mlx_audio_whisper_handler.py`, but `lightning_whisper_mlx_handler.py` was missed.

## Changes

- Import `MLXLockContext` from `speech_to_speech.utils.mlx_lock`
- Wrap `warmup()` transcribe call with `MLXLockContext`
- Wrap all three `process()` transcribe calls with `MLXLockContext` (forced-language path, auto-detect path, and re-transcribe fallback)

## Verification

```bash
$ git grep -l mlx_lock -- src/
src/speech_to_speech/LLM/language_model.py
src/speech_to_speech/STT/lightning_whisper_mlx_handler.py  # now listed
src/speech_to_speech/STT/mlx_audio_whisper_handler.py
src/speech_to_speech/STT/parakeet_tdt_handler.py
src/speech_to_speech/TTS/kokoro_handler.py
src/speech_to_speech/TTS/qwen3_tts_handler.py
src/speech_to_speech/s2s_pipeline.py
src/speech_to_speech/utils/mlx_lock.py
```

All four MLX inference call sites in the handler are now guarded, following the same pattern as the already-merged #379 fix.

### DIFF
--- a/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
+++ b/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
@@ -11,6 +11,7 @@ from rich.console import Console
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
 from speech_to_speech.pipeline.messages import Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
+from speech_to_speech.utils.mlx_lock import MLXLockContext
 
 logger = logging.getLogger(__name__)
 
@@ -63,21 +64,25 @@ class LightningWhisperSTTHandler(BaseSTTHandler):
         dummy_input = np.array([0] * 512)
 
         for _ in range(n_steps):
-            _ = self.model.transcribe(dummy_input)["text"].strip()
+            with MLXLockContext(handler_name=self.__class__.__name__):
+                _ = self.model.transcribe(dummy_input)["text"].strip()
 
     def process(self, vad_audio: STTIn) -> Iterator[STTOut]:
         logger.debug("infering whisper...")
 
         audio = vad_audio.audio
         if self.start_language != "auto":
-            transcription_dict = self.model.transcribe(audio, language=self.start_language)
+            with MLXLockContext(handler_name=self.__class__.__name__):
+                transcription_dict = self.model.transcribe(audio, language=self.start_language)
         else:
-            transcription_dict = self.model.transcribe(audio)
+            with MLXLockContext(handler_name=self.__class__.__name__):
+                transcription_dict = self.model.transcribe(audio)
             language_code = transcription_dict["language"]
             if language_code not in SUPPORTED_LANGUAGES:
                 logger.warning(f"Whisper detected unsupported language: {language_code}")
                 if self.last_language in SUPPORTED_LANGUAGES:  # reprocess with the last language
-                    transcription_dict = self.model.transcribe(audio, language=self.last_language)
+                    with MLXLockContext(handler_name=self.__class__.__name__):
+                        transcription_dict = self.model.transcribe(audio, language=self.last_language)
                 else:
                     transcription_dict = {"text": "", "language": "en"}
             else:


andimarafioti (approved):
Good work, thank you!

PR #388, opened 2026-07-30
Relax MLX dependency pins on macOS

### LINKED ISSUE #386: Consider relaxing the mlx==0.31.1 pin — MLX gained multithreaded support in 0.31.2
`pyproject.toml` pins exactly, not as a floor:

```
mlx==0.31.1;       platform_system == 'Darwin'
mlx-metal==0.31.1; platform_system == 'Darwin'
mlx-lm==0.31.1;    platform_system == 'Darwin'
```

Two reasons to revisit:

1. **The global MLX lock exists because MLX wasn't thread-safe. In 0.31.2 it is.**
   The 0.31.2 release notes say verbatim: *"MLX can be used by multiple threads for independent
   computations (#3405, #3348, #3281, #3423)"*, and ml-explore/mlx#3078 ("thread safety") was
   closed with *"MLX has thread safety support in 0.31.2 now"*. `0.31.1` is the version
   immediately before that work landed; latest is 0.32.0.
2. **The specific crash frame is a fixed MLX bug.** ml-explore/mlx#3216 —
   `tryCoalescingPreviousComputeCommandEncoderWithConfig` reached from `eval_gpu`, i.e. the same
   frame as the crash in #385 — was closed as completed on 2026-04-01, the day #3348
   (thread-local `CommandEncoder`) merged.

Note this does **not** replace the lock fix in #385: the reporter of #3216 tried
`MLX_NUM_THREADS=1`, `TOKENIZERS_PARALLELISM=false` and asyncio serialization, none of which
helped on the affected version — an application-level lock is necessary but, on 0.31.1, not
provably sufficient. Both changes are worth having.

Caveats I'd expect you to weigh (and I haven't tested any of this):

- `mlx-metal` must move in lockstep with `mlx`.
- ml-explore/mlx#3391 (`There is no Stream(gpu, N) in current thread`) is a *new* failure mode
  introduced by the thread-local work, hitting exactly this project's shape — model loaded on one
  thread, inference on per-handler threads. It's closed, but it's the first thing I'd watch for.
- Downstream constraints are already loose (`mlx-lm` wants `mlx>=0.30.4`, `mlx-audio`
  `mlx>=0.25.2`, `lightning-whisper-mlx` unconstrained), so the pin here is the only blocker.

Happy to run a bisect on a Mac if that would help.

abhishek-s12:
MLX 0.31.2 introduced thread safety for independent computations and fixed issues that originally motivated pinning to 0.31.1. Relaxing the pins allows users to benefit from upstream bug fixes while preserving compatibility bounds.

Fixes #386.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,10 +51,10 @@ dependencies = [
     "faster-qwen3-tts>=0.3.2; platform_system == 'Windows'",
     "lingua-language-detector>=2.0.2",
     "miniaudio==1.61; platform_system == 'Darwin'",
-    "mlx==0.31.1; platform_system == 'Darwin'",
+    "mlx>=0.31.2,<0.33; platform_system == 'Darwin'",
     "mlx-audio==0.4.2; platform_system == 'Darwin'",
-    "mlx-lm==0.31.1; platform_system == 'Darwin'",
-    "mlx-metal==0.31.1; platform_system == 'Darwin'",
+    "mlx-lm>=0.31.2,<0.33; platform_system == 'Darwin'",
+    "mlx-metal>=0.31.2,<0.33; platform_system == 'Darwin'",
     "misaki>=0.9.4; platform_system == 'Darwin'",
     "espeakng-loader>=0.2.4; platform_system == 'Darwin'",
     "spacy>=3.8.4; platform_system == 'Darwin'",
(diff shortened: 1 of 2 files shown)


andimarafioti:
I would prefer to upgrade the MLX packages. I just find that it periodically breaks things so you need to really test it. If you can do it, I'd be happy :)

andimarafioti:
no follow-up on my comment, leaving the issue open tho

PR #391, opened 2026-07-30
Stop SpeculativeTurnTracker from committing untracked turns

ztcools:
`_prune_tracked_turns` only walks `_latest_revision`, but all five commit entry points decided whether a revision was still current with

```python
latest = self._latest_revision.get(turn_id, revision)
if revision != latest:
    return False
```

The default is the caller's own revision, so the comparison always succeeds for a turn that is *not* tracked, and `_committed_revision[turn_id]` is written for a turn the tracker no longer knows about. That entry is then unreclaimable — pruning never visits it — and any later turn reusing the id reads as already committed.

Both ways of leaving `_latest_revision` are reachable, and the consequence is not just a leak:

- `VAD/vad_handler.py`'s `on_session_end` resets `_turn_counter` to 0 and calls `speculative_turns.reset()`. The VAD thread resets while the downstream STT/LLM/TTS threads are still draining the previous session, so their late commits land on an empty tracker and re-add the old `turn_N`. Because the counter restarted, the next session produces that same id, and `_should_reopen_current_turn` sees `is_committed(...)` return True and bails out with `return False`. A turn that was never answered can no longer reopen: the user finishes speaking and the assistant stays silent until they speak again.
- LRU eviction reaches the same state without any reset, once a long session pushes a turn past `_MAX_TRACKED_TURNS`.

## Changes

- Extract `_commit_locked` and route all five commit entry points through it. It looks the turn up without a default, so an untracked turn is no longer written back, which establishes `_committed_revision` as a subset of `_latest_revision`. This also removes five copies of the same block.
- An untracked turn still reports success rather than `False`. The return value is the caller's "is my output still valid" answer: `api/openai_realtime/handlers/response.py` treats `False` as "drop this text", and discarding the output of a turn the tracker merely forgot is worse than emitting it. Every entry point keeps its previous return semantics exactly, so no caller changes behavior.

## Tests

Adds four tests — seven cases once the `commit_if_*` variants are expanded — and all of them fail against the previous code:

| Test | Covers |
|------|--------|
| `test_commit_after_reset_does_not_resurrect_untracked_turn` | late commit after `reset()` |
| `test_commit_after_prune_does_not_resurrect_untracked_turn` | late commit after LRU eviction |
| `test_commit_if_latest_variants_keep_untracked_turn_out_of_committed_state` | all four `commit_if_*` entry points, including their unchanged return value |
| `test_reused_turn_id_after_reset_is_not_reported_as_committed` | a turn id reused after a session reset can still reopen |

They exercise the tracker directly, so they need no model and run on Linux CI.

Locally: `ruff check`, `ruff format --check`, `mypy src/` clean, and `pytest tests/` is 622 passed.

### DIFF
--- a/src/speech_to_speech/pipeline/speculative_turns.py
+++ b/src/speech_to_speech/pipeline/speculative_turns.py
@@ -100,26 +100,14 @@ class SpeculativeTurnTracker:
             return True
         with self._condition:
             self._wait_for_pending_reopen_locked(turn_id, revision, self._PENDING_REOPEN_WAIT_TIMEOUT_S)
-            latest = self._latest_revision.get(turn_id, revision)
-            if revision != latest:
-                return False
-            self._committed_revision[turn_id] = revision
-            logger.debug("Committed speculative turn %s revision %d", turn_id, revision)
-            self._condition.notify_all()
-            return True
+            return self._commit_locked(turn_id, revision)
 
     def commit_if_latest_after_reopen_grace(self, turn_id: str | None, revision: int | None) -> bool:
         if turn_id is None or revision is None:
             return True
         with self._condition:
             self._wait_for_reopen_gate_locked(turn_id, revision)
-            latest = self._latest_revision.get(turn_id, revision)
-            if revision != latest:
-                return False
-            self._committed_revision[turn_id] = revision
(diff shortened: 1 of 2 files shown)


andimarafioti (approved):
Good work, thank you ! 🙏

PR #395, opened 2026-07-30
Give every STT language an LLM language name

### LINKED ISSUE #394: --enable_lang_prompt silently emits no instruction for 17 of the default STT's 25 languages
## Summary

`--enable_lang_prompt` is documented as appending a *"Please reply to my message in ..."* instruction when `--language auto` detects the user's language. The instruction is gated on the resolved language **name**, not on the language code:

```python
# LLM/base_openai_compatible_language_model.py:560, and LLM/language_model.py:539
language_code, lang_name = resolve_auto_language(language_code)
if lang_name and self.enable_lang_prompt:
    active_chat.add_item(make_user_message(f"Please reply to my message in {lang_name}."))
```

`resolve_auto_language()` returns `lang_name = None` for any code missing from `WHISPER_LANGUAGE_TO_LLM_LANGUAGE` (`LLM/utils.py:32`), so the instruction is **silently skipped** — no warning, no log line.

That map has 12 entries and matches the Whisper handlers' `SUPPORTED_LANGUAGES`. But **Parakeet TDT v3 is the default STT** and reports 25 languages (`STT/parakeet_tdt_handler.py:41`), of which only 8 overlap. So 17 detectable languages get no instruction:

```
bg cs da et fi hr hu lt lv no ro ru sk sl sr sv uk
```

## Reproduction

No models needed — this isolates the resolution step:

```python
from speech_to_speech.LLM.utils import WHISPER_LANGUAGE_TO_LLM_LANGUAGE, resolve_auto_language
from speech_to_speech.STT.parakeet_tdt_handler import SUPPORTED_LANGUAGES

print(resolve_auto_language("sv-auto"))   # ('sv', None)  <- no instruction emitted
print(resolve_auto_language("de-auto"))   # ('de', 'german')

missing = [c for c in SUPPORTED_LANGUAGES if c not in WHISPER_LANGUAGE_TO_LLM_LANGUAGE]
print(len(SUPPORTED_LANGUAGES), len(WHISPER_LANGUAGE_TO_LLM_LANGUAGE), missing)
# 25 12 ['ru', 'uk', 'cs', 'sk', 'hu', 'ro', 'bg', 'hr', 'sl', 'sr', 'da', 'no', 'sv', 'fi', 'et', 'lv', 'lt']
```

End to end, this is the documented configuration from the README:

```bash
speech-to-speech --stt parakeet-tdt --language auto --enable_lang_prompt ...
```

Speak Swedish: STT transcribes it correctly and forwards `sv-auto`, the code reaches the LLM turn intact, but no language instruction is added. Only the instruction goes missing, so the flag quietly does nothing for those 17 languages — which is exactly the case it exists to serve, since the README notes it is there to help smaller models that don't infer the language from context.

## Why this looks like an oversight rather than a deliberate limit

- The README's Multi-Language Support table lists *"Parakeet TDT (default) | 25 European languages"* and describes `--enable_lang_prompt` with no mention of a narrower subset.
- `TTS/kokoro_handler.py:41` already has a `# Additional European languages from Parakeet - map to closest Kokoro language` block adding `ru`/`uk`, so extending coverage to Parakeet's set downstream was already the intent. The LLM map just wasn't updated alongside it.
- The map is still named `WHISPER_LANGUAGE_TO_LLM_LANGUAGE` and sized to the Whisper handlers, from before Parakeet became the default.

## Suggested fix

Add the 17 missing names, and add a test asserting every bundled STT handler's `SUPPORTED_LANGUAGES` is covered by the map so a handler gaining a language can't silently regress this again.

## Environment

- `speech-to-speech` at `cc37fe8` (also present in 0.2.11)
- Backend-independent: the same gate exists in both `language_model.py` and `base_openai_compatible_language_model.py`

Hotragn:
Fixes #394

## Problem

`--enable_lang_prompt` appends *"Please reply to my message in {name}."*, but the instruction is gated on the resolved language **name**, not the code:

```python
language_code, lang_name = resolve_auto_language(language_code)
if lang_name and self.enable_lang_prompt:
    active_chat.add_item(make_user_message(f"Please reply to my message in {lang_name}."))
```

`resolve_auto_language()` returns `None` for any code missing from `WHISPER_LANGUAGE_TO_LLM_LANGUAGE`, so the instruction is **silently skipped** — no warning, no log.

That map had 12 entries, sized to the Whisper handlers' `SUPPORTED_LANGUAGES`. Parakeet TDT v3 is the **default** STT and reports 25 languages, only 8 of which overlapped. So in the configuration the README documents, 17 detectable languages got no instruction at all:

```
bg cs da et fi hr hu lt lv no ro ru sk sl sr sv uk
```

```python
>>> resolve_auto_language("sv-auto")
('sv', None)          # no instruction emitted
>>> resolve_auto_language("de-auto")
('de', 'german')
```

The STT transcribes those languages correctly and the code reaches the LLM turn intact — only the instruction goes missing. So the flag quietly does nothing for 17 of 25 languages, which is precisely the case it exists for: the README notes it's there to help smaller models that don't infer the language from context.

This reads as an oversight rather than a deliberate limit — `TTS/kokoro_handler.py:41` already carries an `# Additional European languages from Parakeet` block mapping `ru`/`uk`, so covering Parakeet's set downstream was already the intent, and the README's table advertises the 25 languages with no mention of a narrower subset.

## Fix

Adds the 17 missing names. Deliberately minimal:

- No behaviour change when `--enable_lang_prompt` is off, which is the default.
- Nothing else consumes this map, so the blast radius is the instruction string.
- I left the `WHISPER_LANGUAGE_TO_LLM_LANGUAGE` name alone rather than renaming it to something backend-neutral — happy to rename if you'd prefer, but it seemed like avoidable churn for a bug fix.

I did **not** extend `WHISPER_LANGUAGE_TO_KOKORO_LANG`: it already falls back with `.get(language_code, self.lang_code)`, so unmapped languages degrade to the configured voice instead of failing. And Qwen3-TTS (the default) uses its own `--qwen3_tts_language`, defaulting to `auto`, so it isn't affected either way.

## Tests

**`tests/test_llm_utils.py`** — an invariant test: every bundled STT handler's `SUPPORTED_LANGUAGES` must be covered by the map, so a handler gaining a language can't silently regress this again. Handlers behind optional extras are skipped; the three with no optional top-level dependency are asserted importable, so the sweep can't pass by skipping everything. Against the previous map it fails with the exact gap:

```
AssertionError: speech_to_speech.STT.parakeet_tdt_handler can report
['bg', 'cs', 'da', 'et', 'fi', 'hr', 'hu', 'lt', 'lv', 'no', 'ro', 'ru', 'sk', 'sl', 'sr', 'sv', 'uk'],
which have no entry in WHISPER_LANGUAGE_TO_LLM_LANGUAGE, so --enable_lang_prompt
would emit no instruction for them
```

**`tests/test_language_prompt.py`** — end-to-end, not just dict keys. It drives the chat-completions backend with a faked OpenAI client and asserts the instruction actually reaches the outgoing request for every language Parakeet can report, that the flag still gates it, and that an unnameable code emits nothing rather than `"reply in None."`.

18 tests fail against the previous map. Full suite: 657 passed, 1 skipped. `ruff check`, `ruff format --check`, `mypy src/` clean.

### DIFF
--- a/src/speech_to_speech/LLM/utils.py
+++ b/src/speech_to_speech/LLM/utils.py
@@ -29,6 +29,13 @@ def remove_unspeechable(text: str) -> str:
     return SPEECHABLE_PATTERN.sub("", text)
 
 
+# Maps an STT language code to the language name used in the "Please reply ... in {name}"
+# prompt. Every language any bundled STT backend can report needs an entry here, otherwise
+# `--enable_lang_prompt` silently emits no instruction for it. The names are lowercase
+# because they are interpolated mid-sentence.
+#
+# `tests/test_llm_utils.py` asserts this covers the SUPPORTED_LANGUAGES of every bundled STT
+# handler, so adding a language to a handler without adding it here fails CI.
 WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
     "en": "english",
     "fr": "french",
@@ -42,6 +49,24 @@ WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
     "pl": "polish",
     "it": "italian",
     "nl": "dutch",
+    # The remaining languages Parakeet TDT v3 (the default STT) detects and reports.
+    "ru": "russian",
+    "uk": "ukrainian",
+    "cs": "czech",
+    "sk": "slovak",
+    "hu": "hungarian",
+    "ro": "romanian",
+    "bg": "bulgarian",
+    "hr": "croatian",
+    "sl": "slovenian",
+    "sr": "serbian",
+    "da": "danish",
+    "no": "norwegian",
+    "sv": "swedish",
+    "fi": "finnish",
+    "et": "estonian",
+    "lv": "latvian",
+    "lt": "lithuanian",
 }
 
 
--- a/tests/test_language_prompt.py
+++ b/tests/test_language_prompt.py
@@ -1,0 +1,127 @@
+"""End-to-end coverage for the ``--enable_lang_prompt`` language instruction.
+
+The instruction is gated on a language *name*, not on the language code:
+
+    language_code, lang_name = resolve_auto_language(language_code)
+    if lang_name and self.enable_lang_prompt:
+        active_chat.add_item(make_user_message(f"Please reply to my message in {lang_name}."))
+
+so any code missing from ``WHISPER_LANGUAGE_TO_LLM_LANGUAGE`` silently produces no
+instruction at all. Parakeet TDT is the default STT and reports 25 languages, so this
+asserts the instruction actually reaches the outgoing request for the ones it detects --
+not merely that the mapping dict has keys.
+
+The OpenAI client is faked, so this runs with no network and no GPU.
+"""
+
+from __future__ import annotations
+
+import queue
+import threading
+from types import SimpleNamespace
+
+import pytest
+from openai.types.realtime.realtime_session_create_request import RealtimeSessionCreateRequest
+
+import speech_to_speech.LLM.base_openai_compatible_language_model as base_mod
+from speech_to_speech.api.openai_realtime.runtime_config import RuntimeConfig
+from speech_to_speech.LLM.chat import Chat, make_user_message
+from speech_to_speech.LLM.chat_completions_language_model import ChatCompletionsApiModelHandler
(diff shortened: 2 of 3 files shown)


andimarafioti (approved):
Good catch, thanks!

PR #396, opened 2026-07-30
Add Qwen3-ASR STT backend (in-process + HTTP; in-process blocked on macOS)

nam157:
## Summary
- Adds `--stt qwen3-asr`, wiring `Qwen/Qwen3-ASR-1.7B` (and `Qwen/Qwen3-ASR-0.6B`) in through the `qwen-asr` PyPI package, following the same handler/arguments-class pattern as the existing Whisper/Paraformer/Parakeet STT backends.
- Adds `--stt qwen3-asr-http`, a second backend that talks to a separately-running Qwen3-ASR HTTP server instead of loading the model in-process — see "Known limitation" below for why this exists and is the recommended path on macOS.
- New optional extra: `pip install speech-to-speech[qwen3-asr]` (only needed for the in-process backend; `qwen3-asr-http` has no extra dependency beyond the existing `httpx` base dependency).
- New flags:
  - in-process: `--qwen3_asr_model_name`, `--qwen3_asr_device`, `--qwen3_asr_torch_dtype`, `--qwen3_asr_language`, `--qwen3_asr_max_new_tokens`, `--qwen3_asr_max_inference_batch_size`
  - HTTP: `--qwen3_asr_http_base_url`, `--qwen3_asr_http_timeout_s`
- Supports the 30 languages (plus Cantonese) Qwen3-ASR covers, with `--qwen3_asr_language auto` for automatic language detection (in-process backend; the HTTP backend relies on Qwen3-ASR's own auto-detect tag, see limitations).
- Adds `scripts/qwen3_asr_server.py`, a small Flask server that the HTTP backend talks to (see below for why this exists instead of `qwen-asr`'s own `qwen-asr-serve`), plus `docker-compose.qwen3-asr.yml` / `Dockerfile.cpu` / `docker/qwen3-asr/Dockerfile` for a CPU-only, no-NVIDIA-toolkit-needed way to run both sides as containers.
- Documented in the top-level README and `src/speech_to_speech/STT/README.md`.

## ⚠️ Known limitation: `--stt qwen3-asr` (in-process) does not work on macOS as of `qwen-asr==0.0.6`

`qwen-asr==0.0.6` hard-pins `transformers==4.57.6`, which conflicts with this project's
`transformers==5.6.2` pin on macOS (needed by Qwen3-TTS/mlx-audio) — installing the extra requires
`pip install --no-deps qwen-asr==0.0.6` plus its non-`transformers` deps.

On top of `transformers==5.6.2`, `qwen_asr`'s vendored modeling/config code
(`qwen_asr/core/transformers_backend/`) hits **at least four** separate transformers 4→5 breaking
changes it hasn't adapted to upstream — root-caused interactively together with a user testing this
branch live on a real macOS install (Python 3.12, Apple Silicon):

1. `check_model_inputs()` called as an optional-arg decorator factory (4.57.x style); 5.x made `func`
   required — `TypeError: check_model_inputs() missing 1 required positional argument`.
2. `Qwen3ASRConfig.__init__` reads `self.thinker_config` from inside `super().__init__(**kwargs)`
   (5.x's config validation now calls `get_text_config()` eagerly there), before that attribute is
   assigned — `AttributeError: 'Qwen3ASRConfig' object has no attribute 'thinker_config'`.
3. The `"default"` key was removed from `ROPE_INIT_FUNCTIONS` in 5.x (replaced by a
   `rope_parameters`-based API) — `KeyError: 'default'`. Re-registering it with the 4.57.6
   implementation works, but the replacement API also drops `mrope_interleaved`/`mrope_section`,
   meaning this specific patch risks *silently wrong* positional encoding rather than a crash.
4. `Qwen3ASRThinkerConfig` similarly ends up missing `pad_token_id` at model-build time —
   `AttributeError: 'Qwen3ASRThinkerConfig' object has no attribute 'pad_token_id'`. Unresolved.

`qwen3_asr_handler.py` applies runtime shims for (1)–(3) (see `_apply_transformers_5x_compat`; each
checks current state first, so they no-op once `qwen-asr` fixes things upstream), but model
construction still fails on (4). **Do not rely on `--stt qwen3-asr` on macOS** until `qwen-asr` adds
transformers 5.x support upstream. On Linux/CUDA, the base `transformers>=4.57.0` requirement is a
*range*, so `transformers==4.57.6` may resolve and run there without any of this — untested, would
appreciate someone with a CUDA box confirming.

## ✅ `--stt qwen3-asr-http` is the confirmed-working path on macOS

Rather than keep patching an increasingly deep well of incompatibilities (with real correctness risk
on #3 above), `--stt qwen3-asr-http` runs Qwen3-ASR as its own process in a separate virtualenv pinned
to `transformers==4.57.6` — the version it actually needs, no patches required — and the pipeline talks
to it over plain HTTP (`localhost`, negligible added latency).

**Important correction from an earlier version of this PR:** that separate process is
`scripts/qwen3_asr_server.py` (added in this PR), *not* `qwen-asr`'s own `qwen-asr-serve` command.
`qwen-asr-serve` turns out to be a thin wrapper around `vllm serve` and hard-requires the `vllm`
package — confirmed live: it raises `ModuleNotFoundError: No module named 'vllm'` on macOS even when
the transformers backend is otherwise requested, because `vllm` has no wheels for macOS/Apple Silicon
at all. `scripts/qwen3_asr_server.py` is a ~50-line Flask server wrapping the exact
`Qwen3ASRModel.from_pretrained(...).transcribe(...)` call already confirmed working end-to-end on Apple
Silicon MPS (direct Python API, no server), exposing the same OpenAI-chat-completions-shaped
`/v1/chat/completions` contract `Qwen3ASRHTTPSTTHandler` expects. Verified inside the built Docker image
with a mocked model: import succeeds, a POST round-trips the exact `"language X<asr_text>Y"` format the
handler's `_parse_asr_output` parses, and `/health` responds.

```bash
# Terminal 1 — separate venv, only for Qwen3-ASR
uv venv --python 3.12 .venv-qwen3-asr && source .venv-qwen3-asr/bin/activate
uv pip install qwen-asr==0.0.6
python scripts/qwen3_asr_server.py --device mps --host 127.0.0.1 --port 8000
```
```bash
# Terminal 2 — the normal speech-to-speech venv
speech-to-speech --stt qwen3-asr-http --qwen3_asr_http_base_url http://127.0.0.1:8000
```

Or, entirely via Docker (CPU-only, no NVIDIA Container Toolkit needed — see `docker-compose.qwen3-asr.yml`):

```bash
export OPENAI_API_KEY=...
docker compose -f docker-compose.qwen3-asr.yml up
python scripts/listen_and_play.py --host 127.0.0.1   # native, for mic/speaker access
```

Known gaps:
- Per-request language forcing isn't wired up for this path — `scripts/qwen3_asr_server.py` always calls
  `.transcribe(language=None)`, relying on Qwen3-ASR's own auto-detected language tag. Easy to add later
  (thread a `language` field through the request JSON) if needed.
- On a CUDA host with `vllm` installed, the *real* `qwen-asr-serve` would likely outperform this via
  vLLM's batching. Since this handler's HTTP contract only assumes an OpenAI-chat-completions-shaped
  response, swapping the server side later doesn't require touching `qwen3_asr_http_handler.py`.

Given this, I'd treat `--stt qwen3-asr` as **not macOS-ready / blocked upstream** and `--stt
qwen3-asr-http` as the one to actually merge/use — happy to split these into separate PRs if
maintainers would rather land the working HTTP path now and hold the in-process one, or land both
together with the limitation clearly flagged (current state of this PR).

## Test plan
- [x] `ast.parse` syntax-checked all new/modified Python files (no Python 3.10+/uv environment was available in the sandbox to run the full `uv sync && pytest`).
- [x] Added `tests/test_qwen3_asr_transcription_events.py` (in-process handler) and `tests/test_qwen3_asr_http_transcription_events.py` (HTTP handler), mirroring `tests/test_paraformer_transcription_events.py`.
- [x] The HTTP handler's `_parse_asr_output`/`_fix_char_repeats` helpers were executed standalone in the sandbox (no numpy/torch needed) against the four cases documented in `qwen_asr.inference.utils.parse_asr_output`'s own docstring — all passed.
- [x] Updated `tests/test_cli_defaults.py`'s `EXPECTED_FIELD_TYPES` to include both new `*_stt_handler_kwargs` fields on `ParsedArguments`.
- [x] `docker/qwen3-asr/Dockerfile` built successfully in the sandbox (CPU-only, ARM64); `scripts/qwen3_asr_server.py` verified inside that built image — imports cleanly, and its Flask route was exercised end-to-end with a mocked model, producing byte-for-byte the format the client parses.
- [x] End-to-end tested live on macOS (Python 3.12, Apple Silicon): in-process backend confirmed blocked (limitation above); the underlying `Qwen3ASRModel.from_pretrained(...).transcribe(...)` call confirmed working with a real transcription. `scripts/qwen3_asr_server.py` itself (the actual HTTP server) was verified via the Docker build + mocked-model test above, not yet re-run live end-to-end on the user's Mac since it replaced `qwen-asr-serve` partway through testing — a live re-confirmation would still be good before merging.
- [ ] Not run: `pytest`, `ruff check` (no compatible local environment) — please run CI/maintainer review.

### DIFF
--- a/Dockerfile.cpu
+++ b/Dockerfile.cpu
@@ -1,0 +1,39 @@
+FROM python:3.12-slim
+
+ENV PYTHONUNBUFFERED=1
+ENV PATH="/usr/src/app/.venv/bin:${PATH}"
+
+WORKDIR /usr/src/app
+
+# No NVIDIA base image here: this variant is for machines with no GPU (or a GPU Docker
+# can't reach, e.g. Docker Desktop on Apple Silicon has no Metal/MPS passthrough), running
+# everything on CPU. Use Dockerfile / Dockerfile.arm64 instead on a CUDA-capable host.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        git \
+        libportaudio2 \
+        libsndfile1 \
+    && rm -rf /var/lib/apt/lists/*
+RUN pip install --no-cache-dir uv
+
+# torch/torchaudio default to the CUDA-bundled wheel on Linux even with no GPU present;
+# pull from the CPU wheel index instead so `uv sync` doesn't download multiple GB it can't
+# use. index-strategy=unsafe-best-match lets uv pick torch from this extra index while still
+# resolving every other dependency from PyPI as usual.
+ENV UV_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu"
+ENV UV_INDEX_STRATEGY="unsafe-best-match"
+
+COPY pyproject.toml README.md LICENSE MANIFEST.in ./
+RUN uv sync --no-install-project --no-dev
+
+COPY . .
+RUN uv sync --no-dev
+
+# The default Qwen3-TTS GGML backend's qwentts-cpp-python wheel on PyPI targets CUDA 12.8;
+# this image has no GPU, so pull the CPU wheel instead (see README's "CUDA Note for
+# Qwen3-TTS" section).
+RUN uv pip install "qwentts-cpp-python==0.3.1+cpu" \
+    -f https://huggingface.co/datasets/andito/qwentts-cpp-python-wheels/tree/main/whl/cpu
+
+RUN python -c "import nltk; nltk.download('punkt_tab'); nltk.download('averaged_perceptron_tagger_eng')"
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ pip install "speech-to-speech[facebook-mms]"    # MMS TTS
 pip install "speech-to-speech[faster-whisper]"  # Faster Whisper STT
 pip install "speech-to-speech[whisper-mlx]"     # Lightning Whisper MLX STT on macOS
 pip install "speech-to-speech[paraformer]"      # Paraformer STT through FunASR
+pip install "speech-to-speech[qwen3-asr]"       # Qwen3-ASR STT
 pip install "speech-to-speech[mlx-lm]"          # mlx-vlm support for vision models on macOS
 ```
 
@@ -160,6 +161,7 @@ This installs the package in editable mode and makes the `speech-to-speech` CLI 
 | STT | [Lightning Whisper MLX](https://github.com/mustafaaljadery/lightning-whisper-mlx) | Apple Silicon | `whisper-mlx` |
 | STT | [MLX Audio Whisper](https://github.com/huggingface/mlx-audio) | Apple Silicon | built-in on macOS |
 | STT | [Paraformer](https://github.com/modelscope/FunASR) | CUDA / CPU | `paraformer` |
+| STT | [Qwen3-ASR](https://huggingface.co/Qwen/Qwen3-ASR-1.7B) | CUDA / CPU / Apple Silicon | `qwen3-asr` |
 | LLM | OpenAI-compatible API (`responses-api`, `chat-completions`) | hosted providers or self-hosted servers | built-in |
 | LLM | [Transformers](https://huggingface.co/models?pipeline_tag=text-generation&sort=trending) | CUDA / CPU | built-in |
 | LLM | [mlx-lm](https://github.com/ml-explore/mlx-lm) | Apple Silicon | built-in on macOS |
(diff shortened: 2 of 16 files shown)


andimarafioti:
Generally happy to add this model but some things in this PR are super odd. Like, why would you create a Dockerfile.cpu that does apparently nothing? And what is qwen3asr-http? IDK, it looks odd at a first glance

ebezzam:
pardon me jumping in! but there's a similar [PR](https://github.com/huggingface/speech-to-speech/pull/382), which uses the recent Transformers implementation. From our benchmarking on the [Open ASR Leaderboard](https://huggingface.co/spaces/hf-audio/open_asr_leaderboard), the Transformers implementation is faster 🤗  (can be used with torch compile)

andimarafioti:
Closing this since the other PR is better aligned with the project

andimarafioti:
Thank you for your effort!

PR #405, opened 2026-07-31
Stop whisper-mlx dropping turns in an unsupported language

### LINKED ISSUE #404: whisper-mlx: auto mode silently drops the user's turn when the detected language is outside SUPPORTED_LANGUAGES
## Summary

With `--stt whisper-mlx --language auto`, `LightningWhisperSTTHandler.process()` throws away the transcription whenever Whisper detects a language that is not in the handler's 12-entry `SUPPORTED_LANGUAGES` list:

```python
# STT/lightning_whisper_mlx_handler.py
language_code = transcription_dict["language"]
if language_code not in SUPPORTED_LANGUAGES:
    logger.warning(f"Whisper detected unsupported language: {language_code}")
    if self.last_language in SUPPORTED_LANGUAGES:      # reprocess with the last language
        transcription_dict = self.model.transcribe(audio, language=self.last_language)
    else:
        transcription_dict = {"text": "", "language": "en"}   # <- transcription discarded
```

Two distinct bad outcomes:

1. **The turn is dropped entirely.** `setup()` stores `--language auto` verbatim in `last_language`, and `"auto"` is not in `SUPPORTED_LANGUAGES`, so the `else` branch is the one taken on the first such utterance. `pred_text` becomes `""` and an empty transcription is emitted — the user spoke, and nothing reaches the LLM. The only trace is the `unsupported language` warning.
2. **The transcription is replaced by a wrong-language one.** Once any supported language has been seen, the audio is re-transcribed forcing that older language, discarding the correct result. Russian after German comes back as German.

Whisper transcribes these languages perfectly well — `SUPPORTED_LANGUAGES` here is a 12-entry list that governs the sticky fallback, not a statement about what the model can do. Checking the detected language against it, and then discarding good output, is the actual defect.

Affected languages are everything Whisper can detect beyond `en fr es zh ja ko hi de pt pl it nl` — Russian, Arabic, Turkish, Swedish, Ukrainian, and so on.

## Reproduction

```bash
speech-to-speech --stt whisper-mlx --stt_model_name large-v3 --language auto ...
```

Speak Russian (or any language outside the 12). Expected: the Russian transcription forwarded to the LLM. Actual: empty text on the first such turn, or German/whatever-was-last on later ones.

Isolated, without needing the model — the handler's own logic on a scripted result:

```python
transcription_dict = {"text": "Privet, kak dela?", "language": "ru"}   # what Whisper returned
# handler, auto mode, last_language == "auto"
# -> pred_text == ""      the turn is gone
```

## Relationship to other reports

This may account for some of the "language X is not recognized" reports where the transcription silently comes back empty or in the wrong language on macOS, e.g. #160 and #79. I have not confirmed those are the same root cause — the checkpoint and backend differ — so I'm not claiming they're duplicates.

## Suggested fix

Report the language that was actually transcribed, and never drop or re-transcribe the audio; let `SUPPORTED_LANGUAGES` govern only what becomes the sticky `last_language`. Reserve the fallback for when no language could be determined at all. Also normalize `--language auto` to `None` in `setup()` instead of storing it as a language code, as `WhisperSTTHandler` and `MLXAudioWhisperSTTHandler` already do.

This is the same resolution already agreed for the transformers-Whisper handler in review on #378 ("retain any actual detected language and reserve `last_language` fallback for cases where no language could be determined"), so applying it here keeps the backends consistent.

## Environment

- `speech-to-speech` at `5b443c8`
- Backend: `--stt whisper-mlx` (`lightning-whisper-mlx`), macOS / Apple Silicon
- The same 12-entry list and `-auto` suffix convention exist in the other Whisper handlers, but this handler is the one that drops the text

Hotragn:
Fixes #404

## Problem

With `--stt whisper-mlx --language auto`, the handler discards the transcription whenever Whisper detects a language outside its 12-entry `SUPPORTED_LANGUAGES`:

```python
if language_code not in SUPPORTED_LANGUAGES:
    if self.last_language in SUPPORTED_LANGUAGES:
        transcription_dict = self.model.transcribe(audio, language=self.last_language)
    else:
        transcription_dict = {"text": "", "language": "en"}   # <- transcription discarded
```

Two bad outcomes:

1. **The turn is dropped entirely.** `setup()` stores `--language auto` verbatim in `last_language`, and `"auto"` isn't in `SUPPORTED_LANGUAGES`, so the `else` branch is taken on the first such utterance. The user speaks and an empty transcription reaches the LLM. The only trace is a warning.
2. **The transcription is replaced by a wrong-language one.** Once a supported language has been seen, the audio is re-transcribed forcing that older language and the correct result is thrown away — Russian after German returns German.

Whisper transcribes these languages fine. `SUPPORTED_LANGUAGES` is a 12-entry list governing the sticky fallback, not a claim about model capability, so checking the detected language against it and then discarding good output is the defect. Affected: everything Whisper detects beyond `en fr es zh ja ko hi de pt pl it nl`.

## Fix

Same resolution already agreed for the transformers-Whisper handler in review on #378 — *"retain any actual detected language and reserve `last_language` fallback for cases where no language could be determined"* — applied here so the backends behave consistently:

- Report the language actually transcribed, even when outside `SUPPORTED_LANGUAGES`; never drop or re-transcribe. That list now only decides what becomes the sticky `last_language`.
- Reserve the `last_language` fallback for when no language could be determined at all, then `DEFAULT_LANGUAGE`.
- Normalize `--language auto` to `None` in `setup()` instead of storing it as a language code, matching `WhisperSTTHandler` and `MLXAudioWhisperSTTHandler`.
- Read `text`/`language` with `.get()` so degenerate model output can't raise `KeyError`.

There's no re-transcription path left, so the unsupported-language case now costs one inference instead of two. Behaviour for a pinned `--language` and for supported detected languages is unchanged.

## Tests

`tests/test_lightning_whisper_language_fallback.py` stubs the macOS-only `lightning_whisper_mlx` dependency and `torch.mps.empty_cache()` (which raises off Apple Silicon), so it runs on Linux CI with no optional extras. Coverage: the dropped turn, the destructive re-transcription, the sticky-fallback rules, `setup()` normalization, forced language, `language=None` vs `"auto"`, and degenerate model output.

11 of the 16 fail against the current code. The headline case fails with exactly the data loss:

```
assert result.text == "Privet, kak dela?"
E   AssertionError: assert '' == 'Privet, kak dela?'
```

The fixture pops the stub-built module from `sys.modules` afterwards, so it can't affect the `test_llm_utils.py` sweep that expects this handler to be unimportable without extras — verified by running both together.

Full suite: 742 passed, 1 skipped. `ruff check`, `ruff format --check`, `mypy src/` clean.

## Note

I deliberately left the unconditional `torch.mps.empty_cache()` in this handler alone — that's the subject of #145 / #370, and this handler is macOS-only so it isn't hit in practice.

### DIFF
--- a/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
+++ b/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
@@ -32,6 +32,8 @@ SUPPORTED_LANGUAGES = [
     "nl",
 ]
 
+DEFAULT_LANGUAGE = "en"
+
 
 class LightningWhisperSTTHandler(BaseSTTHandler):
     """
@@ -52,7 +54,9 @@ class LightningWhisperSTTHandler(BaseSTTHandler):
         self.device = device
         self.model = LightningWhisperMLX(model=model_name, batch_size=6, quant=None)
         self.start_language = language
-        self.last_language = language
+        # "auto" is a request to detect, not a language code, so it must never leak into
+        # last_language -- it would fail every SUPPORTED_LANGUAGES check downstream.
+        self.last_language = language if language != "auto" else None
 
         self.warmup()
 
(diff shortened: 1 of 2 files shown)


banuboy869-netizen:
dd2b24fa008ca39d9bc03e4f4298a6cdfa193460

banuboy869-netizen (commented):
No like

andimarafioti:
This is a very edge case and would be solved much more easily by just returning the transcription instead of removing it. I'll implement that as a fix instead.

PR #407, opened 2026-08-01
Expose Qwen3-TTS temperature/top_p/instruct to the realtime client

chenzhaoxuan0:
## Problem

The Qwen3-TTS handler never forwarded `temperature`, `top_p`, or `instruct` to the `faster-qwen3-tts` `generate_*_streaming` calls — they stayed locked at the library defaults (`temperature=0.9`, `top_p=1.0`). At that temperature the synthesized mood swings widely: a trailing ellipsis in an LLM reply can trigger an unwanted emotional tone (e.g. a crying/tearful delivery for a sentence ending in `…`), with no way for the client to tame it.

## Change

Expose the three generation knobs end-to-end so they are configurable per session:

**Backend** (`qwen3_tts_handler.py`)
- Add `temperature` / `top_p` instance attributes and reuse the existing `instruct`.
- New `_apply_session_tts_overrides()`, mirroring `_apply_session_voice_override()`: reads `temperature`/`top_p` from `audio.output` and `tts_instruct` from the session top level (per-response overrides take priority over the session default).
- Thread them into all three generate paths (`custom_voice`, `voice_design`, `voice_clone`).

**Demo client** (`index.html`, `main.js`, `style.css`, `ws/` + `rtc/` clients)
- New Settings controls: a "TTS mood instruction" textarea + `temperature` (0–1.5) and `top_p` (0.1–1) sliders, persisted in `localStorage` and applied live via `updateSession()` without a reconnect.

## How the params travel

The OpenAI SDK realtime pydantic models default to `extra='allow'`, so `audio.output.temperature` / `audio.output.top_p` and a top-level `session.tts_instruct` are accepted as `__pydantic_extra__` — no changes to the session model or `runtime_config.py` validation were needed. The handler reads them via `getattr`.

## Verification

- All three JS files pass `node --check`.
- Backend `import` of the handler succeeds.
- End-to-end: sent a `session.update` carrying `temperature=0.3`, `top_p=0.8`, `tts_instruct="…"` over WebSocket → no validation error, the Qwen3-TTS handler consumed them and synthesized audio (`response.done`, audio deltas received).

## Commits

Split by layer: backend handler / demo HTML+CSS / demo `main.js` / demo WS+RTC clients.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -274,6 +274,33 @@
             </div>
 
             <label class="field">
+              <span>TTS mood instruction</span>
+              <textarea id="tts-instruct" rows="2" placeholder="e.g. Speak in a calm, neutral tone"></textarea>
+              <small>Optional global instruction to the Qwen3-TTS voice (mood/tone). Leave empty for default.</small>
+            </label>
+
+            <div class="field-row">
+              <label class="field">
+                <span class="field-head">
+                  TTS temperature
+                  <span id="tts-temp-value" class="field-value">0.90</span>
+                </span>
+                <input id="tts-temperature" type="range" min="0" max="1.5" step="0.05" value="0.9"
+                       aria-label="TTS temperature" />
+                <small>Lower = steadier, less emotional speech. Higher = more expressive/varied.</small>
+              </label>
+              <label class="field">
+                <span class="field-head">
+                  TTS top_p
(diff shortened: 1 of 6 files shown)


andimarafioti:
The payload is not OpenAI Realtime protocol-compliant. Sorry, but this extension can't be supported at this time. Instead, you could pass this as parameters when you launch the repo.

PR #408, opened 2026-08-01
fix: keep non-ASCII punctuation and combining marks in remove_unspeechable

thejesh23:
## The bug

`SPEECHABLE_PATTERN` (`src/speech_to_speech/LLM/utils.py:18`) is an allow-list, but it enumerates only **ASCII** punctuation. Everything else falls into the negated class and is deleted — despite the docstring promising support for *"english, arabic, chinese, japanese, korean, etc."*.

`remove_unspeechable` runs on every audio turn (`base_openai_compatible_language_model.py:382,428`, `language_model.py:457`), and the cleaned string is what reaches the **TTS**, the **client transcript** (`lm_output_processor.py:124-135` builds `AssistantTextEvent` from it) and — on the local backends — the **committed chat history** (`language_model.py:561`). So the loss is user-visible and, for local backends, permanent across turns.

### 1. Sentence punctuation is dropped for most supported languages

```
'你好。今天天气怎么样？'  -> '你好今天天气怎么样'
'こんにちは。元気ですか？' -> 'こんにちは元気ですか'
'¿Cómo estás? ¡Bien!'     -> 'Cómo estás? Bien!'
'مرحبا، كيف حالك؟'        -> 'مرحبا كيف حالك'
```

ASCII `.` `,` `!` `?` are explicitly *kept*, so preserving sentence punctuation is clearly the intent — the non-ASCII forms just fall through. Losing them flattens TTS prosody and degrades downstream sentence splitting.

### 2. Combining marks are stripped from the middle of words

`\w` does not match combining marks, so they match the negated class too. Hindi is a supported language (`hi` in `WHISPER_LANGUAGE_TO_LLM_LANGUAGE`, and `parakeet-tdt`, the default STT, reports it):

```
'नमस्ते' -> 'नमसत'
```

The matras and the virama are gone — the word is mangled, not merely unpunctuated. The same applies to Arabic/Hebrew diacritics and to any NFD-decomposed Latin accent (`café` in NFD → `cafe`).

Repro against `main`:

```python
from speech_to_speech.LLM.utils import remove_unspeechable
assert remove_unspeechable("नमस्ते।") == "नमस्ते।"   # fails: 'नमसत'
```

## The fix

Keep the regex as a **fast pre-filter** — ASCII-only text still matches nothing and pays no per-character cost — and decide each matched character by its Unicode general category, keeping punctuation (`P*`), combining marks (`M*`) and currency (`Sc`).

Emoji and other symbols are `So`/`Sm`/`Sk` and are still dropped. Because the regex only *matches* characters that are removed today, the callback can only ever keep **more** — this is strictly a superset of the previous behaviour, so no character that used to be kept is now removed.

## Tests

Added to `tests/test_llm_utils.py`: zh/ja/ko/hi/es/de/ar sentence punctuation, combining marks (Devanagari + NFD Latin), and a guard that emoji and other symbols are still removed. All 8 new assertions fail on `main` and pass here.

`ruff check`, `ruff format --check`, `mypy src/` and the full `pytest tests/` suite are green (735 passed, 1 skipped — 726 before this change).

### DIFF
--- a/src/speech_to_speech/LLM/utils.py
+++ b/src/speech_to_speech/LLM/utils.py
@@ -1,6 +1,7 @@
 import base64
 import io
 import re
+import unicodedata
 from typing import Optional
 
 import requests  # type: ignore[import-untyped]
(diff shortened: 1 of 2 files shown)


andimarafioti (changes requested):
Against the PR’s own description:

- It does preserve the demonstrated international punctuation and combining marks.
- It improves TTS prosody.
- It does not fix downstream sentence splitting. The pinned NLTK tokenizer recognizes only `.`, `?`, and `!` as sentence endings, so CJK `。？！`, Arabic `؟`, and Devanagari `।` still leave the response as one utterance. [[NLTK 3.10 source](https://github.com/nltk/nltk/blob/v3.10.0/nltk/tokenize/punkt.py#L226)](https://github.com/nltk/nltk/blob/v3.10.0/nltk/tokenize/punkt.py#L226)

Thus it solves a punctuation-loss issue, but not a CJK/multilingual streaming-latency or sentence-segmentation issue.

## Finding

[P2] Keeping every `M*` Unicode category also preserves emoji components. In the [[new category filter](https://github.com/huggingface/speech-to-speech/blob/a1d4bc8248cf27df89681634d1da14cba83dc400/src/speech_to_speech/LLM/utils.py#L23-L42)](https://github.com/huggingface/speech-to-speech/blob/a1d4bc8248cf27df89681634d1da14cba83dc400/src/speech_to_speech/LLM/utils.py#L23-L42):

```python
remove_unspeechable("1️⃣") == "1️⃣"  # previously "1"
remove_unspeechable("☀️") == "\ufe0f"  # orphan variation selector
```

Variation selectors are `Mn`, and the keycap component is `Me`. This contradicts the PR’s guarantee that emoji remain removed and can put invisible orphan characters into transcripts/history. The tests only cover `✅`, which has no variation selector.

The implementation needs either an explicit policy for emoji-only marks or grapheme-aware filtering, plus regression tests for `☀️`, `❤️`, and `1️⃣`.

## Simplicity and current state

- For a CJK-only punctuation issue, the simpler solution is already merged in [[PR #430](https://github.com/huggingface/speech-to-speech/pull/430)](https://github.com/huggingface/speech-to-speech/pull/430), which adds `，。！？；：、` directly to the regex.
- For a genuinely multilingual punctuation/combining-mark issue, `unicodedata` is a reasonable standard-library approach once the emoji-mark bug is addressed.

PR #409, opened 2026-08-01
Fix(logging): Avoid configuring the root logger when importing optional handlers

salignatmoandal:
## Summary
- Remove import-time `logging.basicConfig` from optional TTS/STT handlers (`facebookmms`, `chatTTS`, `paraformer`) so importing a library module no longer overwrites the host app's root logger.
- Replace setup/cleanup `print()` calls with the module logger in `paraformer` and `faster_whisper`.
- Add regression tests that assert imports do not call `basicConfig`, sources have no `print()`, and setup/cleanup messages go through `caplog`.

## Why / example use case
An integrator may already configure logging for their app, then import an optional handler later:

```python
import logging
logging.basicConfig(level=logging.WARNING)

from speech_to_speech.TTS.facebookmms_handler import FacebookMMSTTSHandler
```

Previously, that import could call `logging.basicConfig(..., level=DEBUG)` and flood the process with debug logs. After this change, importing the handler stays neutral: the host keeps its WARNING config, and the handler still logs through `logging.getLogger(__name__)`.

## Test plan
- [x] `uv run pytest tests/test_handler_logging_hygiene.py -v`
- [ ] Confirm importing `speech_to_speech.TTS.facebookmms_handler` (or Paraformer/ChatTTS) does not change an existing `logging.WARNING` root config
- [ ] Spot-check that Paraformer setup and FasterWhisper cleanup still emit the expected info logs when the app has configured logging

### DIFF
--- a/src/speech_to_speech/STT/faster_whisper_handler.py
+++ b/src/speech_to_speech/STT/faster_whisper_handler.py
@@ -59,7 +59,7 @@ class FasterWhisperSTTHandler(BaseSTTHandler):
             logger.debug("no text detected. skipping...")
 
     def cleanup(self) -> None:
-        print("Stopping FasterWhisperSTTHandler")
+        logger.info("Stopping FasterWhisperSTTHandler")
         del self.model
 
     def adapt_gen_kwargs(self, gen_kwargs: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


andimarafioti (approved):
Good PR, thank you 🙏  Some parts of the code haven't been touched in a while, so there are weird small things like this, perfect for small PRs :)

PR #411, opened 2026-08-01
Preserve ordered assistant output across text and tool calls

### LINKED ISSUE #309: Support ordered assistant output parts across text and tool calls
## Problem

Assistant output is conceptually ordered: a model should be able to emit sequences like:

```text
text -> tool_call -> text -> tool_call -> text
```

Today the pipeline does not model that ordering cleanly. `LLMResponseChunk` carries `text` plus a list of `tools`, which works for simple `text + tool` cases but cannot represent arbitrary interleaving without losing ordering.

There is also a realtime lifecycle risk after tool-only chunks. `AUDIO_RESPONSE_DONE` is treated as a response boundary, and the router drains pending assistant/tool events before `response.done` so tools are not stranded. That is good for tool-only completion, but makes later assistant text after a tools-only chunk fragile: it may arrive after the response has been completed, reopen unexpectedly, or be attached to the wrong response boundary.

## Desired behavior

The system should preserve ordered assistant output parts and emit them in order, for example:

```text
assistant text
function_call
assistant text
function_call
assistant text
response.done
```

This should work without:

- dropping text after a tool call
- reordering transcript and function-call events
- closing the response before all ordered assistant parts are emitted
- accidentally reopening a new response for text that belongs to the same model turn
- relying on client-side timing between text, tool calls, TTS, and `AUDIO_RESPONSE_DONE`

## Current limitation

The current shape of `LLMResponseChunk(text, tools=[...])` is not expressive enough for ordered interleaving. It also encourages bundling text and tools into one event, where downstream handlers may emit transcript events before tool events regardless of the model's intended order.

## Possible direction

Introduce an ordered response-part abstraction, such as a stream of assistant parts:

```text
AssistantTextPart(text=...)
AssistantToolCallPart(tool=...)
AssistantTextPart(text=...)
```

The realtime router would then emit transcript/tool events from that ordered stream and defer `response.done` until both the ordered part stream and associated audio lifecycle have completed.

## Related context

This came up while reviewing PR #305. That PR can avoid entrenching a "drop text after tools" rule, but full support for `text + tool + text + tool + ...` needs a response lifecycle change beyond that PR's immediate scope.

iamsharduld:
Fixes #309

## Summary

- add an explicit ordered assistant-part model for text and function calls while retaining compatibility with the existing `text`/`tools` fields
- preserve interleaved local-model output such as `text -> tool_call -> text -> tool_call -> text` instead of dropping calls after the first
- keep realtime output indices and item IDs stable across separate chunks, including text-only completion events
- update voice and text tool prompts to permit multiple ordered tool calls
- add regression coverage for parser, LM-to-TTS routing, realtime audio events, text-only events, and cross-chunk ordering

## Root cause

`LLMResponseChunk` and `AssistantTextEvent` represented output as one text field plus a tool list, which imposed text-before-tools ordering. The local parser also explicitly skipped every tool call after the first, and realtime output indices restarted for each event.

## User impact

Assistant output now retains the model's intended text/tool sequence without dropping later calls, reordering client events, or closing text-only output items under the wrong response index.

## Validation

- `uv run --python /usr/bin/python3.12 pytest tests -q` — 731 passed, 1 skipped
- `.venv/bin/ruff check src tests`
- `.venv/bin/ruff format --check src tests`
- `.venv/bin/mypy src/`
- `git diff --check`

### DIFF
--- a/src/speech_to_speech/LLM/language_model.py
+++ b/src/speech_to_speech/LLM/language_model.py
@@ -49,6 +49,7 @@ from speech_to_speech.LLM.voice_prompt import build_voice_system_prompt
 from speech_to_speech.pipeline.cancel_scope import CancelScope
 from speech_to_speech.pipeline.handler_types import LLMIn, LLMOut
 from speech_to_speech.pipeline.messages import (
+    AssistantToolCallPart,
     EndOfResponse,
     GenerateResponseRequest,
     LLMResponseChunk,
@@ -308,7 +309,7 @@ class BaseLanguageModelHandler(BaseHandler[LLMIn, LLMOut], ABC):
         """
         chunks: list[LLMResponseChunk] = []
 
-        if ctx.enter_code and ctx.enter_code in printable_text:
+        while ctx.enter_code and ctx.enter_code in printable_text:
             idx = printable_text.index(ctx.enter_code)
             before = printable_text[:idx]
             code_and_after = printable_text[idx:]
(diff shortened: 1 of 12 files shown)


andimarafioti (commented):
I merged this into another branch to improve it. Thank you for the push!

PR #413, opened 2026-08-03
Send session.updated after session.update

hrqiang:
## Summary

- reply with `session.updated` after a successful `session.update`
- add `SessionHandler.build_session_updated`, mirroring the existing `build_session_created`

## Why

The protocol acknowledges an accepted `session.update`. From the [Realtime conversations guide](https://developers.openai.com/api/docs/guides/realtime-conversations):

> When the session has been updated, the server will emit a `session.updated` event with the new state of the session.

The server applied the incoming update but never sent that acknowledgement, so a client that waits for it before streaming audio had nothing to wait on and sat idle until it gave up and closed the connection.

## User impact

Clients that treat `session.updated` as the go-ahead for audio now proceed instead of stalling. Clients that ignore the event are unaffected, since nothing else in the exchange changed.

## Validation

- `pytest -q tests/` — 728 passed, 1 skipped
- `ruff check .` — passed
- over a WebSocket with the `openai` SDK: `session.created`, then `session.updated` echoing the requested voice back

### DIFF
--- a/src/speech_to_speech/api/openai_realtime/handlers/session.py
+++ b/src/speech_to_speech/api/openai_realtime/handlers/session.py
@@ -6,6 +6,7 @@ from typing import Optional
 from openai.types.realtime import (
     RealtimeErrorEvent,
     SessionCreatedEvent,
+    SessionUpdatedEvent,
     SessionUpdateEvent,
 )
 from openai.types.realtime.realtime_transcription_session_create_request import (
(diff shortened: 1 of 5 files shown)


andimarafioti (commented):
I merged this into an intermediate branch, it is good work, but mypy was failing and there was another detail I wanted to fix. Thank you!

PR #415, opened 2026-08-03
Backfill a placeholder user turn when generating with no user input yet

hrqiang:
## Summary

- backfill a placeholder user turn when a generation request carries no user message yet
- apply it in both the text/mlx-lm and vision-language paths, immediately before `apply_chat_template`
- match the role `Chat.to_transformers_chat()` emits (`"user"`) rather than the configurable `--user_role`

## Why

Some chat templates require at least one plain user-role message. Qwen3.5's scans the history backwards and raises `TemplateError: No user query found in messages.` when it finds none. Qwen3-4B-Instruct's template has no such requirement, so this only appeared after switching models.

Nothing in the protocol makes a user turn a precondition for generating. `response.create` accepts an empty `input` array specifically to "remove existing context" ([Realtime conversations guide](https://developers.openai.com/api/docs/guides/realtime-conversations)), so an empty context is a valid state to generate from. The constraint is the model's template rather than the API, and it rejects a legitimate request: a bare `response.create` used to have the assistant open a conversation from its system instructions, before the user has spoken.

## User impact

Assistant-opens-the-conversation flows work against models whose templates demand a user turn. Histories that already contain one serialize unchanged, and the placeholder exists only in the per-turn copy handed to the template, never in stored conversation state.

## Validation

- `pytest -q tests/` — 733 passed, 1 skipped
- `ruff check .` — passed
- over a WebSocket: `response.create` on an empty conversation returns a greeting instead of failing

### DIFF
--- a/src/speech_to_speech/LLM/language_model.py
+++ b/src/speech_to_speech/LLM/language_model.py
@@ -79,6 +79,24 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+# Some chat templates (e.g. Qwen3.5's) hard-require at least one user-role
+# message and raise a jinja2 TemplateError otherwise. OpenAI's Realtime API
+# allows a response to be generated from system instructions alone (e.g. an
+# initial greeting before the caller has spoken), so we backfill a minimal
+# placeholder turn to keep locally-loaded models with that requirement working.
+_PLACEHOLDER_USER_TURN = "..."
+# Must match the role Chat.to_transformers_chat() emits for user messages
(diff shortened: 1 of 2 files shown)


andimarafioti (changes requested):
This is too broad. Most models don't need this for their chat templates to work and will actually have degraded performance for a user turn with "...". I'm also not sure I understand when this happens. Either you start the conversation with a greeting message from the user, or you wait until there's the first transcription from the user. At no point we send messages to the LLM without a user turn (even after compaction we retain a user turn).
Did you actually find this using the pipeline, or is it a protocol edge case?

PR #416, opened 2026-08-03
Populate response.done's output with the generated message and function calls

hrqiang:
## Summary

- populate `response.done`'s `response.output` with the items the response produced
- add `pending_function_calls` to `ConnState`, plus a `_build_output_items` helper that assembles the list
- accumulate audio transcripts as well as streaming them, so the assistant message item carries the full text
- derive item status from the response status, and reuse the `item_id` from each call's arguments event

## Why

`_build_response()` constructed a `RealtimeResponse` without ever setting `output`, so `response.done` always reported `null`. Function calls went out only on `response.function_call_arguments.done`, which is an incremental event. From the [Realtime conversations guide](https://developers.openai.com/api/docs/guides/realtime-conversations):

> `response.done` will also have the complete data we need to call our function.

A client reading the final state rather than tracking the deltas therefore saw no tool call at all, and never invoked it.

## User impact

Tool calls are visible to clients that read `response.done`, which previously meant following the streaming deltas instead. A cancelled or failed response reports its items as `incomplete` rather than claiming they completed, and a response that produced nothing reports an empty `output` instead of `null`.

## Validation

- `pytest -q tests/` — 732 passed, 1 skipped
- `ruff check .` — passed
- over a WebSocket with a registered tool: `response.done.response.output` carries the assistant message and the function call, with the call's `id` matching its arguments event


andimarafioti:
plz fix ci, then I'll look. seems like a fine idea to me

andimarafioti (approved):
Cool! Looks good but I'll fix a couple of things my agent found before merging to main. Thank you 🙏

PR #420, opened 2026-08-03
Fix(TTS ): Align Facebook MMS language keys with Whisper codes

salignatmoandal:
hi,

This is a small hygiene fix in an older optional TTS path: a few Whisper→MMS language keys were mistyped, so STT language codes never matched and voices fell back to English (:

## Summary
- Fix mistyped Whisper-style keys in `WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE` (`bu`→`bg`, `po`→`pl`, `tu`→`tr`, `tg`→`tl`, `ma`→`mr`, `zl`→`ms`, `ka`→`kn`).
- Keep Facebook MMS model suffixes unchanged; only the STT-facing lookup keys are corrected.
- Add regression tests so non-Whisper keys cannot land in the map again.

## Why / example use case
STT backends emit Whisper codes (e.g. `pl`). Facebook MMS looks them up to load `facebook/mms-tts-*`.

Keys follow [openai/whisper `tokenizer.py`](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py) (e.g. `kn` = Kannada, `ka` = Georgian).

When STT detects Polish (`pl`), the map used to look for the non-Whisper key `po`, hit a `KeyError`, and fall back to English. It now resolves `pl` → `pol` → `facebook/mms-tts-pol`.

## Test plan
- [x] `uv run pytest tests/test_facebookmms_language_map.py -v`
- [ ] Confirm `WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE["pl"] == "pol"` (and `bg` / `tr` / `tl` / `mr` / `ms` / `kn`)
- [ ] Optional: with `--tts facebook-mms`, a Polish turn no longer falls back to English solely because of a map miss

### DIFF
--- a/src/speech_to_speech/TTS/facebookmms_handler.py
+++ b/src/speech_to_speech/TTS/facebookmms_handler.py
@@ -31,7 +31,7 @@ WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE = {
     "ar": "ara",  # Arabic
     "hy": "hyw",  # Armenian
     "az": "azb",  # Azerbaijani
-    "bu": "bul",  # Bulgarian
+    "bg": "bul",  # Bulgarian
     "ca": "cat",  # Catalan
     "nl": "nld",  # Dutch
     "fi": "fin",  # Finnish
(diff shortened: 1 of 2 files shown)


andimarafioti (approved):
Good work! thank you

PR #424, opened 2026-08-03
Report the spoken language from the faster-whisper backend

### LINKED ISSUE #423: faster-whisper backend never reports a language, so --enable_lang_prompt and per-turn TTS voice switching cannot work
## Summary

`FasterWhisperSTTHandler.process()` unpacks the detection result and then never uses it:

```python
# STT/faster_whisper_handler.py
segments, info = self.model.transcribe(vad_audio.audio, **self.gen_kwargs)
...
yield Transcription(
    text=pred_text,
    turn_id=vad_audio.turn_id,          # <- no language_code
    turn_revision=vad_audio.turn_revision,
    speech_stopped_at_s=vad_audio.created_at_s,
)
```

`info` is faster-whisper's `TranscriptionInfo` and carries `.language` and `.language_probability`. It is discarded, so `Transcription.language_code` keeps its `None` default.

Downstream, `resolve_auto_language(None)` returns `(None, None)`, and both LLM backends gate the language instruction on the name:

```python
language_code, lang_name = resolve_auto_language(language_code)
if lang_name and self.enable_lang_prompt:
    active_chat.add_item(make_user_message(f"Please reply to my message in {lang_name}."))
```

So with `--stt faster-whisper`:

- `--enable_lang_prompt` can **never** emit an instruction, for any language, silently.
- TTS handlers that switch per turn on `language_code` (e.g. Kokoro's `WHISPER_LANGUAGE_TO_KOKORO_LANG.get(language_code, self.lang_code)`) always take the fallback.

This happens even when the language is explicitly pinned via `--faster_whisper_stt_gen_language de` — the code is passed to `transcribe()` but never reported onward.

## Second issue: `auto` is not a usable value for this backend

`--faster_whisper_stt_gen_language auto` reaches `transcribe(language="auto")` verbatim. faster-whisper auto-detects only when `language` is absent, and rejects language names it doesn't recognise, so `auto` is not usable here. There is currently no documented way to ask this backend to detect the language, unlike the other STT backends where `--language auto` is the documented switch.

## Reproduction

```python
from speech_to_speech.LLM.utils import resolve_auto_language
# what the handler yields today, for any spoken language:
print(resolve_auto_language(None))     # (None, None)  -> no instruction, ever
```

End to end:

```bash
speech-to-speech --stt faster-whisper --faster_whisper_stt_gen_language de --enable_lang_prompt ...
```

Speak German. Expected: the turn carries `de` and the LLM is told to reply in German. Actual: `language_code` is `None` and no instruction is added.

## Suggested fix

Report the language on the `Transcription`: a pinned language as-is (since `transcribe()` ran with it), otherwise `info.language` tagged `-auto`, matching the convention the other Whisper handlers already use. And normalize `auto` to "no language passed" so auto-detection actually works for this backend rather than raising.

## Environment

- `speech-to-speech` at `5b443c8`
- Backend: `--stt faster-whisper` (`faster-whisper` optional extra)
- Independent of the LLM backend: the same gate exists in `language_model.py` and `base_openai_compatible_language_model.py`

Hotragn:
Fixes #423

## Problem

`process()` unpacks the detection result and never uses it:

```python
segments, info = self.model.transcribe(vad_audio.audio, **self.gen_kwargs)
```

`info` carries `.language` and `.language_probability`, but the handler yields `Transcription` with no `language_code`. The field defaults to `None`, and `resolve_auto_language(None)` returns `(None, None)`, so both LLM backends' gate

```python
if lang_name and self.enable_lang_prompt:
```

**could never fire** with `--stt faster-whisper`, and TTS handlers that switch voice per turn on `language_code` always took their fallback. This held even when the language was explicitly pinned with `--faster_whisper_stt_gen_language de`: the code reached `transcribe()` but was never reported onward.

Separately, `--faster_whisper_stt_gen_language auto` reached `transcribe(language="auto")` verbatim. faster-whisper auto-detects only when `language` is absent and rejects names it doesn't recognise, so `auto` wasn't a usable value and there was no way to ask this backend to detect — unlike every other STT backend, where `--language auto` is the documented switch.

## Fix

- **Report the language.** A pinned language as-is, since `transcribe()` ran with it; otherwise the detected language tagged `-auto`, so `resolve_auto_language()` treats it as changeable between turns. Same convention the other Whisper handlers already use.
- **Make `auto` work.** Normalize `auto` (and `""`, `none`, `null`, non-strings) to omitting `language` entirely, which is how faster-whisper auto-detects. Whitespace around a pinned code is stripped.
- Log the detected language and its probability at debug level.
- Document `auto` in the `--faster_whisper_stt_gen_language` help.

Behaviour with the default `en` is unchanged apart from the language now being reported — which is the point, since that's what makes `--enable_lang_prompt` work here.

## Tests

`tests/test_faster_whisper_language_reporting.py` stubs the optional `faster_whisper` dependency, so it runs on CI without installing the extra. Coverage: pinned and detected paths, that the reported code actually resolves to an LLM language name, every auto sentinel, absent/non-string language, and degenerate detection output (`info` without `.language`, empty string).

13 of the 18 fail against the current code. The headline case fails with exactly the user-visible consequence:

```
assert resolve_auto_language(language_code) == ("de", "german")
E   AssertionError: assert (None, None) == ('de', 'german')
```

The fixture pops the stub-built module from `sys.modules` afterwards so it can't leak into other tests.

Full suite: 744 passed, 1 skipped. `ruff check`, `ruff format --check`, `mypy src/` clean.

## Note on overlap

#374 also touches this file, in `adapt_gen_kwargs`. My change is in `setup()` and `process()`, so they should merge cleanly, but happy to rebase on whichever lands first.


andimarafioti (changes requested):
### Findings

1. **[P1] The reported language can differ from the language actually used.**  
   [`_resolve_language()` returns the requested language without consulting `info.language`](https://github.com/huggingface/speech-to-speech/blob/776d0b4420b778d60f7e3f259ba240d9786b6e31/src/speech_to_speech/STT/faster_whisper_handler.py#L64-L65). With the default `tiny.en` model, faster-whisper overrides a requested non-English language to `en` and records that in `TranscriptionInfo.language` ([[faster-whisper v1.0.3 source](https://github.com/SYSTRAN/faster-whisper/blob/v1.0.3/faster_whisper/transcribe.py#L430-L438)](https://github.com/SYSTRAN/faster-whisper/blob/v1.0.3/faster_whisper/transcribe.py#L430-L438)). The PR would nevertheless report `de`, causing an incorrect German LLM prompt and TTS voice.

   Always report `info.language`; use the requested mode only to decide whether to append `-auto`. Add a regression where the request is `de` but `info.language == "en"`. Also document that auto-detection needs a multilingual model—`tiny.en` always resolves to English.

2. **[P2] Setup consumes the caller’s language configuration.**  
   [`[self.gen_kwargs.pop("language")](https://github.com/huggingface/speech-to-speech/blob/776d0b4420b778d60f7e3f259ba240d9786b6e31/src/speech_to_speech/STT/faster_whisper_handler.py#L35-L40)`](https://github.com/huggingface/speech-to-speech/blob/776d0b4420b778d60f7e3f259ba240d9786b6e31/src/speech_to_speech/STT/faster_whisper_handler.py#L35-L40) operates on the original dictionary because `adapt_gen_kwargs()` currently mutates and returns it. I reproduced:

   ```text
   input:  {"return_timestamps": false, "language": "auto"}
   after:  {"without_timestamps": true}
   ```

   This extends the exact caller-mutation problem covered by [[PR #374](https://github.com/huggingface/speech-to-speech/pull/374)](https://github.com/huggingface/speech-to-speech/pull/374). Rebase after #374 or normalize a copy, and assert that the source dictionary remains unchanged.

3. **[P2] `--enable_lang_prompt` remains broken for most faster-whisper languages.**  
   The PR forwards every detected code at [[line 72](https://github.com/huggingface/speech-to-speech/blob/776d0b4420b778d60f7e3f259ba240d9786b6e31/src/speech_to_speech/STT/faster_whisper_handler.py#L67-L72)](https://github.com/huggingface/speech-to-speech/blob/776d0b4420b778d60f7e3f259ba240d9786b6e31/src/speech_to_speech/STT/faster_whisper_handler.py#L67-L72), but faster-whisper accepts 100 language codes while the LLM name map contains only 29. For example, `ar-auto` becomes `("ar", None)`, so the existing language-prompt gate still emits nothing. This only partially solves [[issue #423](https://github.com/huggingface/speech-to-speech/issues/423)](https://github.com/huggingface/speech-to-speech/issues/423).

   Provide complete Whisper code-to-name coverage or explicitly validate a supported subset, and add faster-whisper to the existing STT language-coverage test.

[P3] Treating `""`, `"none"`, `"null"`, and arbitrary non-strings as auto-detection goes beyond the issue and silently hides invalid configuration. Supporting case-insensitive `"auto"` plus programmatic `None` is simpler and matches the documented interface.

Hotragn:
All four findings addressed in `b2d039a`, rebuilt on current `main`. Thanks — P1 was a real bug and I'd argued myself into it.

## P1 — report the language actually used

You're right, and this is the same defect class the PR set out to fix, reintroduced. I reasoned "a pinned language is authoritative because `transcribe()` ran with it", which is simply false for English-only checkpoints — I should have read faster-whisper's source instead of trusting the API contract.

`_resolve_language()` now always reports `info.language`. The request only decides whether it's auto-detect mode, i.e. whether to append `-auto`. A mismatch logs a warning naming the English-only-checkpoint cause, and the CLI help documents that auto-detection needs a multilingual model.

Regressions added:

- `test_reports_the_transcribed_language_not_the_requested_one` — request `de`, `info.language == "en"` → reports `en`
- `test_auto_suffix_depends_on_the_request_not_the_detection` — pinned request never gets `-auto` even when the model picks a different language

## P2 — setup no longer consumes the caller's dict

Reproduced exactly as you described. `setup()` now normalizes a copy, so it's correct independently of #374 and stays correct once #374 lands. Two tests assert the source dict is unchanged, for the pinned and auto paths.

## P2 — complete Whisper coverage

The map now covers Whisper's full ~100-language set, and the handler declares `SUPPORTED_LANGUAGES` so the existing coverage test picks this backend up.

Two things I was careful about there:

1. **Not vacuous.** My first attempt derived `SUPPORTED_LANGUAGES` from the name map, which would have made the test always pass. It's now an explicit literal, so the test compares two independent sources. Verified by deleting one entry:
   ```
   AssertionError: speech_to_speech.STT.faster_whisper_handler can report ['ar'],
   which have no entry in WHISPER_LANGUAGE_TO_LLM_LANGUAGE, so --enable_lang_prompt
   would emit no instruction for them
   ```
2. **Actually runs on CI.** `faster-whisper` isn't in the dev group, so the new check would have silently skipped everywhere. The coverage test now stubs the optional import, and `faster_whisper_handler` is in `_ALWAYS_IMPORTABLE` so a skip fails loudly instead of passing quietly.

`ar`, `tr`, `he`, `th`, `yue` now resolve to names, with a parametrized test asserting each yields a non-empty name.

## P3 — narrower sentinels

Agreed, those were defensive additions I couldn't justify. Only `None` and case-insensitive `"auto"` mean auto-detect now; `""`, `"none"`, `"null"` and non-strings pass through so faster-whisper rejects them loudly. `test_invalid_language_is_passed_through_rather_than_silently_auto` pins that.

## Notes

- `start_language` is declared as a class attribute, because `tests/test_whisper_progressive_transcription.py` builds this handler via `object.__new__` without `setup()` — my first cut broke those three tests.
- Full suite: **1169 passed, 1 skipped**; ruff, format and mypy clean. One earlier full-suite run showed `test_is_latest_after_stability_window_survives_cancelled_reopen_candidate` failing; it passes isolated (×3), for the whole file, and on a repeat full run, and it's untouched by this PR — looks like a timing-sensitive test under parallel load rather than anything here, but flagging it in case it's a known flake.

andimarafioti (approved):
Thank you for this! it's good work! since main changed, I need to fix some conflicts, but I'll do those independently to not block this further. I'll merge to main with your changes promptly.

PR #425, opened 2026-08-04
Add content to Transformers tool-call messages

### LINKED ISSUE #414: MLX tool-call follow-up fails when assistant tool message omits content
## Environment

- `speech-to-speech`: 0.2.11
- macOS 26.5.2 (arm64)
- Python 3.12
- LLM backend: `mlx-lm`
- Model: `mlx-community/Qwen3-14B-4bit`
- Mode: OpenAI Realtime-compatible server

The service was started with:

```bash
speech-to-speech \
  --device mps \
  --stt paraformer \
  --llm_backend mlx-lm \
  --model_name mlx-community/Qwen3-14B-4bit \
  --tts qwen3
```

## What happened

The model successfully generated a function call and the Realtime client successfully returned a `function_call_output`. When the client requested the follow-up response, `apply_chat_template()` failed before the model could generate the final answer:

```text
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'content'
speech_to_speech.api.openai_realtime.service - INFO - Response failed: Language model generation failed: 'dict object' has no attribute 'content'
speech_to_speech.api.openai_realtime.handlers.response - INFO - Response done (status=failed) — audio=0.00s
```

This means the first tool-call generation succeeds, but the response after the tool result cannot be generated or spoken.

## Root cause

`Chat.to_transformers_chat()` serializes a `RealtimeConversationItemFunctionCall` through `TransformersFunctionCallMessage`. That model currently has `role` and `tool_calls`, but no `content` field, so the assistant tool-call history entry becomes:

```json
{
  "role": "assistant",
  "tool_calls": []
}
```

The Qwen3-14B chat template explicitly supports `message.tool_calls`, but also reads `message.content` unconditionally. The template therefore fails while rendering the conversation history. It expects an assistant tool-call message shaped like:

```json
{
  "role": "assistant",
  "content": "",
  "tool_calls": []
}
```

This is likely to affect other Hugging Face chat templates that assume every assistant message has a `content` key.

## Expected behavior

speech-to-speech should normalize assistant tool-call messages so a model-specific chat template can render a valid tool-call follow-up. The returned tool result should be passed to the LLM and the final text/audio response should be generated.

## Suggested fix

Add an empty content field to the Transformers assistant tool-call message:

```python
class TransformersFunctionCallMessage(BaseModel):
    role: Literal["assistant"] = "assistant"
    content: str = ""
    tool_calls: list[TransformersToolCall]
```

A regression test could build a chat containing user message → function call → function output, assert that the assistant tool-call entry includes `content == ""`, and render it with a template that accesses `message.content` unconditionally.

Mayankpratapsingh022:
## Summary

Fixes #414

- add a `content` field defaulting to `""` on `TransformersFunctionCallMessage`
- add regression tests for the serialized shape and for rendering through a template that reads `content`

## Why

`TransformersFunctionCallMessage` was the only model in the `TransformersChatMessage` union without a `content` field. Its four siblings all have one, so `to_transformers_chat()` emitted assistant tool-call entries with no `content` key:

```python
{"role": "assistant", "tool_calls": [...]}     # before
{"role": "assistant", "content": "", "tool_calls": [...]}   # after
```

Chat templates read `message.content` on every assistant message, tool-call turns included. The Qwen3 template concatenates it directly:

```jinja
{{- '<|im_start|>' + message.role + '\n' + content }}
```

Rendering a history that contained a tool call therefore raised `UndefinedError`, and the follow-up response after the tool result was never generated.

Worth noting why this stayed hidden: the same template guards with `{%- if '</think>' in message.content %}` two lines earlier, and `in` against an undefined value returns `False` rather than raising. Printing `{{ message.content }}` is also silent, rendering as an empty string. Only concatenation and method calls raise, which is why the existing tests passed straight through the bug.

## Reproduction

Against the real `Chat` class and the real Qwen3 tokenizer, for a `user -> function_call -> function_call_output` history:

Before:

```
assistant tool-call entry has 'content' key: False
apply_chat_template() FAILED
  jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'content'
```

After, generating with `mlx-community/Qwen3-0.6B-4bit`:

```
apply_chat_template(): OK
model output: The weather in Paris is currently 18°C with clear skies.
```

The issue reports this on `Qwen3-14B-4bit`. The failure is in prompt construction, before any weights load, so a small model of the same template family reproduces it identically.

## Validation

macOS 26 on Apple Silicon, `uv sync --group dev`:

- `pytest tests/`: 767 passed, 1 skipped
- `ruff check src/ tests/`: passed
- `ruff format --check src/ tests/`: 135 files already formatted
- `mypy src/`: no issues found in 93 source files

To confirm the tests actually cover the defect, I reverted `chat.py` to `main` and kept the new tests:

- `pytest tests/`: 3 failed, 764 passed, 1 skipped

The three failures are the three new tests, each raising the `UndefinedError` from the issue report. No other test changes behavior either way.

### DIFF
--- a/src/speech_to_speech/LLM/chat.py
+++ b/src/speech_to_speech/LLM/chat.py
@@ -728,6 +728,9 @@ class TransformersAssistantMessage(BaseModel):
 
 class TransformersFunctionCallMessage(BaseModel):
     role: Literal["assistant"] = "assistant"
+    # Chat templates read `message.content` on every assistant message, tool-call
+    # turns included, so the key must be present even with no text of its own.
+    content: str = ""
     tool_calls: list[TransformersToolCall]
 
 
--- a/tests/test_chat.py
+++ b/tests/test_chat.py
@@ -779,6 +779,43 @@ class TestToTransformersChat:
         assert result[3]["name"] == "action"
         assert result[4] == {"role": "assistant", "content": "All set."}
 
+    def test_function_call_carries_empty_content(self):
+        chat = Chat(size=5)
+        chat.add_item(_fc("c1", "search", '{"query": "test"}'))
+        chat.add_item(_fco("c1", "ok"))
+        entry = chat.to_transformers_chat()[0]
+        assert entry["content"] == ""
+
+    def test_every_assistant_entry_exposes_content(self):
+        chat = Chat(size=10)
+        chat.add_item(_user("Do it"))
+        chat.add_item(_fc("c1", "action", '{"a": 1}'))
+        chat.add_item(_fco("c1", "done"))
+        chat.add_item(_assistant("All set."))
+
+        assistant_entries = [m for m in chat.to_transformers_chat() if m["role"] == "assistant"]
+        assert len(assistant_entries) == 2
+        assert all("content" in m for m in assistant_entries)
+
+    def test_function_call_renders_in_template_reading_content(self):
+        """Chat templates read ``content`` on every assistant message, tool calls included.
+
+        Concatenation mirrors what the Qwen3 template does; a missing key would
+        leave an undefined value here and raise rather than render empty.
+        """
+        sandbox = pytest.importorskip("jinja2.sandbox")
+
+        chat = Chat(size=5)
+        chat.add_item(_user("What's the weather?"))
+        chat.add_item(_fc("c1", "get_weather", '{"city": "Paris"}'))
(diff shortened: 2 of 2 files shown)


andimarafioti (approved):
Great work!

PR #427, opened 2026-08-05
Use Realtime as the only pipeline engine

### LINKED ISSUE #421: Use Realtime as the only pipeline engine
## Decision

Move directly to one architecture. The Realtime engine becomes the only pipeline engine.

This is a cutover, not a staged migration:

- no deprecation period;
- no compatibility bridge;
- no parallel legacy and Realtime implementations;
- no long-lived feature flags;
- no attempt to preserve `raw-websocket` or TCP `socket` mode.

The implementation can use multiple commits for reviewability, but the merged change should leave the repository in the final architecture.

## Problem

The project currently has two materially different execution paths:

- `realtime` builds a `RealtimeService`/pipeline unit with speculative turn tracking, response cancellation, session state, usage accounting, and the OpenAI Realtime event lifecycle.
- `local`, `raw-websocket`, and `socket` use older direct queue orchestration in `build_pipeline`.

Transport selection therefore changes pipeline capabilities. Features added to Realtime can silently be absent elsewhere, and every behavioral change has to account for multiple lifecycles.

A concrete example surfaced while testing the direct-audio LLM path added in #298:

1. `--mode local --stt none` emits turn IDs and soft-end logs.
2. VAD may finalize a segment after the default 64 ms silence window.
3. Local mode has no `SpeculativeTurnTracker`, so the turn remains `rev=0`, listening is cleared, and resumed speech cannot revise or cancel the in-flight response.
4. Realtime wires the tracker and cancellation machinery correctly.

"Local" currently sounds like a deployment choice but also means a reduced feature set.

## Target architecture

```text
                           ┌─ WebSocket client
                           ├─ WebRTC client
canonical Realtime engine ─┤
                           └─ embedded local audio client
```

There is one pipeline construction path, one conversation state machine, one turn/revision implementation, one cancellation path, and one response lifecycle.

### Realtime server

`realtime` remains the canonical server mode and public protocol.

### Local microphone and speakers

Keep the easy local user experience, but make it a client of the Realtime engine rather than a separate engine.

The existing `--mode local` CLI can remain for convenience, but internally it should:

1. start the normal Realtime server/pipeline on loopback;
2. connect a packaged local microphone/speaker client;
3. send microphone audio through `input_audio_buffer.append`;
4. consume Realtime audio and lifecycle events for playback;
5. shut down both sides cleanly from one command.

Factor the useful implementation from `scripts/listen_and_play_realtime.py` into package code and keep the script as a thin entry point if it remains useful.

A real loopback WebSocket is preferred initially because it guarantees that local use exercises the public protocol. Its overhead is negligible relative to model inference. An in-process transport should only be introduced if measurement demonstrates a need.

### Remove legacy transports

Delete `raw-websocket` and TCP `socket` as core modes in this change.

That includes their mode enum values, transport-specific pipeline branches, argument classes, handlers, scripts/examples that only support those modes, tests, and documentation. Update Docker and deployment examples to use Realtime.

If a raw protocol is needed again later for a concrete downstream requirement, it can be designed as a separate Realtime client/adapter at that time. It should not keep the legacy pipeline alive now.

## Implementation scope

- Make the Realtime pipeline unit/service the only pipeline builder.
- Package the local Realtime microphone/speaker client.
- Reimplement `--mode local` on top of the Realtime engine.
- Remove the non-Realtime direct queue orchestration from `build_pipeline`.
- Remove `raw-websocket` and `socket` mode support and their dead code.
- Remove mode-specific `should_listen`, revision, cancellation, and `RuntimeConfig` special cases that become unnecessary.
- Update Docker Compose and all documentation to use Realtime.
- Update CLI help and validation to expose only the final supported modes.
- Consolidate tests around engine behavior plus small transport/client contract tests.

## Tests

Add shared behavior tests that drive the same event/VAD sequence through external and embedded clients and assert identical:

- turn IDs and revisions;
- reopening a soft-ended turn;
- stale revision suppression;
- response cancellation and barge-in;
- committed assistant output;
- audio output lifecycle;
- session instructions and configuration;
- tool calls and usage events where applicable;
- startup and graceful shutdown.

Include an end-to-end local regression test for the #298 scenario: a soft-ended direct-audio turn resumes while generation is in flight, becomes `rev=1+`, and suppresses the stale response.

## Non-goals

- Preserving `raw-websocket` or `socket` compatibility.
- Maintaining two pipeline engines during a transition.
- Changing whether models run locally or through providers.
- Removing the simple one-command local microphone/speaker experience.
- Introducing a new adapter before a concrete consumer requires it.

## Acceptance criteria

- [ ] Realtime is the only pipeline engine and lifecycle implementation.
- [ ] `--mode local` uses the same `RealtimeService` and pipeline unit as WebSocket/WebRTC.
- [ ] A soft-ended local turn can reopen to `rev=1+` and cancel or suppress stale generation.
- [ ] Local, WebSocket, and WebRTC pass shared parity tests for revisions and response lifecycle.
- [ ] `raw-websocket` and `socket` are removed from the CLI and codebase.
- [ ] The legacy direct queue pipeline builder is deleted.
- [ ] Docker, scripts, and documentation use Realtime.
- [ ] No compatibility bridge, deprecation-only scaffolding, or duplicate engine remains.
- [ ] The full test suite and install smoke tests pass.

## Expected simplification

The cutover should leave the project with:

- one pipeline builder;
- one state machine for turns and responses;
- one cancellation implementation;
- one session/configuration model;
- one feature behavior across clients;
- engine tests plus thin transport/client tests instead of a feature-by-mode matrix.

The issue is not complete merely when local audio can connect to Realtime. It is complete when the duplicate orchestration and legacy modes have been deleted.

andimarafioti:
## Summary

- make `RealtimeService` and pooled `PipelineUnit`s the only pipeline engine
- implement `--mode local` as a packaged microphone/speaker client over the loopback Realtime WebSocket
- remove the legacy raw-WebSocket/TCP modes, direct queue orchestration, transport arguments, handlers, and examples
- route transcription and direct-audio input through the single Realtime session, revision, and cancellation lifecycle
- update Docker, CLI help, documentation, and behavior/client contract tests

Closes #421

## Testing

- `pytest -q` — 791 passed
- `ruff check .`
- `mypy src`
- `python tests/install_smoke.py`
- `docker compose config --quiet`


andimarafioti (commented):
Overall, this is the right architectural cutover. The PR makes `RealtimeService`/pooled `PipelineUnit` instances the only engine, `local` uses a real loopback Realtime client instead of rebuilding the pipeline, and the legacy TCP/raw-WebSocket paths are actually removed. I did not find a blocking engine defect, and the revision/direct-audio lifecycle coverage plus the green Linux/macOS checks give me confidence in the refactor.

Before merging, I would like the CLI to expose that architecture directly instead of publishing the intermediate `--mode` surface:

```bash
# Canonical server
speech-to-speech serve ...

# Standalone microphone/speaker client
speech-to-speech talk --url ws://127.0.0.1:8765/v1/realtime

# Convenience composition of the same two components
speech-to-speech local ...
```

Expected shape:

- Remove `--mode` rather than adding a compatibility/deprecation layer.
- `serve` is the only command that constructs the pipeline pool and Realtime server.
- `talk` is a lightweight client command with one canonical full `--url`; avoid exposing overlapping `host`/`port`/`base-url`/`websocket-base-url` connection options. Normalize the full `/v1/realtime` endpoint internally for the OpenAI SDK.
- `local` calls the exact same server builder and client runner in-process, forces loopback, and does not become a second engine or a subprocess wrapper.
- Replace the standalone `scripts/listen_and_play_realtime.py` parser with `talk` so there is one client implementation and one CLI surface. Since the module is no longer local-only, consider renaming `local_client.py` to `audio_client.py`.
- Rename `--local_mac_optimal_settings` to something like `--mac-optimal-settings`; it should select model/device defaults without also selecting a command.
- Use normal Ctrl-C/signal shutdown for `talk` rather than a blocking `input()` worker, which can remain alive after a disconnect.
- Add CLI contract tests for command-specific flags, full-URL normalization, `local == serve + talk` composition, and client shutdown.

I would also default `serve` to `127.0.0.1`, requiring an explicit `--host 0.0.0.0` for network exposure. The Realtime API is unauthenticated and may be backed by paid upstream credentials, so loopback is the safer default.

This command family also leaves a clean future `speech-to-speech demo ...` composition: the same canonical server plus the packaged browser client, not another pipeline path.

PR #434, opened 2026-08-06
Add reversible pre-confirmation audio ducking

### LINKED ISSUE #433: Reduce barge-in latency with reversible pre-confirmation audio ducking
## Problem

Realtime sessions can already receive microphone audio while the assistant is speaking, but audible interruption waits for the confirmed `SpeechStartedEvent`. With the default `min_speech_ms=384`, the demo may continue playing the assistant for roughly four tenths of a second after the user begins a barge-in.

Lowering `min_speech_ms` globally is not a safe fix because short noise bursts would become real conversation turns and cancel model work.

This is a focused follow-up to #134. It should remain compatible with the speculative turn lifecycle in #308.

## Proposed design

Introduce a reversible pre-confirmation phase:

1. VAD emits an internal speech-candidate event after a short configurable amount of active speech.
2. WebSocket and WebRTC demo clients reduce output gain immediately, but do not cancel the response or mutate conversation state.
3. If VAD confirms the normal `speech_started` turn, the existing path performs the hard cancel and flush.
4. If the candidate ends below `min_speech_ms`, the client restores output gain and the response continues.

The candidate events are an optional protocol extension. Clients that do not understand them continue to use the existing confirmed interruption behavior.

## Acceptance criteria

- Candidate detection never creates a conversation item, increments usage, or cancels LLM/TTS work.
- A false candidate restores normal output without losing buffered audio.
- A confirmed candidate still uses the existing cancellation and generation-discard path.
- WebSocket and WebRTC demo clients apply the same gain behavior.
- Candidate state resets on session reset and disconnect.
- Unit and integration tests cover candidate start, false-candidate release, and confirmed transition.
- Existing interruption tests remain green.

## Out of scope

- Feeding partial transcripts to the LLM.
- Speculative tool execution.
- Replacing STT, LLM, or TTS models.
- Changing the speculative turn ownership refactor in #308.

Zippland:
## Summary

- emit a configurable VAD speech candidate after 96 ms of active speech without creating a turn, cancelling LLM/TTS work, or flushing audio
- share a candidate-scoped Web Audio gain state machine across the WebSocket and WebRTC demos, restoring playback for rejected noise and hard-muting only after confirmed speech
- preserve playback when `interrupt_response` is disabled and expose the decision on confirmed speech events
- document the optional Realtime extensions and cover them with VAD, service, router, official OpenAI SDK, and browser-client tests

## Compatibility

Clients that ignore `input_audio_buffer.speech_candidate_started` and `input_audio_buffer.speech_candidate_rejected` keep the existing confirmed interruption behavior. The stock OpenAI Python SDK path is covered by an integration test.

## Tests

- `pytest tests/ -x -q`: 803 passed, 1 skipped
- `mypy src/`: passed
- `ruff check src/ tests/`: passed
- `ruff format --check src/ tests/`: passed
- demo JavaScript syntax checks: passed
- installed-package smoke test: passed

Closes #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ The compose file starts a llama.cpp server with Gemma 4, starts the TCP socket s
 
 ## Realtime API
 
-Realtime mode supports the OpenAI Realtime protocol over WebSocket and WebRTC, with live transcription and low-latency turn-taking. WebSocket clients connect at `/v1/realtime`:
+Realtime mode supports the OpenAI Realtime protocol over WebSocket and WebRTC, with live transcription and low-latency turn-taking. The bundled browser clients also use reversible barge-in: after 96 ms of active speech they temporarily duck assistant playback, but model cancellation and audio-buffer clearing still wait for the normal 384 ms speech confirmation. If the candidate was only a short noise, playback fades back in without restarting the response. WebSocket clients connect at `/v1/realtime`:
 
 ```python
 from openai import OpenAI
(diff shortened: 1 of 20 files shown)


andimarafioti:
This adds a whole new way to detect speech candidates. We already have all the code for that, so it's very redundant. Because it's not properly designed, I'll close it for now.

PR #435, opened 2026-08-06
Populate response.done with generated output items

andimarafioti:
## Summary

- merge the work from #416 while preserving the contributor's commits and authorship
- accumulate generated assistant messages and function-call items for the active response
- include those items in `response.done`'s `response.output`
- give output items stable, distinct IDs that correlate with the emitted streaming events, while preserving upstream function-call IDs
- finalize assistant and function-call item statuses as completed or incomplete based on the response outcome
- add focused regression coverage for assistant and function-call output, multiple chunks and calls, event-to-item correlation, statuses, and state reset

## Why

Clients that follow the Realtime protocol and inspect `response.done` could not discover generated function calls because `response.output` was unset. Function calls were available only through the separate `response.function_call_arguments.done` event.

This change keeps response-scoped output bookkeeping so the final response contains the generated assistant and function-call items, with identities that clients can correlate to the corresponding streaming events.

## Scope

This PR is intentionally focused on accumulating output items and returning them in `response.done.response.output`, including their IDs, indexes, content, and final item statuses.

Broader response-ordering and terminal-lifecycle edge cases—such as audio arriving first, cancellation races, and function-call-only responses—are tracked separately in #438.

Transcript delta/done semantics, example-client compatibility, and related documentation are tracked in #439.

Session-level output-modality behavior predates this change and is outside this PR's scope. Some tests exercise ordering or cancellation scenarios to protect the output bookkeeping implemented here; they are not intended to claim that the broader lifecycle work is complete.

## User impact

Clients can consume generated assistant messages and function calls from `response.done`, correlate them with the streaming events by item identity, and inspect their final completed or incomplete status without relying solely on incremental function-call events.

## Commit structure

The original #416 commits remain intact in the branch history. Follow-up commits focus the final diff on response output bookkeeping and its regression coverage.

## Validation

- `uv run pytest -q tests/openai_realtime/test_realtime_service.py` — 150 passed
- `uv run pytest -q tests/openai_realtime` — 287 passed
- `uv run ruff check .`
- `uv run ruff format --check` on the changed Python files
- `uv run mypy src/`


andimarafioti (commented):
Thanks for working through the edge cases here. The current PR has grown substantially beyond its original goal of populating generated output items in `response.done.output`, and I think we should reduce its scope before merging.

Please keep this PR focused on:

- accumulating the generated assistant and function-call output items;
- returning them in `response.done.output`;
- giving those items stable identities that correlate with the emitted streaming events;
- setting their completed/incomplete status correctly; and
- focused tests for that behavior.

The broader response ordering, audio-first, cancellation, and tool-only lifecycle work is valuable, but it can be reviewed independently in #438.

The transcript delta/done protocol change, example-client compatibility work, and associated documentation should likewise move to #439.

This is not a request to discard that work—the follow-up issues are intended to preserve it while restoring a clear, reviewable boundary for this PR. The session-level output-modality behavior discussed during review also appears to predate this change, so it should not expand this PR further.

PR #437, opened 2026-08-06
Introduce backend registry for pipeline construction

### LINKED ISSUE #436: Introduce a backend registry for STT, LLM, and TTS construction
## Summary

Introduce an explicit registry for built-in STT, LLM, and TTS backends so backend configuration, validation, shared-resource setup, and per-pipeline handler construction have one source of truth.

This is an internal construction/configuration registry. It is **not** runtime service discovery and does not replace endpoint admission or inference-server scheduling.

## Motivation

Backend selection and construction are currently spread across `ModuleArguments`, `ParsedArguments`, the `HfArgumentParser` class list, `prepare_all_args()`, `rename_args()`, `_build_pipeline_unit()`, `_build_handlers()`, and the `get_*_handler()` conditionals in `s2s_pipeline.py`.

Adding one backend therefore requires:

- extending selector literals and error messages;
- adding another field to `ParsedArguments`;
- registering and instantiating its argument dataclass even when it is inactive;
- threading its configuration through several function signatures;
- deep-copying it for every pipeline;
- mutating prefixed dataclass fields through `__dict__`/`vars()`;
- adding another factory branch.

The OpenAI-compatible STT/TTS work in #369 makes this cost more visible. Remote backends also need a clean distinction between resources shared once per endpoint/process—such as HTTP connection pools and client-side admission—and cheap adapters created once per pipeline.

A registry should make adding or removing a backend local and mechanical while preserving the existing CLI and runtime behavior.

## Proposed shape

Use a small explicit catalog rather than decorator-driven registration:

```python
@dataclass(frozen=True)
class BackendSpec:
    name: str
    kind: Literal["stt", "llm", "tts"]
    config_type: type
    create_handler: Callable
    create_shared: Callable | None = None
    required_extra: str | None = None
    capabilities: BackendCapabilities | None = None
```

For example:

```python
STT_BACKENDS = {
    "parakeet-tdt": BackendSpec(
        name="parakeet-tdt",
        kind="stt",
        config_type=ParakeetTDTSTTHandlerArguments,
        create_handler=create_parakeet_tdt,
    ),
    "openai": BackendSpec(
        name="openai",
        kind="stt",
        config_type=OpenAICompatibleSTTHandlerArguments,
        create_shared=create_openai_stt_resources,
        create_handler=create_openai_stt,
    ),
}
```

Factories should receive a standard context instead of long lists of queues and lifecycle objects:

```python
@dataclass
class HandlerContext:
    stop_event: Event
    queue_in: Queue
    queue_out: Queue
    text_output_queue: Queue
    should_listen: Event
    cancel_scope: CancelScope
    speculative_turns: SpeculativeTurnTracker
    pipeline_index: int
```

The lifecycle is intentionally split:

- `create_shared(config)`: optional resources created once for the selected backend configuration, such as a persistent HTTP client or endpoint admission controller;
- `create_handler(context, config, shared)`: a pipeline-local handler/adapter;
- shared resources have explicit idempotent cleanup owned by the pipeline/server builder.

The exact types and module layout are implementation details. Plain dictionaries and explicit imports are preferred for the first version.

## Goals

- Centralize backend names, configuration types, factories, dependency hints, and useful capabilities.
- Parse and normalize only the selected STT, LLM, and TTS backend configurations.
- Replace long backend-specific builder signatures with standard context objects.
- Support optional process/endpoint-shared resources without global hidden ownership.
- Remove backend selection `if/elif` chains from the central pipeline builder.
- Stop mutating argument dataclass `__dict__` values to create handler setup dictionaries.
- Derive CLI choices and validation from the registry.
- Preserve the current public CLI, JSON configuration behavior, defaults, and lazy backend imports during migration.
- Make local and remote backends use the same construction path.

## Non-goals

- Runtime inference-server discovery or replica routing.
- Replacing the endpoint admission controller proposed in #367/#369.
- Designing the STT/TTS wire protocol from #363.
- Third-party plugin/entry-point discovery in the first implementation.
- Migrating every local model into a server as part of this issue.
- Changing which backends are installed by default.

Python entry-point registration can be considered later once the built-in registry has a stable contract.

## Suggested migration

1. Add `BackendSpec`, context types, and explicit built-in registries with characterization tests.
2. Move current `get_stt_handler()`, `get_llm_handler()`, and `get_tts_handler()` branches into registered factory functions without changing behavior.
3. Add optional shared-resource construction and migrate the OpenAI-compatible STT admission/client resources.
4. Make argument parsing obtain selected configuration types and choices from the registry.
5. Replace prefixed-field mutation with explicit normalized configuration objects.
6. Remove obsolete `ParsedArguments` fields, builder parameters, and conditional factories.

This can land incrementally; each step should leave all existing backends usable.

## Acceptance criteria

- Adding a test backend requires a configuration type, factory, and registry entry, without editing the central pipeline builder.
- Only the selected backend configuration for each stage is carried into pipeline construction.
- Backend factories receive a documented standard context rather than backend-specific arguments threaded through multiple builder layers.
- A shared resource is created once and reused by multiple pipeline-local adapters, with deterministic cleanup.
- Existing `serve`, `local`, and `talk` behavior and supported backend selections remain unchanged.
- Existing CLI flags and JSON configuration remain compatible for the migration period.
- Unsupported backend/dependency errors name the selected backend and required extra clearly.
- No backend setup depends on popping or injecting dataclass fields through `__dict__`/`vars()`.
- Tests cover registry lookup, duplicate names, selected-config parsing, factory construction, shared-resource lifetime, cleanup after partial construction failure, and lazy imports.
- The full test, Ruff, mypy, package, and install-smoke checks pass.

## Relationship to existing work

- #369 / #367 add OpenAI-compatible STT/TTS clients and expose the need for endpoint-shared resources.
- #363 defines the larger shared inference-service/coordinator architecture.
- This issue supplies the gateway-side construction boundary needed to migrate local and remote implementations without continuing to expand `s2s_pipeline.py`.

andimarafioti:
## Summary

- add explicit built-in STT, LLM, and TTS registries with lazy factories, dependency hints, and capabilities
- parse and normalize only the selected backend configuration, deriving selector choices and validation from the registry
- replace backend-specific builder signatures with a standard `HandlerContext`
- remove the central backend-selection conditionals and dataclass `vars()` mutation

## Why

Backend configuration and construction were duplicated across parser registration, `ParsedArguments`, argument preparation, pipeline assembly, and three conditional factory functions. Adding a backend therefore required coordinated edits throughout the central builder.

This change makes each built-in backend an explicit catalog entry while preserving existing selectors, defaults, lazy imports, CLI/JSON compatibility, and serve/local/talk behavior.

## Scope

Process/endpoint-shared resource creation and cleanup are intentionally deferred until the first production backend needs persistent clients or admission state. No built-in backend in this PR needs that lifecycle, so the registry factories remain pipeline-local and `ThreadManager` is unchanged. The updated scope is documented on issue #436.

## Compatibility

- known CLI options for inactive backends remain accepted and are warning-ignored, matching JSON handling of inactive keys
- truly unknown options and invalid backend selectors retain their existing argparse errors
- the global device override only updates selected configs that expose a device field
- MLX Audio Whisper retains the existing shared `--language` behavior required by selected-only parsing
- unsupported optional backends report the selected backend, required installation extra, and original import error
- invalid `--enable_llm_proxy` combinations fail before any model construction
- Facebook MMS `--tts_language` and `--facebook_mms_model_name` options now reach the handler; configured custom models are preserved across language switches and sessions

## Validation

- `uv run pytest -q` (872 passed, 1 skipped)
- `uv run ruff check .`
- `uv run ruff format --check` on the changed Python files
- `uv run mypy src/`
- GitHub Actions: pytest, Ruff, mypy, package, Linux install smoke, and macOS ARM64 install smoke all pass

Closes #436

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,8 +495,8 @@ Language coverage depends on the STT and TTS backends you pick, not on the pipel
 
 Make sure the STT, LLM, and TTS you pair all cover your target language(s). Two usage patterns:
 
-- **Single language**: set `--language` to the target language code. The default is `en`.
-- **Language switching**: set `--language auto`. The STT detects the language of each spoken prompt and forwards it to the LLM. Optionally add `--enable_lang_prompt` to append a "Please reply to my message in ..." instruction. It defaults to `False`; large LLMs usually infer the language from context, but the explicit instruction can help smaller models.
+- **Single language**: with Parakeet TDT or a Whisper-family backend, set `--language` to the target language code. The default is `en` for Whisper-family backends; Parakeet auto-detects when the flag is omitted.
+- **Language switching**: with Parakeet TDT or a Whisper-family backend, set `--language auto`. The STT detects the language of each spoken prompt and forwards it to the LLM. Optionally add `--enable_lang_prompt` to append a "Please reply to my message in ..." instruction. It defaults to `False`; large LLMs usually infer the language from context, but the explicit instruction can help smaller models.
 
 Automatic language detection:
 
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ This installs the package in editable mode and makes the `speech-to-speech` CLI 
 | TTS | [ChatTTS](https://github.com/2noise/ChatTTS) | CUDA / CPU | `chattts` |
 | TTS | [MMS TTS](https://huggingface.co/docs/transformers/model_doc/mms) | CUDA / CPU | built-in |
 
-Select implementations with `--stt`, `--llm_backend`, and `--tts`. The CLI registers options only for the selected backends, so inactive backend options are rejected; JSON configuration may include extra inactive-backend keys, which are ignored. Run `speech-to-speech serve -h` for the defaults, or pass selectors before `-h` to see another combination's backend-specific flags (for example, `speech-to-speech serve --stt mlx-audio-whisper -h`).
+Select implementations with `--stt`, `--llm_backend`, and `--tts`. The CLI constructs configuration only for the selected backends; known options for inactive backends remain accepted for compatibility but are ignored with a warning. JSON configuration may likewise include extra inactive-backend keys, which are ignored. Run `speech-to-speech serve -h` for the defaults, or pass selectors before `-h` to see another combination's backend-specific flags (for example, `speech-to-speech serve --stt mlx-audio-whisper -h`).
 
 ## Commands
 
--- a/README.md
+++ b/README.md
@@ -495,8 +495,8 @@ Language coverage depends on the STT and TTS backends you pick, not on the pipel
 
 Make sure the STT, LLM, and TTS you pair all cover your target language(s). Two usage patterns:
 
-- **Single language**: with Parakeet TDT or a Whisper-family backend, set `--language` to the target language code. The default is `en` for Whisper-family backends; Parakeet auto-detects when the flag is omitted.
-- **Language switching**: with Parakeet TDT or a Whisper-family backend, set `--language auto`. The STT detects the language of each spoken prompt and forwards it to the LLM. Optionally add `--enable_lang_prompt` to append a "Please reply to my message in ..." instruction. It defaults to `False`; large LLMs usually infer the language from context, but the explicit instruction can help smaller models.
+- **Single language**: set `--language` to the target language code. The default is `en`.
+- **Language switching**: set `--language auto`. The STT detects the language of each spoken prompt and forwards it to the LLM. Optionally add `--enable_lang_prompt` to append a "Please reply to my message in ..." instruction. It defaults to `False`; large LLMs usually infer the language from context, but the explicit instruction can help smaller models.
 
 Automatic language detection:
 
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ This installs the package in editable mode and makes the `speech-to-speech` CLI 
 | TTS | [ChatTTS](https://github.com/2noise/ChatTTS) | CUDA / CPU | `chattts` |
 | TTS | [MMS TTS](https://huggingface.co/docs/transformers/model_doc/mms) | CUDA / CPU | built-in |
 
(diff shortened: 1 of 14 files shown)


andimarafioti (commented):
Overall, the core registry direction looks good, but I recommend reducing the scope before merging.

The shared-resource lifecycle should be removed from this PR and deferred until the first production backend needs it. No built-in backend in this branch defines `create_shared` or `cleanup_shared`; those hooks currently exist only for tests. Plumbing `shared_resources.close` through `ThreadManager` makes the generic thread utility responsible for non-thread resources, which is what led to the large lifecycle state machine, partial-start handling, timed-join reaper, and associated tests.

For this PR, I would keep:

- the explicit STT/LLM/TTS catalogs;
- selected-backend configuration and normalization;
- the standard `HandlerContext` and registered factories;
- registry-derived selector choices and capabilities;
- removal of the central backend-selection conditionals and `vars()` mutation.

I would remove or defer:

- `BackendSpec.create_shared` / `cleanup_shared`;
- `SharedBackendResources`;
- shared-resource plumbing through the pipeline builders;
- cleanup callbacks and all changes to `ThreadManager`;
- lifecycle and partial-construction tests that exist specifically to prove the deferred resource ownership model.

When the OpenAI-compatible STT/TTS backend or another concrete consumer needs persistent clients/admission state, that work can define its actual cancellation and lifetime requirements. A pipeline/server runtime that owns both the workers and resources will likely be a clearer boundary than extending `ThreadManager`.

The language alias/documentation changes are also unrelated to the registry migration and are better moved to a follow-up. If they remain here, `--stt faster-whisper --language auto` needs fixing before merge: Faster Whisper expects `language=None` for detection, while the current handler passes the literal `"auto"` and discards `info.language`. Parakeet also initializes its fallback language to the literal `"auto"` when detection fails.

With those scope reductions, this becomes a much smaller and easier-to-review registry migration.

PR #450, opened 2026-08-08
Upgrade MLX stack to tested thread-safe releases

### LINKED ISSUE #386: Consider relaxing the mlx==0.31.1 pin — MLX gained multithreaded support in 0.31.2

andimarafioti:
## Summary

- pin `mlx` and `mlx-metal` to `0.32.0`
- pin `mlx-lm` to `0.31.3` and `mlx-audio` to `0.4.7` as a coordinated compatible stack
- update the macOS `transformers` pin to `5.14.1`, as required by `mlx-audio 0.4.7`
- declare `librosa==0.11.0` directly because repository handlers import it and `mlx-audio 0.4.7` no longer supplies it transitively
- retain the global MLX lock and document why core multithread support is not sufficient to remove it from this pipeline
- update the macOS installed-package smoke assertions

## Why

`mlx 0.31.1` crashes under concurrent GPU work. `0.31.2` fixes that race but deterministically segfaults when a Qwen3-TTS worker thread exits. Core `0.32.0` fixes the cleanup crash, but the previous `mlx-lm 0.31.1` pin does not support the repository setup-thread/handler-thread lifecycle and raises `There is no Stream(gpu, 0) in current thread`.

`mlx-lm 0.31.3` adds the required thread-local generation stream. `mlx-audio 0.4.7` contains the subsequent worker-thread, array handoff, and Qwen3-TTS cache fixes; it requires `transformers>=5.14.0`, so this pins the currently tested `5.14.1`. The MLX packages remain exact pins to keep untested releases out of the resolver.

### Why the MLX lock remains

MLX 0.32.0 documents support for independent computations from multiple threads, and two-way concurrent probes were stable. That runtime guarantee did not make this application safe under its real three-component topology, however. Unrestricted three-way STT/LLM/TTS testing intermittently caused:

- a Metal MMU GPU restart with Lightning Whisper + MLX-LM + Qwen3-TTS
- a Parakeet decoder `IndexError` with Parakeet + MLX-LM + Qwen3-TTS

The global lock and the existing multi-pipeline live-transcription safeguard therefore remain. `mlx_lock.py` records the exact tested stack and these failure modes so the lock is not removed based only on the MLX core thread-safety claim. Removing or relaxing it should be a separate change backed by stable end-to-end three-way testing across supported hardware.

Closes #386.

## Validation

- reproduced `mlx 0.31.1` concurrent crashes in 10/10 runs
- reproduced `mlx 0.31.2` Qwen3-TTS worker-exit segfaults in 6/6 runs
- verified `mlx 0.32.0` two-way concurrent probes in 10/10 runs
- exercised the repository Qwen3-TTS, MLX-LM, Whisper, Parakeet, and Kokoro handlers through worker-thread inference
- reproduced both unsafe unrestricted three-way failure modes described above
- installed-package smoke test and dependency compatibility check passed
- `ruff` and `mypy` passed
- full test suite: `908 passed`


binlecode:
Tested PR #450 stack (mlx 0.32.0, mlx-lm 0.31.3, mlx-audio 0.4.7, mlx-metal 0.32.0, transformers 5.14.1) on a MacBook with M3 Max, 36GB, running speech-to-speech 0.2.11. Full pipeline: whisper-mlx large-v3 + Qwen3.6-35B-A3B-4bit + Qwen3-TTS 1.7B CustomVoice.

**Probes: all green.**
- `mlx_race_stress.py` 20 runs × 30s two-way (STT+TTS) concurrent: 0/20 crashes, all reached `joined`.
- `mlx_thread_exit_probe.py` all 4 modes (stt/tts/tts-warm/both): exit 0, no segfault, no `Stream(gpu, N)` error.

**End-to-end: no crashes, but measurable latency regression.**

| Metric | Old (mlx 0.31.1) | New (PR #450) | Delta |
|---|---|---|---|
| speech→first-audio | 1.39–1.54s | 1.90–2.35s | **+0.5–0.8s** |
| LLM lock hold | 0.91–1.04s | 1.41–1.47s | **+50%** |
| TTS lock hold (short) | 0.28–1.27s | 0.81–1.56s | **+30%** |

The primary regression is mlx-lm 0.31.3 — LLM inference is ~50% slower than 0.31.1 at the same model and batch size. This dominates the end-to-end latency gap. TTS is also slower, likely from the mlx-audio 0.4.7 worker-thread changes.

The stack is stable under concurrent load, but the latency cost is user-visible. Worth investigating whether the mlx-lm slowdown is inherent to the thread-local stream fix or something addressable.

andimarafioti:
Thanks for the detailed measurements. I ran a controlled old/new comparison on an M3 Pro with 36 GB, using isolated Python 3.12 environments with the exact dependency versions from before and after this PR.

I could not reproduce the reported 50% LLM or 30% TTS regressions:

| Post-warmup benchmark | Old stack | PR #450 stack | Delta |
|---|---:|---:|---:|
| Qwen3 8B 4-bit, same 24-token response | 1.184s median | 1.178s median | -0.5% |
| Qwen3.5 hybrid 4B 4-bit, handler-like worker thread | 0.799s median | 0.807s median | +1.0% |
| Qwen3.5 hybrid time to first token | 275ms median | 273ms median | -0.5% |
| Qwen3-TTS 1.7B CustomVoice 6-bit, first streamed chunk | 163ms median | 158ms median | -3.0% |

The TTS benchmark used the repository 0.32s MLX streaming interval. Seeded full-generation times were also approximately equal, but the two mlx-audio versions generated different output-token counts, so total lock duration is not directly comparable.

I could not run the exact Qwen3.6-35B model because the available machine did not have enough safe free disk for the roughly 20 GB artifact. There is an upstream report of a model-specific Qwen3.5 35B hybrid/MoE decode regression in mlx-lm 0.31.3, but its measured size is about 7-8%, with TTFT up roughly 2%, rather than 50%: https://github.com/ml-explore/mlx-lm/issues/1425

A rollback to mlx-lm 0.31.1 would not be safe. Later changes include the thread-local generation stream required by the handler-thread lifecycle, plus Qwen3.5 cache correctness and memory fixes:

- https://github.com/ml-explore/mlx-lm/pull/1090
- https://github.com/ml-explore/mlx-lm/pull/1024
- https://github.com/ml-explore/mlx-lm/pull/1077

One measurement concern: LLM lock hold covers the complete generated response, and TTS lock hold covers complete sentence synthesis. Both vary with output-token count and sentence length; neither isolates time to first output. The component measurements above suggest the large end-to-end difference is not a general cost of the thread-local stream or mlx-audio worker-thread changes.

Could you share the exact Qwen3.6 model repository/revision, prompts, generated token counts, sampling arguments and seed, and side-by-side raw timing logs? With those, we can distinguish a model-specific decode regression from different response lengths or pipeline scheduling.

binlecode:
Thanks for the info, my latency numbers don't hold. Retracting that part of my earlier comment. The stability results are actually unaffected.

I compared two different sessions, not two stacks. My "old" column came from a log from 16 days earlier, and that session differed in two ways I didn't check:

- it ran the 8-bit quant of the model, not the 4-bit one
- its prompts were about a third the size (`input_tokens` p50 506 vs 1563)

The extra ~1050 input tokens are things I added to my own app in between (longer system prompt, plus replaying recent conversation turns after a reconnect). At the prefill rate I measure, that alone is roughly +0.5s, which is basically the entire "regression". And your point about lock hold is the other half of it: it covers the whole generation, so it tracks output length, and the two sessions don't even overlap there (118-709 vs 10-137 tokens).

So I redid it properly. Same machine (M3 Max, 36 GB), two isolated venvs, identical weights (`mlx-community/Qwen3.6-35B-A3B-4bit`, revision `38740b847e4cb78f352aba30aa41c76e08e6eb46`), identical prompt, `temp=0` greedy, `mx.random.seed(0)`, fixed `max_tokens`, one warmup then 5 reps, median:

| prompt / gen                                 | metric  |       0.31.1 |       0.31.3 | delta |
| -------------------------------------------- | ------- | -----------: | -----------: | ----: |
| 64 in / 128 out                              | TTFT    |       209 ms |       206 ms | -1.4% |
|                                              | decode  | 118.24 tok/s | 114.82 tok/s | -2.9% |
| 606 in / 128 out                             | TTFT    |       516 ms |       505 ms | -2.1% |
|                                              | prefill |   1521 tok/s |   1550 tok/s | +1.9% |
|                                              | decode  | 114.53 tok/s | 113.93 tok/s | -0.5% |
| 1636 in / 19 out (thinking off, runs to EOS) | TTFT    |      1006 ms |      1060 ms | +5.4% |
|                                              | total   |      1.172 s |      1.227 s | +4.7% |

Two of the three tiers produced byte-identical output on both stacks (128 tokens and 19 tokens respectively), so those rows are strict same-load comparisons. The +5% on the last tier is inside run-to-run noise: re-running the first tier on the same stack moved TTFT by 7% (209 -> 195 ms). The middle tier diverged in text partway through generation at equal token count, which I'd guess is mlx-lm#1024 changing the numerical path.

Short version: no decode or TTFT regression I can measure for this model on mlx-lm 0.31.3. I don't reproduce mlx-lm#1425 here either, though that's a different model family so it doesn't say much either way.

TTS is the same story. Our logs already emit an RTF per synthesis (audio seconds per wall-clock second), which normalizes for sentence length: 0.31.1 sessions sit at p50 4.27 and the PR #450 session at 4.69, so if anything it got ~10% faster. The "+30%" was sentence length, as you suspected.

Agreed on not rolling back, and thanks for the links. #1090 is the one that matters for us, since our handlers run inference off the main thread. That lines up with what the stress runs showed on the pin question: 0.31.1 crashes 20/20 on the first concurrent STT+TTS iteration, 0.31.2 crashes 20/20 at thread exit, and 0.32.0 is 0/20 clean. 0.32.0 is the floor.

So from my side PR #450 looks good: stable under concurrent load, and no latency cost I can measure. Happy to drop in the bench script and the raw per-rep timings if that's useful.

PR #451, opened 2026-08-08
fix(stt): stop Whisper handlers finalizing turns on progressive audio

### LINKED ISSUE #412: Whisper-family STT handlers ignore `mode`, causing premature turn finalization mid-utterance in `--mode realtime`
## Summary
In `--mode realtime`, speech is truncated to only the first ~384ms fragment of an utterance. Everything after that is dropped as `input-after-final`, even though the user is still speaking and VAD has not detected the end of speech yet.

## Root cause
`STT/base_stt_handler.py`'s `before_emit_output` locks a turn/revision as completed the moment ANY `Transcription` output is seen, without checking whether the input chunk that produced it was `mode="progressive"` or `mode="final"`:
```python
def before_emit_output(self, output: STTOut) -> None:
    if isinstance(output, Transcription):
        self._mark_completed_final_revision(output)
```
`parakeet_tdt_handler.py` and `paraformer_handler.py` correctly yield `PartialTranscription` (not `Transcription`) while `mode=="progressive"`, via `smart_progressive_streaming.py`. The three Whisper-family handlers never check `vad_audio.mode` at all and always yield `Transcription`:
- `STT/whisper_stt_handler.py:141`
- `STT/faster_whisper_handler.py:52`
- `STT/lightning_whisper_mlx_handler.py:102`

So the very first progressive chunk of any utterance permanently finalizes the turn for these handlers. Since Parakeet TDT only covers ~25 European languages, any non-Parakeet language (e.g. Japanese) is forced onto a Whisper handler and hits this on every turn, in realtime mode, regardless of VAD tuning.

Confirmed (main = `5b443c8`):
```
STT/whisper_stt_handler.py:141           yield Transcription(
STT/faster_whisper_handler.py:52         yield Transcription(
STT/lightning_whisper_mlx_handler.py:102 yield Transcription(
STT/base_stt_handler.py:73               def before_emit_output(...)
STT/parakeet_tdt_handler.py:244          is_progressive = vad_audio.mode == "progressive"
```
Parakeet's handler branches on `is_progressive` and yields the interim result as `PartialTranscription`, which is exactly what the Whisper-family handlers skip.

## Reproduction
`--mode realtime --stt whisper-mlx --stt_model_name small --language ja`, speak a full sentence.
```
Speech started (confirmed, active=384ms, min=384ms, segment=384ms, turn=turn_1 rev=0)
Transcription completed (language=ja): 今日は
dropping stale STT input-after-final for turn=turn_1 rev=0 age=0.42s
Speech soft-ended (segment=3092ms, active=1856ms, turn=turn_1 rev=0)
```
Full utterance: 今日はとてもいい天気ですね → transcribed as just 今日は.

## Fix (validated locally)
Mirror Parakeet's pattern in the Whisper handlers' `process()`:
```python
if getattr(vad_audio, "mode", None) == "progressive":
    yield PartialTranscription(text=pred_text, turn_id=vad_audio.turn_id, turn_revision=vad_audio.turn_revision)
    return
```
Applied to `lightning_whisper_mlx_handler.py` and confirmed: full-sentence transcription now succeeds, `dropping stale` log disappears (was on every turn, now zero). Same fix still needed in `whisper_stt_handler.py` and `faster_whisper_handler.py` for full coverage.

Trade-off: STT now runs twice per utterance (progressive preview + final).
Measured end-of-speech→first-audio after the fix: 1.7–2.7s on an M2 Mac mini
(whisper-mlx small + Kokoro, remote LLM). The pre-fix figure is not a fair
baseline, since it was produced from a truncated transcript.

Mayankpratapsingh022:
Fixes #412

## Summary

- branch on `vad_audio.mode` in the four Whisper-family STT handlers so progressive
  audio yields `PartialTranscription` rather than `Transcription`
- add `tests/test_whisper_progressive_transcription.py`, covering all four backends

The diagnosis and the shape of the fix are from @whitepepper355 in #412, including the
follow-up comment identifying `mlx_audio_whisper_handler.py` as a fourth affected file.

## Problem

In `--mode realtime` the VAD emits a progressive chunk while the user is still speaking,
then a final chunk once speech ends. `BaseSTTHandler` closes a turn on the first
`Transcription` it observes:

```python
def before_emit_output(self, output: STTOut) -> None:
    if isinstance(output, Transcription):
        self._mark_completed_final_revision(output)
```

`should_process_input` subsequently rejects every later chunk carrying that
`(turn_id, turn_revision)` as `input-after-final`.

`ParakeetTDTSTTHandler` avoids this by inspecting the mode at
`parakeet_tdt_handler.py:244`. The four Whisper-family handlers never read
`vad_audio.mode` and yielded `Transcription` unconditionally, so the progressive chunk
closed the turn and the remainder of the utterance was discarded:

- `STT/whisper_stt_handler.py`
- `STT/faster_whisper_handler.py`
- `STT/lightning_whisper_mlx_handler.py`
- `STT/mlx_audio_whisper_handler.py`

Parakeet is the default STT and covers roughly 25 languages. The defect is therefore
reached only after switching to a Whisper backend, which any language outside that set
requires.

## Why the mode check alone is sufficient

`ParakeetTDTSTTHandler` gates its progressive path behind `enable_live_transcription`.
An equivalent flag is not required in these handlers. `s2s_pipeline.py:500` sets
`vad_kw.enable_realtime_transcription` from `--enable_live_transcription`, and
`vad_handler.py:623` gates the progressive yield on that value. Progressive chunks
therefore exist only when live transcription is enabled, and a handler that receives one
is already operating in that mode.

## Verification

Each backend was exercised with real model weights, driven by the real `VADHandler`
rather than synthetic chunk boundaries. The VAD emitted `['progressive', 'final']` in
every case.

Reference transcription of the complete utterance:

```
I think the weather today is really quite nice and I would like to go outside.
```

| Backend | Model | Delivered to the LLM before | After |
| --- | --- | --- | --- |
| `whisper` | `openai/whisper-small` | `I think` | matches reference |
| `faster-whisper` | `small.en`, int8 | `I think.` | matches reference |
| `whisper-mlx` | `small` | `I think` | matches reference |
| `mlx-audio-whisper` | `whisper-small-mlx` | `I think` | matches reference |

Before, on each backend:

```
[progressive] -> LLM Transcription 'I think'
dropping stale STT input-after-final for turn=turn_1 rev=0
[final] REJECTED by should_process_input
```

After:

```
[progressive]  draft PartialTranscription 'I think'
[final] -> LLM Transcription 'I think the weather today is really quite nice and I would like to go outside.'
```

The `dropping stale` line falls from one occurrence per turn to zero, which is the
signal #412 predicted.

On this branch: 923 passed, 1 skipped. `ruff check`, `ruff format --check` and
`mypy src/` report no issues.

## Test design

The regression test drives the `BaseSTTHandler` hooks in `run()` order and leaves the
input queue empty between chunks. Enqueuing the whole utterance up front instead trips
the separate `progressive-before-final` guard in `should_process_input`, which discards
the progressive chunks and allows the test to pass against unfixed code. Reverting
`src/` while retaining the test produces 8 failures and 8 passes.

The optional backends are stubbed at import time so all four are covered on Linux CI,
where `faster-whisper` and `lightning-whisper-mlx` are not installed.

## Out of scope

Exercising the transformers handler against real weights also surfaces an `IndexError`
at `whisper_stt_handler.py:120`, where `pred_ids[0, 1]` reads a text token rather than a
language tag under transformers 5.6.2. That behaviour is #375 and is already addressed
by #378, so it is left untouched here.

This change does not attempt to quantify the added latency of running STT twice per
utterance, which #412 raises. Measuring it requires the full LLM and TTS chain.

### DIFF
--- a/src/speech_to_speech/STT/faster_whisper_handler.py
+++ b/src/speech_to_speech/STT/faster_whisper_handler.py
@@ -8,7 +8,7 @@ from faster_whisper import WhisperModel
 from rich.console import Console
 
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
-from speech_to_speech.pipeline.messages import Transcription
+from speech_to_speech.pipeline.messages import PartialTranscription, Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
 
 console = Console()
@@ -48,6 +48,14 @@ class FasterWhisperSTTHandler(BaseSTTHandler):
         logger.debug("finished whisper inference")
         if pred_text:
             console.print(f"[yellow]USER: {pred_text}")
+
+            if vad_audio.mode == "progressive":
+                yield PartialTranscription(
+                    text=pred_text,
+                    turn_id=vad_audio.turn_id,
+                    turn_revision=vad_audio.turn_revision,
+                )
+                return
 
             yield Transcription(
                 text=pred_text,
--- a/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
+++ b/src/speech_to_speech/STT/lightning_whisper_mlx_handler.py
@@ -9,7 +9,7 @@ from lightning_whisper_mlx import LightningWhisperMLX
 from rich.console import Console
 
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
-from speech_to_speech.pipeline.messages import Transcription
+from speech_to_speech.pipeline.messages import PartialTranscription, Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
 from speech_to_speech.utils.mlx_lock import MLXLockContext
 
@@ -98,6 +98,14 @@ class LightningWhisperSTTHandler(BaseSTTHandler):
 
         if self.start_language == "auto":
             language_code += "-auto"
+
+        if vad_audio.mode == "progressive":
+            yield PartialTranscription(
+                text=pred_text,
+                turn_id=vad_audio.turn_id,
+                turn_revision=vad_audio.turn_revision,
+            )
+            return
 
         yield Transcription(
             text=pred_text,
--- a/src/speech_to_speech/STT/mlx_audio_whisper_handler.py
+++ b/src/speech_to_speech/STT/mlx_audio_whisper_handler.py
@@ -7,7 +7,7 @@ import numpy as np
 from rich.console import Console
 
 from speech_to_speech.pipeline.handler_types import STTIn, STTOut
-from speech_to_speech.pipeline.messages import Transcription
+from speech_to_speech.pipeline.messages import PartialTranscription, Transcription
 from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
 from speech_to_speech.utils.mlx_lock import MLXLockContext
 
@@ -164,6 +164,14 @@ class MLXAudioWhisperSTTHandler(BaseSTTHandler):
 
         if self.start_language == "auto":
             language_code += "-auto"
+
+        if vad_audio.mode == "progressive":
+            yield PartialTranscription(
+                text=pred_text,
+                turn_id=vad_audio.turn_id,
+                turn_revision=vad_audio.turn_revision,
+            )
+            return
 
         yield Transcription(
             text=pred_text,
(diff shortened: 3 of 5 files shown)


andimarafioti (commented):
Thank you! This is an important fix. I haven't used whisper in a while, which is why I missed it. Codex found a few issues, but I want to merge this quickly, so I'll merge this to a branch before targetting main. I'll keep your commits in the history!

andimarafioti (approved):

PR #452, opened 2026-08-08
Add private conversation console mode

Hacastro:
## What changed

- add `--no_show_conversation_text` for private server sessions
- route STT and TTS conversation displays through one suppressible Rich console
- preserve the current visible-output default and normal lifecycle/error logging
- document the switch and cover CLI/default/local Parakeet+Kokoro wiring

## Why

The server currently prints user and assistant text even at `--log_level error`. Persistent local services can therefore retain private conversation content in service-manager logs. This provides an explicit opt-in privacy mode without changing existing defaults.

## Verification

- Ruff check and format: pass
- mypy: 92 source files pass
- targeted tests: 47 passed
- broad suite excluding one unrelated pre-existing Windows timing-boundary test: 875 passed, 18 skipped

The excluded test is `test_vad_complete_smart_turn_selects_shorter_speculative_grace`; on Windows its upper-bound assertion observes exactly 0.800 seconds after rounding. This change does not touch VAD or speculative-turn behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,6 +574,7 @@ See [ModuleArguments](./src/speech_to_speech/arguments_classes/module_arguments.
 - LLM backend (`--llm_backend`: `transformers`, `mlx-lm`, `responses-api`, or `chat-completions`)
 - TTS implementation (`--tts`)
 - logging level
+- optional conversation text suppression (`--no_show_conversation_text`) for private sessions
 - realtime pipeline pool size (`--num_pipelines`)
 
 ### VAD Parameters
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -501,7 +501,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 if self._generation_is_stale(turn.gen):
                     logger.info("LLM generation cancelled (interruption)")
                 else:
-                    logger.debug(f"Clean text: {state.clean_text}")
+                    logger.debug("Clean text generated (characters=%d)", len(state.clean_text))
                     yield from _flush(sentence_batch)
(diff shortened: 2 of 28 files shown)


Hacastro:
Follow-up at 68d5683 closes the two confirmed output paths: private mode now propagates into the packaged local audio client, suppresses transcript/JSON/tool-argument/raw-error content while retaining content-free lifecycle output, and serve-mode transcript/response logs now record only metadata. Added sentinel regression coverage for local stdout plus serve-mode STT and LLM logs. Verification: Ruff and format clean, mypy clean across 92 source files, 86 focused tests pass; broad suite 876 passed/18 skipped/1 pre-existing Windows timing test deselected, and the two incidental Windows timing failures passed on immediate isolated rerun. No runtime or host persistence change.

Hacastro:
Follow-up b01d0a4 closes the MM4 review residual: all three INFO logs that serialized argument-bearing tool-call objects now emit only Tools generated (count=N). The streaming chat-completions regression drives a private argument sentinel through a real tool call, confirms the functional tool arguments remain intact, and asserts the sentinel is absent from captured INFO logs. Current focused gate: Ruff/format clean, mypy 92 source files, 113 tests passed. No runtime or host change.

Hacastro:
Independent read-only MM4 review of b01d0a4: APPROVE. The real chat-completions handler reassembled a neutral sentinel across two streamed tool-argument deltas; the functional tool call remained intact, the sentinel was absent from captured INFO logs, and the only operational line was Tools generated (count=1). No runtime, port, persistence, or robot action was used for this final check.

andimarafioti:
This is important, so I created this issue to track it: https://github.com/huggingface/speech-to-speech/issues/475
I think the design you're introducing, although technically correct, carries an amount of technical debt. The logging in the repo needs to be rethought a bit, and the change should be more significant than what you're introducing here. I'll work on it.

PR #466, opened 2026-08-11
Preserve queued provider token usage on session disconnect

### LINKED ISSUE #456: Preserve queued provider token usage on session disconnect
## Problem

A client disconnect can discard provider-reported token usage that has reached a router output queue but has not yet been dispatched into `RealtimeService` accounting.

On `main`, `LMOutputProcessor` routes `TokenUsageEvent` through `text_output_queue`. Session release calls `_clean_unit(unit)`, which drains that queue before `RealtimeService.unregister()` rolls connection usage into global totals. A queued usage event is therefore removed without ever reaching either `response_usage` or `pending_token_usage`.

This is pre-existing behavior on `main`; it was discovered while reviewing #453 and is not introduced by that PR.

## Reproduction

1. Register a realtime session and start or queue a keyed response.
2. Put a `TokenUsageEvent` with nonzero input/output tokens on the router queue, without allowing the send loop to consume it.
3. Release the session, which calls `_clean_unit`.
4. Let `SESSION_END` drain and unregister the connection.
5. Observe that global token totals remain unchanged.

## Expected behavior

Provider-reported usage that has already entered the pipeline should be retained in global accounting even when the client disconnects before the router publishes a response terminal.

## Acceptance criteria

- Disconnect cleanup preserves or directly accounts queued `TokenUsageEvent` values before unregistering the connection.
- Stale assistant text, tools, and audio are still discarded and cannot leak into the next session.
- WebSocket and WebRTC release paths behave consistently.
- A regression test covers queued usage followed immediately by disconnect.

Hotragn:
Fixes #456

## Problem

`LMOutputProcessor` routes `TokenUsageEvent` through the router's output queues. Releasing a session calls `_clean_unit()`, which drains those queues — and by then the send loop is gone, so a usage event that reached a queue but was never dispatched was discarded before `RealtimeService.unregister()` rolled the connection's usage into the global totals.

The provider had already billed those tokens, so a client disconnecting at the wrong moment silently under-reported usage. `_dispatch_pipeline_event()` already treats usage as billable accounting that cancellation must not make stale — the release path was the remaining hole.

## Fix

`_release_session()` now accounts queued usage directly, before the flush and while the session is still registered:

- **`_drain_token_usage()`** removes only `TokenUsageEvent`s from a queue and re-inserts every other item at the front in its original order, under the queue mutex, mirroring `_flush_queue()`. That makes it correct independently of a flush, rather than only valid when called immediately before one.
- **`_account_queued_token_usage()`** feeds each drained event through `service.dispatch_pipeline_event()` — the same entry point the send loop uses, so accounting goes through one path rather than a parallel one. Returned server events are discarded, since the client is gone and only the accounting side effect matters. Both `output_queue` and `text_output_queue` are covered.
- Accounting failures are logged, never raised, so release can't be blocked by them.

## Acceptance criteria

- **Queued usage accounted before unregistering** — done, directly rather than by preserving, since after release nothing remains to consume the queue.
- **Stale text/tools/audio still discarded** — unchanged. Only `TokenUsageEvent` is removed ahead of the flush; `_clean_unit()` itself is untouched. There's an explicit test asserting both queues end up empty and no assistant output survives.
- **WebSocket and WebRTC consistent** — both share `_release_session()`, so they get this identically.
- **Regression test for queued usage then immediate disconnect** — `test_release_session_preserves_queued_usage` drives the real `_release_session()`.

## Tests

`tests/openai_realtime/test_release_token_usage.py` — 12 cases: usage queued on either queue and on both at once, multiple events, the no-usage case, an accounting failure, an unknown session, `_drain_token_usage` ordering/isolation, and stale assistant output still being dropped.

The end-to-end case fails against current `main` with exactly the reported symptom:

```
assert totals(unit.service) == (64, 32)
E   assert (0, 0) == (64, 32)
```

I deliberately routed that test through `_release_session()` rather than the new helper, so it exercises the real disconnect path instead of just the code I added.

Full suite: **1023 passed, 1 skipped**. `ruff check`, `ruff format --check`, `mypy src/` clean.

### DIFF
--- a/src/speech_to_speech/api/openai_realtime/websocket_router.py
+++ b/src/speech_to_speech/api/openai_realtime/websocket_router.py
@@ -181,6 +181,52 @@ def _flush_queue(q: Queue[QItem], *, preserve: Callable[[QItem], bool] | None = 
             q.not_empty.notify(len(preserved))
 
 
+def _drain_token_usage(q: Queue[Any]) -> list[TokenUsageEvent]:
+    """Remove and return queued ``TokenUsageEvent``s, leaving every other item in place.
+
+    Non-matching items are re-inserted at the front in their original order (atomically
+    under the queue's mutex), mirroring :func:`_flush_queue`, so this is safe to call
+    independently of a flush.
+    """
+    usage: list[TokenUsageEvent] = []
+    remaining: list[Any] = []
+    while True:
+        try:
+            item = q.get_nowait()
+        except Empty:
+            break
+        if isinstance(item, TokenUsageEvent):
+            usage.append(item)
+        else:
+            remaining.append(item)
+    if remaining:
+        with q.mutex:
+            for item in reversed(remaining):
+                q.queue.appendleft(item)
+            q.not_empty.notify(len(remaining))
+    return usage
+
+
+def _account_queued_token_usage(unit: PipelineUnit, session_id: str) -> None:
+    """Fold queued provider usage into session accounting before release discards it.
+
+    Releasing a session calls :func:`_clean_unit`, which drains the output queues, and by
+    then the send loop is gone — so a ``TokenUsageEvent`` that reached a queue but was
+    never dispatched would be dropped before ``unregister()`` rolls this session's usage
+    into the global totals. Billable usage the provider already reported must survive a
+    client disconnect, so account it directly here.
+
+    The returned server events are discarded: the client is gone, and only the accounting
+    side effect matters.
+    """
+    for q in (unit.output_queue, unit.text_output_queue):
+        for event in _drain_token_usage(q):
+            try:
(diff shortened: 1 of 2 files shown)


andimarafioti:
Commit [4925dbc](https://github.com/huggingface/speech-to-speech/commit/4925dbc65640470c307fe447d5a4d9b03dc24443) already implements the simpler, safer solution:
Accounting occurs inside the existing queue flush through on_discard=account_usage, eliminating the drain-then-flush gap.
It also accounts usage temporarily held in session output state.
Non-usage assistant text, tools, and audio remain discarded.
WebSocket and WebRTC still share _release_session().
main includes a disconnect regression test covering queued and held usage.

PR #481, opened 2026-08-12
Fix whitespace loss between streamed LLM chunks

### LINKED ISSUE #480: Streaming LLM output loses whitespace between chunks
When using an OpenAI-compatible LLM with streaming enabled, whitespace can be lost between streamed chunks.

Example provider chunks:

"What is "
"the weather today?"

**Expected**:

What is the weather today?

**Actual**:

What isthe weather today?

This appears to happen in BaseOpenAICompatibleHandler._consume_streaming after sent_tokenize() processes printable_text. The remaining sentence can lose its trailing whitespace before the next streamed delta is appended.

**Expected behavior**: preserve provider-supplied whitespace across streaming chunk boundaries.

rishanreddy:
Fixes #480


andimarafioti (approved):
Thank you!

PR #491, opened 2026-08-16
Honor explicit reasoning effort for official OpenAI requests

### LINKED ISSUE #490: Chat Completions drops configured reasoning effort for official OpenAI endpoints
## Summary

The `chat-completions` backend silently drops an explicitly configured `--responses_api_reasoning_effort none` when using the official OpenAI API (either the default/unset base URL or `https://api.openai.com/v1`).

This breaks `gpt-5.6-terra` sessions that expose function tools. Because the request omits the configured effort, the model uses its default `medium` effort and OpenAI rejects the request:

```text
Function tools with reasoning_effort are not supported for gpt-5.6-terra in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.
```

## Reproduction

1. Start the server with:
   - `--llm_backend chat-completions`
   - `--model_name gpt-5.6-terra`
   - official OpenAI API credentials
   - `--responses_api_reasoning_effort none`
2. Create a Realtime session containing function tools.
3. Trigger a response.

The backend warmup succeeds because it does not include tools. The first runtime request containing function tools fails with the 400 above.

This was reproduced on a newly started container after confirming that its endpoint configuration contained `RESPONSES_API_REASONING_EFFORT=none`, ruling out a stale rollout.

## Root cause

`BaseOpenAICompatibleHandler._build_extra_body()` returns before considering `reasoning_effort` whenever `base_url is None` or points to official OpenAI:

https://github.com/huggingface/speech-to-speech/blob/41536471e074723543ba6fc023319a96544f4303/src/speech_to_speech/LLM/base_openai_compatible_language_model.py#L249-L252

The Chat Completions request only forwards the resulting `extra_body`, so no effort is sent:

https://github.com/huggingface/speech-to-speech/blob/41536471e074723543ba6fc023319a96544f4303/src/speech_to_speech/LLM/chat_completions_language_model.py#L160-L180

The existing unit test currently codifies the incorrect behavior by expecting explicit `none` to be discarded for official OpenAI:

https://github.com/huggingface/speech-to-speech/blob/41536471e074723543ba6fc023319a96544f4303/tests/test_chat_completions_backend.py#L200-L208

## Expected behavior

An explicitly configured `reasoning_effort=none` must be included in the official OpenAI Chat Completions request. OpenAI documents `reasoning_effort` as a supported top-level Chat Completions parameter, including the value `none`:

https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create

OpenAI recommends the Responses API for reasoning and tool-calling workflows, but Chat Completions with function tools is supported here when the effective effort is `none`:

https://developers.openai.com/api/docs/guides/latest-model

## Smallest correction

Either:

- prioritize an explicit `reasoning_effort` before the provider-specific/official-OpenAI guard so it is merged into the request body; or
- pass it through the OpenAI SDK's typed top-level `reasoning_effort` parameter for official OpenAI requests.

Keep the existing `chat_template_kwargs.enable_thinking=false` behavior for non-OpenAI-compatible providers unchanged.

## Acceptance criteria

- With an unset/default OpenAI base URL and configured effort `none`, the outbound Chat Completions request includes `reasoning_effort: "none"`.
- The same holds for an explicit `https://api.openai.com/v1` base URL.
- A request containing function tools no longer fails with the reported 400.
- Existing provider-specific reasoning/disable-thinking behavior remains unchanged.

atirna:
## Description
I changed the chat-completions backend so an explicitly configured `responses_api_reasoning_effort` is honored on the official OpenAI API. Previously, `_build_extra_body` returned early for official OpenAI (or an unset base URL) before considering the configured effort, so it was silently dropped and reasoning models exposing function tools fell back to the default effort, which `/v1/chat/completions` rejects; now the effort reaches the request, while `chat_template_kwargs` still never goes to official OpenAI.

Fixes #490

## Why
without the effort the first runtime request with function tools fails with the 400 from the issue, even though warmup succeeds

## Verification
- before, on current `main`: `pytest tests/test_chat_completions_backend.py -q` — the new official-OpenAI cases fail because the effort is dropped
- after: `pytest tests/test_chat_completions_backend.py -q` — 40 passed
- also ran `pytest tests/ -q` — 1125 passed, 2 skipped

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -243,13 +243,17 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
         ``chat_template_kwargs.enable_thinking=false``, while others (e.g. GLM via
         the HF router) ignore that and require ``reasoning_effort='none'``. A
         non-empty ``reasoning_effort`` therefore takes precedence; otherwise we fall
-        back to the chat-template flag. None of this applies to the official
-        OpenAI server, which rejects unknown extra_body keys.
+        back to the chat-template flag. An explicit effort is honored even for the
+        official OpenAI server, where ``reasoning_effort`` is a documented
+        top-level Chat Completions parameter (e.g. ``none`` is required to combine
+        function tools with reasoning models); the provider-specific
+        chat-template flag never reaches it, as the server rejects unknown
+        extra_body keys.
         """
-        if base_url is None or cls._is_official_openai(base_url):
-            return None
         if reasoning_effort:
             return {"reasoning_effort": reasoning_effort}
+        if base_url is None or cls._is_official_openai(base_url):
+            return None
         if disable_thinking:
             return {"chat_template_kwargs": {"enable_thinking": False}}
         return None
--- a/tests/test_chat_completions_backend.py
+++ b/tests/test_chat_completions_backend.py
@@ -13,6 +13,7 @@ import json
 import queue
 import threading
 from types import SimpleNamespace
+from typing import Optional
 
 import httpx
 import numpy as np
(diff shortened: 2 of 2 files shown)


andimarafioti:
Superseeded by #494 
sorry, I didn't cat it when I started working on #490. This was a good PR :(

atirna:
no problem, thanks for review

PR #492, opened 2026-08-16
Fix(stt): Propagate `FasterWhisper language_code` on Transcription

salignatmoandal:
Hi,

Small follow-up in the same spirit as the Facebook MMS language-key fix: FasterWhisper already detects a language via `info.language`, but the handler never put it on `Transcription`, so downstream lang-prompt / multilingual TTS saw `None`.

## Summary
- Set `language_code=info.language` on final `Transcription` outputs from FasterWhisper.
- Add regression tests (mocked `faster_whisper`) for language propagation, empty transcripts, and progressive partials.

## Why / example use case
Other STT backends already emit Whisper-style codes on `Transcription`. With FasterWhisper, a Polish turn could produce text while leaving `language_code=None`, so `--enable_lang_prompt` and multilingual TTS could not follow the detected language.

## Test plan
- [x] `uv run pytest tests/test_faster_whisper_language_code.py -v`
- [ ] Confirm a non-empty FasterWhisper transcript carries `info.language` on `Transcription`
- [ ] Confirm empty / whitespace-only transcripts still yield nothing

### DIFF
--- a/src/speech_to_speech/STT/faster_whisper_handler.py
+++ b/src/speech_to_speech/STT/faster_whisper_handler.py
@@ -59,6 +59,7 @@ class FasterWhisperSTTHandler(BaseSTTHandler):
 
             yield Transcription(
                 text=pred_text,
+                language_code=info.language,
                 turn_id=vad_audio.turn_id,
                 turn_revision=vad_audio.turn_revision,
                 speech_stopped_at_s=vad_audio.created_at_s,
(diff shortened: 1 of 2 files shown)


andimarafioti (approved):
thank you for this!
The only issue I see is with the test that's failing. I'll have codex merge this to a branch and fix the test before merging to main. I'll keep your commit in the history!

PR #493, opened 2026-08-16
Add CONTRIBUTING.md

nslatour:
## Add CONTRIBUTING.md

Repository has no contributing guidelines. Adding one helps new contributors.

## Changes

Modified 1 file(s): CONTRIBUTING.md

## Checklist

- [x] Code follows the project's style guidelines
- [x] Changes have been tested locally
- [x] Documentation updated if needed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,70 @@
+# Contributing to speech-to-speech
+
+Thank you for your interest in contributing to speech-to-speech! This document outlines the process for contributing to the project.
+
+## Getting Started
+
+1. Fork the repository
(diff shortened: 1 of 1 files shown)


andimarafioti:
Thanks—having a dedicated contributing guide could be useful. As written, though, this doesn’t match the repository: there is no Makefile, so make install and make test fail, while the project uses uv, pytest, ruff, and mypy. It also introduces Conventional Commits as project policy without that being established consistently. I'll try to work on this, feel free to create an issue!

nslatour:
Thanks for the detailed feedback — genuinely helpful, and you're right about the Makefile/tooling mismatch. Appreciate you taking the time.

PR #497, opened 2026-08-17
Fix Markdown pronunciation in TTS output

### LINKED ISSUE #338: LLM responses containing Markdown are read literally by TTS (asterisks, headings, bullets, etc.)

andimarafioti:
## Summary

- merge PR #344 onto current `main` without squashing or rewriting its original commits
- strip the Markdown forms from #338 only after streamed text has been reassembled into complete spoken segments
- handle matched emphasis adjacent to CJK or other word characters while preserving compact operator expressions
- buffer streamed fenced code through sentence tokenization, removing its fences without applying prose cleanup inside the code body
- preserve text-only output verbatim and avoid changing compact math, snake_case identifiers, or non-heading `#` text
- keep provider-native and locally parsed tool names and arguments completely outside Markdown cleanup
- keep link rewriting out of this fix because it is not part of #338

Closes #338.

## Test plan

- `uv run pytest tests/ -x -q` — 1,166 passed, 1 skipped
- `uv run ruff check src/ tests/`
- `uv run ruff format --check src/ tests/`
- `uv run mypy src/`

This supersedes #344 while preserving both of its original commits in the merge history.

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -29,7 +29,7 @@ from speech_to_speech.LLM.chat import (
 )
 from speech_to_speech.LLM.compaction_prompt import CompactGenerateFn, build_compactor
 from speech_to_speech.LLM.text_prompt import build_text_system_prompt
-from speech_to_speech.LLM.utils import remove_unspeechable, resolve_auto_language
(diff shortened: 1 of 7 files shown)


adaofeliz:
Thanks @andimarafioti

PR #501, opened 2026-08-17
Fix(stt): Propagate Paraformer language_code on Transcription.

salignatmoandal:
hey again, (: 

Small follow-up in the same spirit as #492 (FasterWhisper): Paraformer was emitting `Transcription` without `language_code`, so downstream `--enable_lang_prompt` and multilingual TTS saw `None`.

Paraformer has no language detection — the default model is `paraformer-zh` — so this PR declares `language_code="zh"` on final transcriptions.

## Changes
- Set `language_code="zh"` on `Transcription` in `paraformer_handler.py`
- Assert `language_code` in existing Paraformer transcription tests

## Test plan
- [x] `uv run pytest tests/test_paraformer_transcription_events.py -v`

### DIFF
--- a/src/speech_to_speech/STT/paraformer_handler.py
+++ b/src/speech_to_speech/STT/paraformer_handler.py
@@ -70,6 +70,7 @@ class ParaformerSTTHandler(BaseSTTHandler):
         else:
             yield Transcription(
                 text=pred_text,
+                language_code="zh",
                 turn_id=vad_audio.turn_id,
                 turn_revision=vad_audio.turn_revision,
                 speech_stopped_at_s=vad_audio.created_at_s,
--- a/src/speech_to_speech/STT/paraformer_handler.py
+++ b/src/speech_to_speech/STT/paraformer_handler.py
@@ -16,20 +16,6 @@ logger = logging.getLogger(__name__)
 console = Console()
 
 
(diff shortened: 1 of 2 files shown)


src/speech_to_speech/STT/paraformer_handler.py
```diff
@@ -70,6 +70,7 @@ def process(self, vad_audio: STTIn) -> Iterator[STTOut]:
         else:
             yield Transcription(
                 text=pred_text,
+                language_code="zh",
```
andimarafioti: This is good, but there is also a paraformer english model https://huggingface.co/funasr/paraformer-en
So I would rather we extract the language from `--paraformer_stt_model_name`

andimarafioti (changes requested):
one small request

src/speech_to_speech/STT/paraformer_handler.py
salignatmoandal: Good catch.  Updated.

language_code is now taken from `--paraformer_stt_model_name` (`paraformer-zh` → `zh`, `funasr/paraformer-en` →`en`) instead of a hardcoded `zh`. 

Tell me if it's good!

andimarafioti (approved):
Thank you!!

Issue #482, opened 2026-08-13
PocketTTSHandler always loads the English model — `language_code` is read but never used


lixyra05-debug:
## Summary

`PocketTTSHandler` calls `TTSModel.load_model()` with no arguments, so it always loads
the English weights. It reads `tts_input.language_code` but only writes it to a debug log,
and there is no `pocket_tts_language` CLI flag, so no combination of options can make the
`pocket` backend speak another language.

The failure is silent: the audio is clean — healthy peaks, no clipping — it is simply an
English speaker reading French phonetically. Nothing in the logs signals a problem.

## Environment

| | |
|---|---|
| speech-to-speech | 0.2.12, commit `3e77fda` |
| pocket-tts | 2.1.0 |
| Platform | macOS 15 / Apple Silicon (M3), `--pocket_tts_device cpu` |
| Backends | `--stt parakeet-tdt --tts pocket` |

## Expected vs actual

**Expected:** with a French conversation, the Pocket backend synthesizes French.

**Actual:** it synthesizes French text with English weights and an English voice. Round-trip
WER (synthesize, then transcribe with Kyutai STT-1B `en_fr`, 8 French sentences chosen for
liaisons, nasals and numbers):

| configuration | round-trip WER (median) |
|---|---|
| current behaviour (English weights + default voice `jean`) | **87.8 %** |
| `load_model(language="french_24l")` + voice `estelle` | **13.6 %** |

Example, same sentence, same TTS, same transcriber:

```
reference : Quand est-ce qu'ils arriveront à l'aéroport de Roissy ?
current   : behind the essay quills arrive on a letter for the Rossi
fixed     : Quand est-ce qu'ils arriveront à l'aéroport de Rois
```

## Reproduce

Minimal, no pipeline needed — this is what the handler does at setup:

```python
from pocket_tts import TTSModel
model = TTSModel.load_model()      # exactly the call in PocketTTSHandler.setup()
print(model.origin.stem)           # -> "english"
```

End to end:

```bash
speech-to-speech serve --stt parakeet-tdt --parakeet_tdt_language fr \
                       --tts pocket --log_level debug
```

Speak French. The debug log shows `Received language code: fr` while the model that was
loaded is the English one.

## Root cause

**1. The model is loaded without a language.**
`src/speech_to_speech/TTS/pocket_tts_handler.py:70-71`

```python
logger.info("Loading Pocket TTS model")
self.model = TTSModel.load_model()
```

[...]

PR #502, opened 2026-08-18
Add configurable language support for PocketTTS

raphaelp:
## Summary

Addresses #482 by adding startup language configuration for the PocketTTS backend.

- Add a `--pocket_tts_language` option.
- Pass the configured language to `TTSModel.load_model()`.
- Preserve `english` as the default.
- Add tests for backend argument normalization and PocketTTS model initialization.

## Testing

- Full test suite passes: `1149 passed, 1 skipped`.
- Verified `--pocket_tts_language french_24l` is exposed through the CLI.
- Verified the speech-to-speech pipeline initializes PocketTTS with the `french_24l` model.
- Manually synthesized French speech using both the default model and `french_24l`; the language-specific model produced better French pronunciation.

## Scope

This change configures the PocketTTS language at model initialization.

Dynamic per-utterance switching based on `tts_input.language_code` is intentionally not included here, since that would require defining model reload/caching behavior.

### DIFF
--- a/src/speech_to_speech/TTS/pocket_tts_handler.py
+++ b/src/speech_to_speech/TTS/pocket_tts_handler.py
@@ -29,6 +29,7 @@ class PocketTTSHandler(BaseHandler[TTSIn, TTSOut]):
         should_listen: Event,
         device: str = "cpu",
         voice: str = "alba",  # Default voice from catalog
+        language: str = "english",
         sample_rate: int = 16000,  # Match the pipeline's native audio output rate.
         blocksize: int = 512,
         max_tokens: int = 50,
@@ -46,6 +47,7 @@ class PocketTTSHandler(BaseHandler[TTSIn, TTSOut]):
                 - A preset name: 'alba', 'marius', 'javert', 'jean', 'fantine', 'cosette', 'eponine', 'azelma'
                 - A local audio file path
                 - A Hugging Face path like "hf://kyutai/tts-voices/..."
+            language: PocketTTS language/model configuration to load
             sample_rate: Output sample rate (pocket-tts generates at 24kHz and will be resampled to this rate). Default 16kHz matches the pipeline's audio output.
             blocksize: Size of audio blocks to yield
             max_tokens: Maximum tokens to generate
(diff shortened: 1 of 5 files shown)


src/speech_to_speech/TTS/pocket_tts_handler.py
```diff
@@ -67,8 +70,8 @@ def setup(
         # Import and load model
         from pocket_tts import TTSModel
 
-        logger.info("Loading Pocket TTS model")
-        self.model = TTSModel.load_model()
+        logger.info(f"Loading Pocket TTS model for language: {self.language}")
+        self.model = TTSModel.load_model(language=self.language)
```
andimarafioti, line 74: `language=` was introduced in PocketTTS 2.0.0, but the optional dependency still permits `pocket-tts>=0.1.0`. With the supported 1.1.1 release this startup path raises `TypeError: unexpected keyword argument 'language'`. Update the dependency floor to `pocket-tts>=2.0.0`

tests/test_pocket_tts_handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.parametrize(
+    ("setup_kwargs", "expected_language"),
+    [
+        ({}, "english"),
+        ({"language": "french_24l"}, "french_24l"),
+    ],
+)
+def test_pocket_tts_setup_loads_language(monkeypatch, setup_kwargs, expected_language):
+    import pocket_tts
```
andimarafioti: CI runs `uv sync --group dev` without the `pocket` extra, so this import raises `ModuleNotFoundError` and both new cases fail. Inject a fake `pocket_tts` module through `sys.modules`, or explicitly install the extra in CI.

andimarafioti (changes requested):
two small comments

src/speech_to_speech/TTS/pocket_tts_handler.py
raphaelp: Good catch, thanks! Updated the PocketTTS dependency floor to >=2.0.0.

tests/test_pocket_tts_handler.py
raphaelp: Thanks! Updated the test to inject a fake pocket_tts module through sys.modules.

andimarafioti (approved):
Awesome work, thank you so much!

Issue #151, opened 2024-12-19
No Acoustic Echo Cancellation


Galaxywalk:
It seems that the current pipeline is almost linear and no acoustic echo cancellation is implemented. It means that when the speaker speaks the responses from LLM, the microphone may record this audio incorrectly?

eustlb:
Hey @Galaxywalk,
The pipeline is implemented such that as soon as an end of speech is detected by the VAD, it stops listening to the microphone until the spoken LLM answer has been fully played. 
Echo cancellation is the one big feature we would love to implement because it would enable interruptions.

Galaxywalk:
I'm currently working on acoustic echo cancellation for a language-instructed robot, which I believe is crucial. Could you share your schedule for this project? Also, packaging the current pipeline into a ROS package would significantly enhance its usability and maintainability for ROS users. Do you have any plans for this?

I've already made some progress in this area and would be eager to contribute. What are your thoughts on external contributions, and would it be possible for an external PR about it?

PR #503, opened 2026-08-18
Stop --audio_enhancement claiming echo cancellation

### LINKED ISSUE #483: `--audio_enhancement` help text mentions echo cancellation, which DeepFilterNet cannot provide
## Summary

The help text for `--audio_enhancement` says it applies "noise reduction, equalization,
and echo cancellation". The implementation is DeepFilterNet, a single-channel speech
enhancer: it has no far-end reference and therefore cannot cancel acoustic echo.

This is a documentation issue rather than a code defect, but it is a costly one to hit.
Acoustic feedback is the first thing you meet when running `speech-to-speech local` on
laptop speakers, and this flag is the only thing in the CLI that appears to address it.
The flag that *does* fix it — `--local_audio_block_mic_during_playback` — is easy to miss
because its own (accurate) help text does not mention echo.

## Environment

speech-to-speech 0.2.12, commit `3e77fda`, macOS 15 / Apple Silicon,
`speech-to-speech local` with built-in mic and speakers, no headset.

## The claim

`src/speech_to_speech/arguments_classes/vad_arguments.py:48-52`

```python
audio_enhancement: bool = field(
    default=False,
    metadata={
        "help": "improves sound quality by applying techniques like noise reduction, equalization, and echo cancellation. Default is False."
    },
)
```

## Why echo cancellation is not achievable at this point in the pipeline

`src/speech_to_speech/VAD/vad_handler.py:45` imports `from df.enhance import enhance,
init_df`, and calls it at line 820:

```python
enhanced = enhance(self.enhanced_model, self.df_state, torch.from_numpy(array))
```

The call takes only the near-end signal. Acoustic echo cancellation needs the **far-end
reference** — the samples that went to the speaker — so an adaptive filter can estimate
the room path and subtract it. No reference is available in the VAD handler, which sits
on the microphone path only.

There is a second reason, specific to this application. DeepFilterNet is trained to
suppress noise and preserve speech. In a voice assistant the echo **is** speech — the
system's own TTS output, clean and well-formed. A speech-preserving denoiser is therefore
the wrong tool for it: at best it leaves the echo intact, at worst it cleans it up and
makes it easier for the STT to transcribe.

Searching the repository for an echo-cancellation implementation
(`echo.?cancel|aec|webrtc.?apm|speex`) returns one hit: the help string quoted above.

## Measured impact of the underlying problem

Ten minutes of French conversation on laptop speakers, `--tts pocket`. Counting user turns
whose transcript shares a word trigram with the immediately preceding assistant reply:

| configuration | echoed turns |
|---|---|
| defaults | **10 / 58** |
| `--local_audio_block_mic_during_playback` | **0 / 43** |

The runaway is characteristic — five consecutive turns re-entering on the same opening:

```
assistant : "Vous voulez parler de la confiture de myrtilles, n'est-ce pas ?"
heard     : "Vous voulez parler de la machine télévonnaire, oui."
heard     : "Vous voulez parler de la vie, c'est pas une machine tes millionnaires."
heard     : "Vous voulez parler de la machine qui est millionnaire ici un euro c'est vingt…"
```

Each pass degrades further, and the STT eventually drifts across languages — we saw
Cyrillic and English fragments appear despite `--parakeet_tdt_language fr`.

## Suggested change

1. Reword the help text, e.g. *"single-channel speech enhancement (DeepFilterNet).
   Reduces background noise. It does not cancel acoustic echo from your own playback."*
2. Optionally cross-reference `--local_audio_block_mic_during_playback` there, since that
   is what users reaching for this flag are usually looking for. Its cost is barge-in,
   which is worth stating in the same breath.
3. Larger and entirely optional: a real AEC in the local audio client, where the playback
   stream is already available as a reference (`PlaybackBuffer` in
   `api/openai_realtime/audio_client.py`). That is the only route that keeps barge-in on
   open speakers.

Happy to send a PR for (1) and (2) if useful.

Hotragn:
Fixes #483

## Problem

The help text promises *"noise reduction, equalization, and echo cancellation"*. The implementation is DeepFilterNet, a single-channel speech enhancer — `VAD/vad_handler.py` calls:

```python
enhance(self.enhanced_model, self.df_state, <microphone array>)
```

with only the near-end signal. Acoustic echo cancellation needs the **far-end reference** — the samples that went to the speaker — so an adaptive filter can estimate the room path and subtract it. No such reference exists at that point in the pipeline, so this claim cannot be satisfied by this implementation, now or after any amount of tuning. "Equalization" also isn't what a denoiser does.

As #483 argues, it's an expensive claim to get wrong: acoustic feedback is the first thing users meet running `speech-to-speech local` on laptop speakers, and this was the only flag that looked like it addressed it. The flag that *does* help — `--local_audio_block_mic_during_playback` — was easy to miss, because its own (accurate) help text never mentioned echo or feedback.

## Fix

Documentation only; neither flag's behaviour changes.

- **`--audio_enhancement`** now says what DeepFilterNet actually does, states plainly that it does not cancel acoustic echo *and why* (no reference to what was played), and points at the flag that does.
- **`--local_audio_block_mic_during_playback`** now names acoustic feedback so it's findable from the symptom, while still documenting the barge-in trade-off it makes.

I fixed both halves deliberately: correcting only the false claim would leave a user with feedback and no idea what to reach for instead, which is the actual reported pain.

## Tests

`tests/test_audio_enhancement_help.py` pins the claim so it can't come back — help text has no other guard, and this one survived a long time. It asserts the text doesn't promise echo cancellation or equalization, still describes what the flag does (DeepFilterNet, noise, default), cross-references the working flag, and that the block-mic help stays discoverable without losing the barge-in caveat.

All five fail against the previous wording.

Full suite: **1185 passed, 1 skipped**. `ruff check`, `ruff format --check`, `mypy src/` clean.

## Note

This may also account for some of #151 ("No Acoustic Echo Cancellation") — the flag reads as though AEC is already supported. I've kept this PR to the documentation claim rather than touching that request.

### DIFF
--- a/src/speech_to_speech/arguments_classes/vad_arguments.py
+++ b/src/speech_to_speech/arguments_classes/vad_arguments.py
@@ -48,7 +48,7 @@ class VADHandlerArguments:
     audio_enhancement: bool = field(
         default=False,
         metadata={
-            "help": "improves sound quality by applying techniques like noise reduction, equalization, and echo cancellation. Default is False."
+            "help": "improves sound quality by applying techniques like noise reduction and equalization. Default is False."
         },
     )
     enable_realtime_transcription: bool = field(


src/speech_to_speech/arguments_classes/local_audio_arguments.py
```diff
@@ -26,7 +26,11 @@ class LocalAudioArguments:
     local_audio_block_mic_during_playback: bool = field(
         default=False,
         metadata={
-            "help": "Pause local microphone capture while audio is playing. Disabled by default so barge-in works."
+            "help": (
+                "Pause local microphone capture while audio is playing. This is what stops the "
+                "assistant hearing itself through laptop speakers (acoustic feedback); "
+                "--audio_enhancement cannot do that. Disabled by default so barge-in works."
+            )
```
andimarafioti: I wouldn't change this one

tests/test_audio_enhancement_help.py
```diff
@@ -0,0 +1,56 @@
+"""The `--audio_enhancement` help text must not promise echo cancellation.
+
+The flag runs DeepFilterNet, a single-channel speech enhancer: `VAD/vad_handler.py` calls
+`enhance(model, df_state, <mic array>)` with only the near-end signal. Acoustic echo
+cancellation needs the far-end reference — the samples that went to the speaker — so an
```
andimarafioti: This whole test file is unnecesary

src/speech_to_speech/arguments_classes/vad_arguments.py
```diff
@@ -48,7 +48,13 @@ class VADHandlerArguments:
     audio_enhancement: bool = field(
         default=False,
         metadata={
-            "help": "improves sound quality by applying techniques like noise reduction, equalization, and echo cancellation. Default is False."
+            "help": (
+                "Run DeepFilterNet single-channel speech enhancement over captured audio to suppress "
+                "background noise. This does not cancel acoustic echo: enhancement only sees the "
+                "microphone signal, with no reference to what was played, so it cannot subtract the "
+                "speaker output. For feedback from laptop speakers use "
+                "--local_audio_block_mic_during_playback, or a headset. Default is False."
```
andimarafioti: Here you can just put:
"improves sound quality by applying techniques like noise reduction and equalization. Default is False."

andimarafioti (changes requested):
The goal of this PR is ok, but the reach is far too broad. I would just change the small helper text in vad_arguments

Hotragn:
All three addressed in `350bb6d` — the PR is now 1 file, +1/-1.

- `vad_arguments.py`: applied your exact wording.
- `local_audio_arguments.py`: reverted, untouched.
- `tests/test_audio_enhancement_help.py`: deleted.

You were right that this was over-built. Removing the false claim was the whole ask; the cross-reference and the test file were scope I added on my own, which is the same thing you flagged as P3 on #424. Rebuilt on current `main` rather than patching over the old commits, so the diff is just the one line.

Full suite 1198 passed, 1 skipped; ruff, format and mypy clean.

andimarafioti (approved):
Thank you!

PR #505, opened 2026-08-18
Feat/gemini realtime voice integration

Phyrakset:


andimarafioti:
It doesn't make sense to connect this to gemini live, or any of the live voice API providers. Those implement everything that's here in their backends.

PR #513, opened 2026-08-19
Fix(STT): Guard `torch.mps.empty_cache()`  when not on MPS.

salignatmoandal:
hey again, (:

Paraformer and LightningWhisper called `torch.mps.empty_cache()` after every inference. On Linux/CUDA this can fail because MPS is not available.

Same pattern as `ChatTTSHandler` in `src/speech_to_speech/TTS/chatTTS_handler.py`: only clear MPS cache when `device == "mps"`.

## Test plan
- [x] `uv run pytest tests/test_paraformer_transcription_events.py -v`


andimarafioti (approved):
Good work! Thank you :)

PR #517, opened 2026-08-21
Keep conversation transcripts out of operational logs

### LINKED ISSUE #475: Keep conversation transcripts out of operational logs
## Problem

Operational logs can include user or assistant transcript text. These logs may be retained by service managers, containers, or hosted logging systems, so conversation content should not be written through normal application loggers by default.

## Desired behavior

Operational logs should be content-free by default while retaining useful diagnostics such as processing stage, event type, status, language, identifiers, token or character counts, and timing information.

Full transcripts are nevertheless valuable when debugging STT, LLM, TTS, and Realtime behavior. Provide an explicit opt-in flag such as `--log-transcripts`, disabled by default. When enabled, it may emit full user and assistant transcript content for diagnostics and must display a prominent startup warning that sensitive conversation data can be written to retained logs.

The default-off policy applies to transcript-bearing log records across STT, LLM, TTS, and Realtime server/client paths, including error handling when an exception or validation message may echo transcript content.

## Acceptance criteria

- With transcript logging disabled, user and assistant transcript text is not emitted through application loggers at any configured log level.
- An explicit `--log-transcripts` opt-in enables the transcript logging needed to debug STT, LLM, TTS, and Realtime flows.
- Enabling transcript logging emits a prominent warning before conversation processing begins, explaining that sensitive conversation content may be retained wherever logs are collected.
- Tool arguments or error details follow the same default-off policy when they contain conversation transcript text.
- Normal logs retain content-free operational metadata needed for debugging.
- Regression tests assert that sentinel transcript content is absent by default and present only when the explicit opt-in is enabled, including representative STT, LLM, TTS, Realtime, and error paths.
- Existing conversation processing and OpenAI Realtime protocol events are unchanged.

## Out of scope

- Terminal or Rich conversation display controls.
- Changes to conversation behavior, protocol payloads, or client-visible events.
- A broader privacy mode or persistence policy.

Hotragn:
Fixes #475

## Problem

Application loggers carried user and assistant transcript text. Those logs are retained by service managers, containers and hosted logging systems, so conversation content outlived the conversation in places the user never agreed to.

One leak was at **INFO** level, so it applied to default deployments rather than only debug runs:

```python
logger.info("Transcription completed (language=%s): %s", language_code, transcript)
```

Truncation didn't help either — a 40- or 80-character prefix of what someone said is still what they said.

## Fix

Every content-bearing logger call across the four paths now logs a measurement instead of the text:

| path | sites |
|---|---|
| **STT** | `transcription_notifier` (partial + both final branches), `faster_whisper_handler` segment logging |
| **LLM** | `base_openai_compatible_language_model` (both "Clean text" sites), `language_model`, `lm_output_processor` "Forwarding to TTS" |
| **TTS** | `facebookmms_handler` (tokenizing, processing), `pocket_tts_handler` |
| **Realtime** | `handlers/conversation.py`, which logged *both* the emitted transcript prefix and the incoming hypothesis |

Diagnostics are preserved — stage, event type, language, identifiers, character counts and timing — so the logs remain useful for debugging turn flow.

Deliberately **out of scope and untouched**: `console.print` conversation display. That's an operator watching a live session, not a retained log, and the issue excludes it. Protocol payloads and conversation behaviour are unchanged; `PartialTranscriptionEvent` and friends still carry the full text, and there's a test asserting that.

One thing I looked at and left alone: `LLM/tool_call/function_call.py` logs the *names* of dropped positional arguments, not their values, so it isn't a leak.

## A test asserted the old behaviour

`tests/test_transcription_notifier.py::test_non_empty_final_transcription_logs_full_text_at_info` explicitly asserted that the full transcript reaches INFO:

```python
assert "Transcription completed (language=en): " + transcript in caplog.text
```

I rewrote it to assert the new contract — completion, `language=en` and `chars=N` present, content absent — rather than deleting it. Flagging it prominently since it's a deliberate behaviour change to an existing pinned expectation.

## Tests

`tests/test_transcript_log_hygiene.py` works at two levels:

1. **Behavioural** — pushes sentinel transcript text through the notifier's partial and final paths and asserts it never reaches `caplog`, including the truncated-prefix case, plus a test that the metadata *does* survive so this doesn't quietly cost observability.
2. **Source sweep** — walks every module in the package and fails any `logger.*` call that passes conversation content rather than a count. This covers the LLM, TTS and Realtime paths a behavioural test can't cheaply construct, and means a new leak fails CI wherever it's added. `len(...)` is pruned before the scan so counting is allowed.

13 of these fail against the current code, spanning all four subsystems. I tightened the sweep to include the generic `text` / `.text` cases after confirming it produces zero false positives on the fixed tree.

Full suite: **1321 passed, 1 skipped**. `ruff check`, `ruff format --check`, `mypy src/` clean.

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -56,6 +56,7 @@ from speech_to_speech.pipeline.messages import (
     TokenUsage,
 )
 from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
+from speech_to_speech.pipeline.transcript_logging import transcript_for_log
 from speech_to_speech.utils.utils import is_out_of_band, response_wants_audio
 
 logger = logging.getLogger(__name__)
@@ -666,7 +667,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 if self._turn_is_cancelled(turn):
                     logger.info("LLM generation cancelled (interruption)")
                 else:
-                    logger.debug("Clean text: chars=%d", len(state.clean_text))
+                    logger.debug("Clean text: %s", transcript_for_log(state.clean_text))
                     yield from _flush(sentence_batch)
             logger.info(f"Tools: {state.tools}")
         return (
@@ -712,7 +713,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 ):
                     state.output_emitted = True
                     yield self._chunk(turn, text=out)
-        logger.debug("Clean text: chars=%d", len(state.clean_text))
+        logger.debug("Clean text: %s", transcript_for_log(state.clean_text))
         logger.info(f"Tools: {state.tools}")
         return (
             not cancelled
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -666,7 +666,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 if self._turn_is_cancelled(turn):
                     logger.info("LLM generation cancelled (interruption)")
                 else:
-                    logger.debug(f"Clean text: {state.clean_text}")
+                    logger.debug("Clean text: chars=%d", len(state.clean_text))
                     yield from _flush(sentence_batch)
             logger.info(f"Tools: {state.tools}")
         return (
@@ -712,7 +712,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 ):
                     state.output_emitted = True
                     yield self._chunk(turn, text=out)
-        logger.debug(f"Clean text: {state.clean_text}")
+        logger.debug("Clean text: chars=%d", len(state.clean_text))
         logger.info(f"Tools: {state.tools}")
         return (
             not cancelled
--- a/src/speech_to_speech/LLM/language_model.py
+++ b/src/speech_to_speech/LLM/language_model.py
@@ -68,6 +68,7 @@ from speech_to_speech.pipeline.messages import (
     TokenUsage,
 )
 from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
+from speech_to_speech.pipeline.transcript_logging import transcript_for_log
 from speech_to_speech.utils.utils import is_out_of_band, response_wants_audio
 
 try:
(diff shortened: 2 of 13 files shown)


andimarafioti (commented):
After thinking this through, I agree that full transcripts should stay out of normal application logs by default. Moving them to DEBUG would not be sufficient because debug logs can still be collected and retained.

However, we still need a supported way to preserve transcripts when diagnosing STT, LLM, TTS, and Realtime issues. Please put transcript logging behind an explicit opt-in flag such as `--log-transcripts`, disabled by default, and emit a prominent startup warning that enabling it may write sensitive conversation content to retained logs. With the flag off, the metadata-only behavior in this PR should remain the default.

I updated #475 to capture this requirement, including coverage for the default-off and explicit opt-in paths.

Hotragn:
Implemented in `7bd1de8`, rebased onto current `main`. Agreed on the reasoning — DEBUG isn't a privacy boundary, since debug logs get collected and retained like any other.

## The gate

New `pipeline/transcript_logging.py`:

```python
transcript_for_log(text) -> "chars=42"          # default
transcript_for_log(text) -> "see you at nine"   # with --log_transcripts
```

All twelve call sites now delegate to it rather than formatting a count themselves, so each line reads the same in both modes and there's exactly one audited place where content can be emitted.

I made it a **process-global flag set once at startup**, rather than plumbing a parameter through every handler constructor. The reason is the default: it has to be "off" everywhere, including in paths that never consider the question, and a single default-off switch can't be missed by a new call site the way a constructor argument can. Happy to switch to explicit plumbing if you'd rather have it visible in the signatures.

## The flag and the warning

`--log_transcripts`, off by default. `run_pipeline_command` sets the gate and warns immediately after `setup_logger`, **before any handler is built**, so the notice lands ahead of the first turn rather than after content has already been written. The warning names the flag and spells out that journald, container logs and hosted aggregators will hold conversation data.

One naming note: it's `--log_transcripts`, not `--log-transcripts` as in your comment. `HfArgumentParser` derives flags from field names and every other flag here uses underscores, so I matched the convention — say the word if you'd prefer an explicit hyphenated alias.

## Test sweep

The source sweep from the first commit now treats `transcript_for_log(...)` as compliant, the same way it already ignored `len(...)` — those are the two audited ways to mention content in a log call. A **raw** transcript still fails the sweep, so the default-off guarantee stays enforced for new code rather than being weakened by adding the opt-in.

Both states are covered per the updated criteria: sentinel content absent by default, present when opted in (and not truncated, since opting in exists for debugging), the gate's default, `transcript_for_log` directly, the warning firing only when enabled, and the CLI field defaulting to `False`. One test asserts the gate is restored after an opt-in test, so a process-global can't leak enabled state into later tests through ordering.

Full suite **1369 passed, 1 skipped**; ruff, format and mypy clean.

One flake to flag, unrelated to this PR: on a heavily loaded full-suite run `test_llm_proxy.py::TestStreamingPassthrough::test_frames_forward_as_they_arrive_not_buffered` failed once. It passes isolated (×3), passes on clean `main` (×2), and passes on a repeat full run — it looks timing-sensitive under parallel load, like `test_is_latest_after_stability_window_survives_cancelled_reopen_candidate` did earlier.

andimarafioti:
Thank you for tackling this! i made a few changes in another PR before merging and kept your commits. Good work!

PR #523, opened 2026-08-23
Preserve chat turn order during non-interrupting back-to-back speech

### LINKED ISSUE #454: Preserve chat turn order during non-interrupting back-to-back speech
## Problem

When `interrupt_response=False`, a second transcription can be appended to `Chat` while the first response is still generating. The first response then appends its assistant history after that newer user item.

The resulting order is:

```text
user A -> user B -> assistant A
```

The queued response for B therefore receives the same incorrectly ordered provider snapshot.

## Reproduction

1. Start generation for transcript A.
2. While A is still generating, deliver non-interrupting speech/transcription B.
3. Let A finish and commit its assistant response.
4. Inspect `Chat.buffer` or the provider input for B.

This reproduces on unmodified `main` and affects both local/OpenAI-compatible history paths and both realtime transports.

## Additional provisional-history failure mode

With #453's provisional response history, the same overlap can also make rollback incomplete:

1. Response A writes provisional assistant output.
2. Non-interrupting turn B arrives before A's delivery is final.
3. `trim_if_needed()` compacts the buffer and summarizes A's provisional output, replacing its tracked item IDs.
4. A is cancelled, so rollback removes the original IDs but cannot remove A's content from the summary.

A fix must keep turn order correct and prevent compaction from absorbing provisional items that may still be rolled back.

## Expected behavior

Conversation history and the provider input for B should preserve turn order:

```text
user A -> assistant A -> user B
```

Cancelling A must also remove all of A's provisional content, including when the chat exceeds its compaction threshold.

## Scope note

The base ordering race was found while reviewing #453 but is pre-existing on `main`, so it is tracked separately rather than expanding that PR. The provisional-compaction variant depends on that same overlapping-turn race and belongs in this fix as well.

ousamabenyounes:
## Summary

Fix #454 — the base ordering race.

When `interrupt_response=False`, a response committed its assistant history by
appending to the chat buffer. A transcription that landed while the model was
still running therefore ended up *ahead* of the answer it did not interrupt:

```text
user A -> user B -> assistant A
```

Each generation now snapshots a history anchor (`Chat.history_anchor_id()`)
before it starts and writes its output back into its own turn, so later user
turns are left untouched:

```text
user A -> assistant A -> user B
```

- `Chat` gained `history_anchor_id()` plus an `after_item_id=` argument on
  `add_item`, `add_ordered_function_call` and `add_provisional_generation_items`.
  Without that argument every call behaves exactly as before (plain append).
- Both generation paths capture the anchor before generating:
  `BaseLanguageModelHandler` (local models) and `BaseOpenAICompatibleHandler`
  (text and audio turns, including the mid-stream tool-call commit). The audio
  turn anchors on the user message it writes itself.
- If the anchor is gone (evicted, compacted, rolled back) the item is appended,
  i.e. today's behaviour.

Once an unpaired ordered function call can sit mid-buffer, the two serializers
needed a matching change: they used to `break` on such a call, which was
harmless only while it was always the last item. They now omit just the turn
that call opened, so a newer user turn still reaches the provider
(`_drop_unpaired_ordered_turns_locked`). Same output as before when the call is
at the tail.

The issue's second failure mode (compaction absorbing provisional output) is
already handled on `main`: `trim_if_needed` defers the compactor and
`_snapshot_for_compaction` bails out while `_provisional_generations` is
non-empty, covered by `TestCompaction::test_cancelled_provisional_history_is_not_compacted`.
Nothing was added for it.

Not covered here: the anchor is taken when the LLM thread picks a request up.
A request that waits in the queue behind an in-flight response can still be
anchored after a newer transcript. That is never worse than today, and closing
it means carrying the anchor on `GenerateResponseRequest` — happy to do that in
a follow-up if you'd prefer it in scope.

## Test verification (RED -> GREEN)

RED — the 11 new tests on unmodified `main` (only the tests applied, `src/` untouched):

```text
E       AssertionError: assert ['A', 'B', 'answer A'] == ['A', 'answer A', 'B']
E       AssertionError: assert ['A', 'B', 'answer A'] == ['A', 'answer A', 'B']
FAILED tests/test_chat_turn_order.py::test_local_response_history_precedes_speech_that_arrived_during_generation
FAILED tests/test_chat_turn_order.py::test_keyless_commit_places_text_and_tool_calls_in_their_own_turn
FAILED tests/test_chat.py::TestTurnOrdering::test_generation_output_stays_ahead_of_speech_that_arrived_meanwhile
FAILED tests/test_chat.py::TestTurnOrdering::test_multi_item_generation_keeps_its_own_emission_order
FAILED tests/test_chat.py::TestTurnOrdering::test_generation_started_on_empty_history_precedes_later_speech
FAILED tests/test_chat.py::TestTurnOrdering::test_output_is_appended_when_its_anchor_turn_is_gone
FAILED tests/test_chat.py::TestTurnOrdering::test_ordered_function_call_is_placed_in_its_own_turn
FAILED tests/test_chat.py::TestTurnOrdering::test_unpaired_ordered_call_hides_its_own_turn_but_not_later_speech
FAILED tests/test_chat.py::TestTurnOrdering::test_paired_call_is_serialized_once_its_output_arrives_after_later_speech
FAILED tests/test_chat.py::TestTurnOrdering::test_second_commit_of_a_turn_follows_the_items_it_already_wrote
FAILED tests/test_responses_api_language_model.py::test_response_history_precedes_speech_that_arrived_during_generation
11 failed in 7.61s
```

The two end-to-end tests show the reported symptom directly:

```text
E       AssertionError: assert ['A', 'B', 'answer A'] == ['A', 'answer A', 'B']
E         At index 1 diff: 'B' != 'answer A'
```

GREEN — same tests on this branch, and the full suite:

```text
=== ruff-check ===
All checks passed!
=== ruff-format ===
155 files already formatted
=== mypy ===
Success: no issues found in 96 source files
=== pytest ===
1263 passed, 1 skipped, 4 warnings in 59.05s
```

Baseline on `main` with the same commands: `1252 passed, 1 skipped`. No test
changed state; the 11 added tests account for the difference.

## Files changed

| File | Change |
|------|--------|
| `LLM/chat.py` | `history_anchor_id()`, turn-anchored insertion, unpaired-ordered-call serialization boundary |
| `LLM/language_model.py` | capture the anchor before generating, pass it through `_commit_ordered_output` |
| `LLM/base_openai_compatible_language_model.py` | same for the text and audio turns via `_Turn.history_anchor_id` |
| `tests/test_chat.py` | `TestTurnOrdering` — ordering, anchor fallback, serializer boundary |
| `tests/test_chat_turn_order.py` | local-model handler ordering end to end |
| `tests/test_responses_api_language_model.py` | OpenAI-compatible handler ordering end to end |

### DIFF
--- a/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
+++ b/src/speech_to_speech/LLM/base_openai_compatible_language_model.py
@@ -121,6 +121,9 @@ class _Turn(BaseModel):
     wants_audio: bool
     response_key: str
     prefetch_transaction: ResponsePrefetchTransaction | None = None
+    # End of the conversation when this turn started; keeps its output ahead of
+    # user messages appended while the model was still running.
+    history_anchor_id: str | None = None
 
 
 class _GenState(BaseModel):
@@ -563,6 +566,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             recorded_items = chat.add_provisional_generation_items(
                 turn.response_key,
                 [*state.pending, fc_item],
+                after_item_id=turn.history_anchor_id,
             )
             state.pending.clear()
             if recorded_items is None:
@@ -848,6 +852,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                             committed_item_ids=(
                                 {transactional_user_message_id} if transactional_user_message_id is not None else None
                             ),
+                            after_item_id=turn.history_anchor_id,
                         )
                         if recorded_items is None:
                             can_commit = False
@@ -930,6 +935,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             return
 
         original_chat = runtime_config.chat
+        history_anchor_id: str | None = None
         if not is_out_of_band(response) and original_chat.has_pending_tool_calls():
             yield EndOfResponse(
                 turn_id=turn_id,
@@ -996,6 +1002,9 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
                 return
             assert provisional_message.id is not None
             transactional_user_message_id = provisional_message.id
+            # This turn writes its own user message, so anchor its output after
+            # that message: speech arriving later must not overtake it.
+            history_anchor_id = transactional_user_message_id
 
             def commit_audio_history() -> None:
                 original_chat.compact_audio_history(self.audio_history_turns)
@@ -1016,6 +1025,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             wants_audio=wants_audio,
             response_key=request.response_key,
             prefetch_transaction=request.prefetch_transaction,
+            history_anchor_id=history_anchor_id,
         )
         yield from self._generate(
             active_chat,
@@ -1052,6 +1062,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             return
 
         original_chat = runtime_config.chat
+        history_anchor_id = original_chat.history_anchor_id()
         if not is_out_of_band(response) and original_chat.has_pending_tool_calls():
             yield EndOfResponse(
                 turn_id=turn_id,
@@ -1110,6 +1121,7 @@ class BaseOpenAICompatibleHandler(BaseHandler[LLMIn, LLMOut], ABC):
             wants_audio=wants_audio,
             response_key=request.response_key,
             prefetch_transaction=request.prefetch_transaction,
+            history_anchor_id=history_anchor_id,
         )
         yield from self._generate(active_chat, original_chat, turn, optional_kwargs)
 
--- a/src/speech_to_speech/LLM/chat.py
+++ b/src/speech_to_speech/LLM/chat.py
@@ -45,6 +45,11 @@ logger = logging.getLogger(__name__)
 
 AUDIO_INPUT_HISTORY_PLACEHOLDER = "[User audio input]"
 
+# Anchor returned by :meth:`Chat.history_anchor_id` for an empty conversation,
+# so a generation that starts from empty history is still written before user
+# messages that arrive while it runs.
+HISTORY_START_ANCHOR = "__history_start__"
+
 
 class ChatItemError(Exception):
     """Raised when a conversation item fails validation in :meth:`Chat.add_item`."""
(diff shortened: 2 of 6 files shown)


ousamabenyounes:
Pushed the verified version of this branch (force-update: the earlier commit was from an interrupted run).

Local CI replay, scoped to the `ci.yml` jobs a `src/`+`tests/` Python diff can affect — `ruff` (`ruff check` + `ruff format --check` on `src/ tests/`), `mypy src/`, and `pytest tests/` — run on the upstream baseline and on this commit:

| | `main` (9f59bc7) | this branch |
|---|---|---|
| ruff check / format | pass | pass |
| mypy `src/` | pass | pass |
| pytest `tests/` | 1252 passed, 1 skipped | 1263 passed, 1 skipped |

Same pass set, no test changed state; the 11 added tests account for the delta. Diff coverage on changed production lines is 100% (54 lines, `diff-cover` against `origin/main`).

Not replayed locally: `agents-sdk-websocket`, `agents-sdk-webrtc`, `package` and `install-smoke` — they exercise the demo front-end and packaging, which this diff does not touch. CI here covers them.

One thing worth a look during review: the serialization change. Anchoring a response back into its own turn means an unpaired ordered `function_call` can now sit mid-buffer, and both serializers previously `break`-ed on one — correct only while such a call was always last. They now skip just the turn that call opened, so a newer user turn still reaches the provider. `TestTurnOrdering::test_unpaired_ordered_call_hides_its_own_turn_but_not_later_speech` and `::test_paired_call_is_serialized_once_its_output_arrives_after_later_speech` pin both halves of that.

andimarafioti (approved):
Great work! You took a hard one and delivered a great implementation, kudos!

ousamabenyounes:
Thank you so much, @andimarafioti — this means a lot ! 
Congratulation to all HuggingFace team about the news we just heard ! 
Incredible and so nice! You're the github of IA ! and we wish you lot of success :)

I’d also like to share something behind this contribution: 
it was produced end-to-end by an AI agentic system I’ve been building for the past 18 months. 
It investigates the issue, reproduces the bug, implements the fix, writes and runs the tests, and reviews its own work through multiple quality gates. If the result does not meet the required standard, it does not open a PR.

This approach has now led to more than 300 merged contributions across major open-source projects, and we intend to push it much further (all visible on my public github profile)

As someone deeply invested in both AI and voice AI, it is a real honor to contribute to Hugging Face projects — here in speech-to-speech, but also in projects such as LeRobot. Seeing this work recognized and merged by your team makes me incredibly proud.

Thank you again for the review, the trust, and the kind words!

bbmonica46-cyber:
image

PR #524, opened 2026-08-23
Fix(TTS): strip the -auto suffix before mapping a language

Hotragn:
## Problem

With `--language auto` the STT handlers report `"de-auto"` rather than `"de"`, so downstream code knows the language may change between turns. Both Kokoro and Facebook MMS pass that value straight into language maps keyed on **bare** Whisper codes, so the lookup never matches:

```python
>>> WHISPER_LANGUAGE_TO_KOKORO_LANG.get("de")          # 'b'
>>> WHISPER_LANGUAGE_TO_KOKORO_LANG.get("de-auto")     # None  -> falls back
>>> WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE["de"]        # 'deu'
>>> WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE["de-auto"]   # KeyError
```

- **Kokoro** silently keeps its startup voice and never switches language — the feature that code exists to provide. Its `"Language change detected"` log line never fires, so there's no signal that anything went wrong.
- **Facebook MMS** catches the `KeyError` and reports `"Unsupported language: de-auto"` before falling back to English — for a language it fully supports. The message blames the language rather than the suffix.

So `--language auto` with either backend speaks every language in the startup voice. `tests/test_facebookmms_language_map.py` (added in #420) already documents this failure mode — *"Wrong keys silently fall back to English in `load_model`"* — but only checks the map's keys, not the suffixed input.

## Fix

Both handlers normalize once where they read `tts_input.language_code`, via the existing `resolve_auto_language()` — the canonical implementation of this convention — so every comparison and lookup downstream sees a bare code.

Normalizing at the **read site** rather than at each lookup also fixes a second MMS bug: the reload guard compared `self.language != language_code`, so an already-loaded `"de"` looked different from an incoming `"de-auto"` and forced a redundant model reload every turn.

`None` is preserved, since it means "no per-turn language" rather than an unknown one.

Source diff is 8 lines across 2 files.

## Scope

`SupertonicTTSHandler` already strips the suffix in its own `_normalize_language_code`, so it's unaffected and I left it alone — it could adopt the shared helper later if you'd prefer one implementation. Also worth noting: this imports `resolve_auto_language` from `LLM/utils` into `TTS/`, which is a slightly odd direction. I reused it rather than duplicating a third normalizer, but I'm happy to move the helper somewhere neutral (e.g. alongside the pipeline messages that carry the convention) if you'd rather not have that import.

## Tests

`tests/test_tts_auto_language_suffix.py` drives both handlers' `process()` with a suffixed code and asserts the bare code reaches the language decision, that a pinned code and `None` pass through unchanged, that MMS no longer reloads when the language already matches, and that a genuinely unsupported language still warns (normalizing mustn't swallow the real case).

Three fail against current code. The MMS one fails with the symptom itself:

```
assert requested == ["de"]
E   AssertionError: assert [] == ['de']
WARNING  facebookmms_handler.py:186 Unsupported language: de-auto
```

Full suite: **1264 passed, 1 skipped**. `ruff check`, `ruff format --check`, `mypy src/` clean.

### DIFF
--- a/src/speech_to_speech/TTS/facebookmms_handler.py
+++ b/src/speech_to_speech/TTS/facebookmms_handler.py
@@ -11,6 +11,7 @@ from rich.console import Console
 from transformers import AutoTokenizer, VitsModel
 
 from speech_to_speech.baseHandler import BaseHandler
+from speech_to_speech.LLM.utils import resolve_auto_language
 from speech_to_speech.pipeline.cancel_scope import CancelScope
 from speech_to_speech.pipeline.handler_types import TTSIn, TTSOut
 from speech_to_speech.pipeline.messages import AUDIO_RESPONSE_DONE, EndOfResponse
@@ -162,7 +163,9 @@ class FacebookMMSTTSHandler(BaseHandler[TTSIn, TTSOut]):
             speculative_turns.commit(tts_input.turn_id, tts_input.turn_revision)
 
         gen = self.cancel_scope.generation if self.cancel_scope else None
-        language_code = tts_input.language_code
+        # STT reports "de-auto" in auto-detect mode; WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE is
+        # keyed on bare codes, so the suffix has to go before the lookup and the comparisons.
+        language_code, _ = resolve_auto_language(tts_input.language_code)
         text = tts_input.text
 
         console.print(f"[green]ASSISTANT: {text}")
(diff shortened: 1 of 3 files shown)


andimarafioti:
Thanks for the contribution. I tested the current production pipeline with `de-auto` through both LLM paths and `LMOutputProcessor`; in every case, it is normalized to `de` before reaching TTS.

The failure only occurs when these TTS handlers are invoked directly with an input the production pipeline does not send. Adding duplicate normalization to each handler—and coupling TTS to `LLM.utils`—is more defensive than warranted here.

Since the reported production bug is not reproducible, I’m closing this PR. We can revisit if a production path that bypasses the existing normalization is demonstrated.

Issue #536, opened 2026-08-28
Support stateful streaming STT sessions instead of repeated whole-utterance uploads


andimarafioti:
## Summary

Add a stateful streaming STT backend for live microphone audio. The current OpenAI-compatible STT backend is request-oriented: it sends accumulated or final audio as an in-memory WAV to `POST /v1/audio/transcriptions`. Progressive transcription therefore re-uploads the accumulated utterance, while final transcription independently uploads the full utterance again and waits for the complete response.

Keep the current file/request backend for compatible servers, but add a transport that streams each audio chunk once over a persistent transcription session and commits the buffered turn when speech-to-speech's VAD/Smart Turn logic decides the user turn is ready.

This is a follow-up to #367, where stateful/bidirectional microphone transcription was explicitly a non-goal, and is related to the deferred streaming-STT work in #363.

## Current limitation

`OpenAICompatibleSTTHandler` currently:

- encodes the current 16 kHz mono utterance as a complete WAV;
- calls `POST /v1/audio/transcriptions` for each dispatched progressive snapshot;
- allows at most one progressive request in flight and drops newer progressive work while it runs;
- sends the final transcription as a separate request without reusing an active progressive session or its result.

This preserves broad `/audio/transcriptions` compatibility, but it has poor scaling characteristics for live audio:

- progressive bandwidth and provider work grow with the accumulated utterance;
- the final request repeats the full upload;
- the LLM still waits for a new final request after local turn detection;
- the adapter cannot use remote STT implementations that return transcript deltas while audio is arriving.

OpenAI's current live-audio workflow uses a Realtime transcription session: audio is appended incrementally, transcript deltas arrive during speech, and the application can explicitly commit the input buffer at its own turn boundary. See: https://developers.openai.com/api/docs/guides/realtime-transcription

## Proposed behavior

Introduce a separate stateful streaming STT backend or transport capability, for example `--stt openai-realtime`, without changing the existing `--stt openai` file/request contract.

The intended ownership is:

- speech-to-speech continues to own input VAD, Smart Turn, turn revisions, session state, and response admission;
- the remote transcription service owns only incremental ASR inference;
- upstream automatic turn detection is disabled when supported;
- incoming PCM is appended to the active remote transcription buffer once;
- speech-to-speech explicitly commits the buffer when its local turn decision is ready;
- remote transcript deltas become `PartialTranscription` output;
- the committed final transcript becomes the existing final `Transcription` and triggers LLM work;
[...]

PR #537, opened 2026-08-28
Add stateful streaming STT backends

andimarafioti:
## Summary

- add separate `openai-realtime` and experimental `vllm-realtime` STT backends
- stream PCM over persistent WebSocket sessions with local VAD commit/discard boundaries
- preserve the existing request-oriented `openai` backend and pipeline defaults
- add reconnect, reopen, cancellation, session reuse, resampling, protocol-dialect, item-fencing, failure-attribution, and timing tests

Refs #536.

## Current status

This PR remains a draft for review. Initial connection/session-setup failures are now latched to the subsequently assigned logical turn even when no PCM was sent. Reopened revisions of that failed turn are rejected, and a later independent turn can reconnect. Pre-commit item failures close the connection to discard retained remote PCM while preserving same-turn failure fencing; post-commit item failures continue reusing the connection.

The review fixes preserve accepted queued audio across discards and retain short fragments when their continuation begins inside the merge window. Discarding an active buffer whose remote item ID is still unknown now closes the connection, preventing delayed transcripts from entering the next utterance. Commit deadlines run in the session worker from the VAD boundary, including when a stale final is never consumed; expiry rejects the affected turn and releases later commands. Caller waits use the same boundary-based deadline and return even while setup or socket I/O occupies the worker, which retains responsibility for cleanup. OpenAI transcript deltas may omit `content_index`, while explicit indexes and terminal-event identity remain validated.

Speech-start metadata is enqueued before notifying the streaming worker in both VAD speech-start paths, so buffered partials reach the Realtime service after their item mapping exists. Commit-to-final latency is logged when the worker accepts the final, including stale finals that are never consumed; later processing delays are excluded.

The environment API-key fallback is resolved after official endpoint normalization. Both `gpt-transcribe` and `gpt-live-transcribe` send language hints through `languages`, matching the [official committed-turn configuration example](https://developers.openai.com/cookbook/examples/migrating_from_whisper_to_gpt_transcribe#6-use-gpt-transcribe-in-a-committed-turn-realtime-session) for `gpt-transcribe`.

## Verification

Local validation at `38cc11d3ec15b1b64cb8e156ba4080b86c8a8872`:

- streaming STT tests: 89 passed, including local fake WebSocket servers for both dialects
- full suite: 1569 passed, 3 skipped
- mypy: passed for 106 source files
- Ruff lint and changed-file formatting: passed
- `git diff --check`: passed

Regression coverage includes delayed events after discarding unknown and known items, unconsumed stale commits, queued commit expiry, bounded caller waits during stalled setup and slow close, rejection of late partials/finals, same-turn failure retention, later-turn recovery, and deltas with omitted or invalid content indexes. Scheduling-controlled tests cover speech-start ordering through the worker, notifier, and Realtime service in both VAD branches. Receipt-timing tests cover delayed consumption and unconsumed stale finals. These tests use simulated providers; live-provider validation is not claimed.

## Scope

Provider-side VAD remains disabled. Existing `--stt openai`, LLM, and TTS behavior are unchanged. This PR does not add an inference server or merge any release work.


tavallaie (commented):
Two notes from reading the new streaming STT handler. I am holding the VAD silence-forwarding, vLLM default model, and OpenAI query-string comments until live `gpt-live-transcribe` / Voxtral runs can confirm them.

src/speech_to_speech/STT/streaming_handler.py
```diff
@@ -0,0 +1,1066 @@
...
+from collections import deque
+from dataclasses import dataclass
+from queue import Empty, Queue
+from threading import Event, Lock, Thread
+from time import monotonic, perf_counter
+from typing import Any, Callable, Iterator, Literal, Protocol
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
+
+import numpy as np
+import soxr
```
tavallaie, line 16: `soxr` is not a direct dependency in `pyproject.toml`. It arrives only through `librosa`. A missing transitive package fails this import before the first socket opens.

Add `soxr` as a direct dependency, or reuse `_StreamingFIRResampler` from the OpenAI TTS handler.

src/speech_to_speech/STT/streaming_handler.py
```diff
@@ -0,0 +1,1066 @@
...
+        connect_factory: ConnectFactory | None = None,
+        pipeline_index: int | None = None,
+    ) -> None:
+        if not model.strip():
+            raise ValueError("Streaming STT requires a model")
+        if audio_sample_rate <= 0:
+            raise ValueError("Streaming STT audio_sample_rate must be > 0")
+        if connect_timeout <= 0 or final_timeout <= 0:
+            raise ValueError("Streaming STT timeouts must be > 0")
+        if self.protocol_type is VLLMRealtimeProtocol and audio_sample_rate != PIPELINE_SAMPLE_RATE:
```
tavallaie, line 952: vLLM rejects a wrong rate in `setup()`. OpenAI does not. Public transcription PCM is 24 kHz only, so a wrong `--openai_realtime_stt_audio_sample_rate` fails only after the live socket is open.

Reject other rates in `OpenAIRealtimeSTTHandler.setup()` the same way.

src/speech_to_speech/s2s_pipeline.py
```diff
@@ -506,7 +509,7 @@ def _build_pipeline_unit(
         default_instructions=default_instructions,
     )
 
-    if module_kwargs.enable_live_transcription:
+    if module_kwargs.enable_live_transcription and not stt_selection.spec.capabilities.streams_audio_chunks:
```
tavallaie, line 512: Question on the intended meaning of `--enable_live_transcription` for these backends.

This line correctly skips VAD's old progressive HTTP uploads when STT already streams PCM. The streaming worker still emits `PartialTranscription` on every provider delta, even if live transcription is off.

Is the flag only meant to disable those extra VAD uploads, with streaming partials always on? Or should it also suppress `emit_partial()` so a client gets no captions until the VAD commit?

Happy to treat this as working-as-designed if partials from `openai-realtime` / `vllm-realtime` should always flow.

src/speech_to_speech/VAD/vad_handler.py
```diff
@@ -556,9 +594,24 @@ def process(self, audio_chunk: VADIn) -> Iterator[VADOut]:
...
+                duration_ms = sum(len(chunk) for chunk in vad_output) / self.sample_rate * 1000
+            elif is_triggered_now:
+                duration_ms = self._speech_buffer_duration_ms()
+            else:
+                duration_ms = 0.0
+            self._discard_expired_pending_short_segment(max(0, self._audio_ms - int(duration_ms)))
+
+        # Stateful STT receives the same accepted PCM chunk as local VAD. The
+        # sink enqueues transport work without blocking this VAD thread.
+        self._stream_audio_chunk(audio_chunk)
```
tavallaie: I am live-testing this path now (`gpt-live-transcribe`, then Voxtral on vLLM). One thing I want to check: this forwards every listening chunk, including silence before speech and between turns.

If that holds on a real socket, OpenAI will buffer and bill idle PCM, and vLLM can start a generation on the first silent chunk. Streaming only while VAD is triggered (or a short segment is held) would avoid that. I will report what the live runs actually do.

src/speech_to_speech/arguments_classes/vllm_realtime_stt_arguments.py
```diff
@@ -0,0 +1,32 @@
...
+    vllm_realtime_stt_base_url: str = field(
+        default="ws://localhost:8000/v1",
+        metadata={"help": "Base URL for vLLM's Realtime API, including /v1."},
+    )
+    vllm_realtime_stt_api_key: Optional[str] = field(
+        default=None,
+        metadata={"help": "Optional bearer token for the vLLM endpoint."},
+    )
+    vllm_realtime_stt_model: str = field(
+        default="Qwen/Qwen3-ASR-1.7B",
```
tavallaie, line 18: I am setting up a Voxtral realtime vLLM server to live-test this backend. vLLM's own example serves `mistralai/Voxtral-Mini-4B-Realtime-2602`. A `session.update` with this Qwen default against that process should return `model_not_found`.

If the live run confirms that, it would help to default to the Voxtral realtime id, or document that `--vllm_realtime_stt_model` must match the served model.

src/speech_to_speech/STT/streaming_handler.py
```diff
@@ -0,0 +1,1066 @@
...
+        self.cancel_session()
+        super().on_session_end()
+
+    def cleanup(self) -> None:
+        self._session.close()
+
+
+class OpenAIRealtimeSTTHandler(StatefulStreamingSTTHandler):
+    protocol_type = OpenAIRealtimeProtocol
+    include_model_query = True
```
tavallaie, line 1073: I am about to connect this handler to `gpt-live-transcribe`. The URL becomes `wss://api.openai.com/v1/realtime?model=gpt-live-transcribe`, while the session body already sets `session.type=transcription` and `audio.input.transcription.model`.

If the live connect is rejected, the next thing I will try is `intent=transcription` instead of (or in addition to) the query model. Posting this so you can say if that query param is intentional.

tavallaie (commented):
Three more notes from reading the diff. I am live-testing `gpt-live-transcribe` and Voxtral on vLLM now, so these are things I want to check rather than proven faults.

src/speech_to_speech/STT/streaming_handler.py
tavallaie: Live check against `gpt-live-transcribe`:

- `wss://api.openai.com/v1/realtime?model=gpt-live-transcribe` is rejected: `invalid_model`. The server says this is a transcription model and cannot be used as the realtime session model. Put it in `audio.input.transcription.model` instead.
- `wss://api.openai.com/v1/realtime?intent=transcription` returns `session.created`.
- `?intent=transcription&model=gpt-live-transcribe` is also rejected (`not supported in transcription mode`).

With `include_model_query = False` and `?intent=transcription`, I got finals for three turns: first utterance, a new utterance after discard, and a reconnect after session end.

The default `OpenAIRealtimeSTTHandler` URL does not connect today. `intent=transcription` is the query that works.

src/speech_to_speech/arguments_classes/vllm_realtime_stt_arguments.py
tavallaie: Live check against a vLLM process serving `mistralai/Voxtral-Mini-4B-Realtime-2602`:

- `session.update` with the PR default `Qwen/Qwen3-ASR-1.7B` returns `model_not_found` (`The model Qwen/Qwen3-ASR-1.7B does not exist.`).
- `session.update` with `mistralai/Voxtral-Mini-4B-Realtime-2602` is accepted (no error event).
- `VLLMRealtimeSTTHandler` then transcribed two turns on that model.

So `--vllm_realtime_stt_model` must match the served id. The Qwen default fails on a Voxtral realtime server.

tavallaie:
Live-tested the OpenAI URL issue on this PR and opened a one-commit follow-up against this branch:

https://github.com/andimarafioti/speech-to-speech/pull/1

It connects hosted `gpt-live-transcribe` with `?intent=transcription` and keeps the model in `session.update`. Happy to adjust if you would rather take the change as a patch here.

src/speech_to_speech/VAD/vad_handler.py
tavallaie: Live VAD + `gpt-live-transcribe`:

I ran the real `VADHandler` with `streaming_stt_sink` set, 2s of leading silence, then speech, then trailing silence.

- About 2.3s of PCM (72 KiB at 16 kHz) was `append_audio`'d **before** `start_turn`.
- About 4.7s went after `start_turn`.
- Finals were the spoken phrase (`Love me.` / `Love me. Because I love`), not a silence hallucination on OpenAI.

So idle listening audio does reach the provider buffer. On hosted OpenAI it did not pollute the transcript in this run. I have not repeated this VAD path against vLLM, where `start_turn` sends `commit final=false` and can start a generation on whatever is already queued.

andimarafioti: It should still use the VAD, if it doesn't, that's a bug

src/speech_to_speech/arguments_classes/vllm_realtime_stt_arguments.py
andimarafioti: yes, model must match the vLLM served models, that is expected.

src/speech_to_speech/STT/streaming_handler.py
andimarafioti: makes sense, will add it

src/speech_to_speech/STT/streaming_handler.py
andimarafioti: Makes sense, will add it

src/speech_to_speech/s2s_pipeline.py
andimarafioti: enable_live_transcriptions is the poor man's approach to streaming transcriptions. Basically, every x ms we send a transcription to the service, and when we are certain of some words, we start returning them. This obviously is incompatible with a streaming model, where we should just call the streaming model and get it's responses as they come

src/speech_to_speech/VAD/vad_handler.py
andimarafioti: There is by design a bit of silence before speech and between spoken chunks to make the audio more realistic. It helps with transcriptions generally. But if you mean between turns like you finish a turn and while the assistant is speaking it's still sending audio and or storing audio to send, that's a bug.

src/speech_to_speech/arguments_classes/vllm_realtime_stt_arguments.py
andimarafioti: We can document it, it might not be obvious for everyone.

src/speech_to_speech/STT/streaming_handler.py
andimarafioti: added `soxr` as a direct dependency in 2912d47.

src/speech_to_speech/STT/streaming_handler.py
andimarafioti: added the setup check in 2912d47. Hosted OpenAI now rejects anything other than 24 kHz before opening a socket. Custom endpoints can still use their own rate.

src/speech_to_speech/s2s_pipeline.py
andimarafioti: documented it in 2912d47. Native streaming partials always flow; `--enable_live_transcription` controls the repeated whole-utterance path. Added a test for both backends with that path disabled.

src/speech_to_speech/VAD/vad_handler.py
andimarafioti: fixed in 2912d47. It now follows the VAD, keeping the configured pre-speech padding, trailing silence, and gaps while a short fragment is held. The rest of the idle audio is neither sent nor accumulated for later.

Added tests for both backends covering leading silence, silence between turns, interruptions, and no duplicated audio. The full suite passes (1584 tests, 2 skipped). I haven't rerun this against live OpenAI or Voxtral here. Could you repeat your VAD test on the updated branch?

src/speech_to_speech/arguments_classes/vllm_realtime_stt_arguments.py
andimarafioti: documented it in 2912d47, including a Voxtral example and checking `/v1/models`. Kept the Qwen default; the selected model must match what's served.

src/speech_to_speech/STT/streaming_handler.py
andimarafioti: took the fix from your PR in 84f575b, thanks for testing this. Your authorship is preserved. The model stays in the session body and the URL uses `intent=transcription`. Also added a test for a URL that already contains `model=...`.

src/speech_to_speech/VAD/vad_handler.py
tavallaie: I reran the live OpenAI VAD test on cec7492. Same setup as last time. 2s of silence, then speech, then trailing silence.

Last time about 2.3s of PCM hit `append_audio` before `start_turn`. This time it was 8128 bytes, about 0.25s at 16 kHz. After `start_turn` I got about 3.2s. The transcript was the spoken line again, not silence.

The 0.25s looks like pad plus the audio that arrives before `min_speech_ms`. The idle 2s did not go out.

I did not rerun vLLM on this HEAD.

tavallaie (approved):

PR #319, opened 2026-06-16
Add SenseVoice STT handler (FunASR)

LauraGPT:
## What

Adds a **SenseVoiceSmall** STT backend through [FunASR](https://github.com/FunAudioLLM/SenseVoice), alongside the existing Paraformer backend.

The released SenseVoiceSmall checkpoint supports Mandarin, Cantonese, English, Japanese, and Korean, and can tag language, emotion, and audio events. Its non-autoregressive local inference path is a good fit for low-latency voice agents.

## Changes

- `STT/sense_voice_handler.py`: adds `SenseVoiceSTTHandler`, rich-transcription post-processing, final/progressive events, and device-aware cache clearing.
- `arguments_classes/sense_voice_stt_arguments.py`: exposes model, device, and language options.
- `backend_registry.py`: registers `--stt sense-voice` in the current backend-registry architecture, including normalized configuration, lazy construction, speculative-turn tracking, and the `sensevoice` optional-extra hint.
- `pyproject.toml`: adds the `speech-to-speech[sensevoice]` extra.
- README and focused registry/transcription-event tests.

## Usage

```bash
pip install "speech-to-speech[sensevoice]"
speech-to-speech serve \
  --stt sense-voice \
  --sense_voice_stt_model_name iic/SenseVoiceSmall
```

## Verification

Verified locally on exact signed and DCO-compliant head `58e0b2c3dd48a0df843797285b97796a9e83c57f`, merged with upstream `main` at `83abb6b0062c2e021d1fe1043b0da4af97a308e9`:

- Full suite: **976 passed, 40 skipped, 0 failed**.
- Ruff lint and format checks over `src/` and `tests/`: passed.
- Mypy over all 93 source files and `git diff --check`: passed.
- Wheel and sdist build plus strict Twine metadata checks: passed.
  - wheel SHA-256: `f10451b88a6b2eea9043d30011098d8a36b76d2f4d8fb25a09861e9e39592e7e`
  - sdist SHA-256: `8b35127727164ffd553803b3a2f38eb0c07a36a1d77621c95b06af76eb6e8e09`
- Real CPU path `parse_arguments -> BackendSelection -> create_backend_handler -> setup -> warmup -> process` with FunASR 1.4.1 and `iic/SenseVoiceSmall` returned the exact transcript `欢迎大家来体验达摩院推出的语音识别模型。` while preserving turn ID, revision, and speech-stop metadata.

[...]

PR #538, opened 2026-08-29
Document --log_transcripts and the content-free logging default

Hotragn:
Follow-up to #517, which you merged. That PR made application logs content-free by default and added `--log_transcripts` to opt back in — but neither is mentioned in the README.

Two gaps, and the second is the one that matters:

1. The flag itself is undiscoverable.
2. So is the **behaviour change**. An operator debugging STT now sees `chars=42` where the transcript used to be, with nothing in the docs explaining why or how to get it back. That's a support question my own merged change created.

Adds the flag to the module-level parameter list plus a short note covering the default and the startup warning, since enabling it writes conversation content wherever logs are collected.

Every claim is verified against the merged implementation rather than written from memory:

```
flag exists, default = False
default gate       = False
default rendering  = chars=10
startup warns      = True
```

README-only, 7 lines. Full suite 1368 passed, 1 skipped (unchanged — no code touched).

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,7 +620,15 @@ See [ModuleArguments](./src/speech_to_speech/arguments_classes/module_arguments.
 - LLM backend (`--llm_backend`: `transformers`, `mlx-lm`, `responses-api`, or `chat-completions`)
 - TTS implementation (`--tts`)
 - logging level
+- transcript logging (`--log_transcripts`)
 - realtime pipeline pool size (`--num_pipelines`)
+
+Logs are content-free by default: transcript-bearing records report a character count rather
+than the text, because logs are commonly retained by service managers, containers and hosted
+log aggregators. Pass `--log_transcripts` to include full user and assistant transcripts when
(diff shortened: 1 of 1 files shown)


Hotragn:
Updated in `b1d0d16` after re-checking against #529, which landed after I opened this.

#529 wired the same flag into the `talk` client (accepting both `--log-transcripts` and `--log_transcripts`), so the note now says so rather than covering only the server side. Re-verified every claim against main including #529:

```
module flag default              = False
rendering                        = chars=10
client accepts --log-transcripts = True
client accepts --log_transcripts = True
client sets gate + warns         = True
```

I also checked this doesn't collide with anything in flight: the other open PRs touching README (#382, #372, #341, #319, #99) all add backend sections and none touch the module-level parameters or logging region, and nothing on main or in #512/#537 documents the flag. Still README-only, 8 lines.

andimarafioti (approved):
Good job, thank you!

PR #533, opened 2026-08-27
Switch Mac MLX default LLM to Qwen3-4B-4bit

salignatmoandal:
## Summary
Switch Mac default MLX LLM from Qwen3-4B-bf16 to Qwen3-4B-4bit for `--mac-optimal-settings`.

Live E2E on Mac: ~31s → ~1.4s. 4/4 on FR and EN voice tests. Details in #516.

Closes #516.

## Test plan
- [x] `uv run pytest tests/test_cli_defaults.py -v`
- [ ] `uv run speech-to-speech local --mac-optimal-settings` (loads 4bit without `--model_name`)

## Limitations
- Live tests: short 4-turn voice script only (FR+EN); not long chat, tools, or code.
- Same model family (Qwen3-4B), 4bit for speed; bf16 still available via `--model_name`.
- README examples updated to match the new default.
- Newer MLX models (SmolLM3, Ministral, Phi-4) benchmarked live; not ready as default until the multilingual / `--enable_lang_prompt` path is solid (separate issue).

Issue #516, opened 2026-08-20
`--mac-optimal-settings`: default `bf16` LLM ~20–30s; `4bit` ~1.5–3s


salignatmoandal:
hi (: 
## Summary

On Apple Silicon with `--mac-optimal-settings`, end-to-end voice latency (speech_stopped → first audio) is dominated by the default MLX LLM weight (`mlx-community/Qwen3-4B-Instruct-2507-bf16`). Swapping only the model to 4bit keeps the same pipeline and makes local conversation usable again.

Happy to follow up with a docs PR (recommend 4bit for conversational Mac use) if that would help.

Related but distinct from #107 (different hardware/stack; that report was ~1s on GPU with Whisper/Parler, not Mac MLX bf16 at 20–30s).

## Setup

- `speech-to-speech local --mac-optimal-settings`
- STT: Parakeet TDT (MLX)
- TTS: Qwen3-TTS CustomVoice 6bit
- Metric: existing log `Last speech detected to first speech out` (+ `MLX-LLM` lock hold time)

## Results

| Config | TTFA (speech_stopped → first audio) | MLX-LLM lock |
|---|---|---|
| Default bf16 | **21–34 s** (4 turns) | **18–31 s** |
| Same stack, `--model_name mlx-community/Qwen3-4B-Instruct-2507-4bit` | **~1.5–3 s** | **~1.3–2.0 s** |

STT final stayed ~0.05–0.7 s and TTS stage TTFA ~0.15–0.6 s in both cases — so this does not look like a VAD/STT/TTS runtime bug.

## Ask

1. Would you consider documenting 4bit (or changing the mac-optimal default) for conversational local use?
2. Related: structured TTFA / per-stage latency in `GlobalUsageMetrics` (TODO comment already mentions vad/stt/llm/tts p90) would make this kind of diagnosis much faster.

## Out of scope for this issue

Conversation “feel” after 4bit (short default system prompt, EN replies without `--enable_lang_prompt`, English-leaning TTS speaker) — separate from the 20–30s stall.

andimarafioti:
Happy to change the default model, qwen3 is also a bit old.

salignatmoandal:
> Happy to change the default model, qwen3 is also a bit old. 

[...]

Issue #534, opened 2026-08-27
`enable_lang_prompt` breaks strict chat templates (consecutive user messages)


salignatmoandal:
## Problem

With `--enable_lang_prompt`, the LLM handler appends a second `user` message:

```text
user: <STT transcription>
user: Please reply to my message in {language}.
```

Models with strict chat templates, such as `mlx-community/Ministral-3-3B-Instruct-2512-4bit`, reject consecutive `user` roles:

```text
TemplateError: conversation roles must alternate user and assistant roles
```

## Reproduction

```bash
speech-to-speech local \
  --mac-optimal-settings \
  --model_name mlx-community/Ministral-3-3B-Instruct-2512-4bit \
  --enable_lang_prompt \
  --qwen3_tts_language french
```

## Expected

`--enable_lang_prompt` should not produce an invalid chat history.

## Possible fix

Merge the language hint into the existing user message, or move it to the system prompt.

## Context

Found while benchmarking MLX models for Mac (#516). Qwen3 tolerates consecutive `user` messages, while Ministral does not.

andimarafioti:
Do you want to take this?

salignatmoandal:
Yes! sure.

Issue #107, opened 2024-09-18
Latency Optimization for Speech-to-Speech Pipeline


yatharthk2:
Hi,

I am currently running the speech-to-speech pipeline on an AWS EC2 instance (Ubuntu 20.04) with an Nvidia A10g GPU. The pipeline works well, but I am experiencing around 1 second of latency, and I am particularly interested in improving the latency of the entire speech-to-speech pipeline, especially the Text to Speech (TTS) part.

Current Setup:
EC2 Instance: Nvidia A10g GPU, 24GB GPU RAM
OS: Ubuntu 20.04
GPU Driver: NVIDIA-SMI 470.141.03
CUDA Version: 12.2
Pipeline: Using the standard setup from your repo
STT Model: Whisper large-v2
TTS Model: Parler-TTS (default)

Problem:
I’m currently facing around 1 second of latency for the entire pipeline from speech input to speech output. While the STT part works fairly well, the TTS step seems to contribute most to the latency. I would greatly appreciate any suggestions or guidance on reducing the overall latency, particularly for TTS.

Thanks!

sandorkonya:
Take a look at [this ](https://github.com/huggingface/speech-to-speech/issues/72).
The proposed method increases the TTS part.
Also it is [mentioned there](https://github.com/huggingface/speech-to-speech/issues/72#issuecomment-2328994005), that 500 ms is appended after the last chunk, which means, that 500 ms is the delay until the beginning of the LLM --> TTS steps.

yatharthk2:
Thank you for sharing the link and your discussions with the author. I understand the role of Whisper Streamer in accelerating the text-to-speech process. I also recognize the 500ms latency in ParlerTTS, but I believe I am not achieving this latency. Is there any way I can optimize Parler TTS setup to reach the 500ms target?

yatharthk2:
I am trying to make this pipeline really fast. I tried integrating styleTTS ; seems like streaming is not compatible with StyleTTS as of Now, how would you approach the latency optimization?

PR #539, opened 2026-08-30
Log per-turn STT/LLM/TTS latency at `response.done`

salignatmoandal:
hey again (:

Not sure this is useful *right now*, or if it's even on your priority list — happy to close if the timing is off.

After #516 / #533 I was trying to explain Mac turn latency and kept scraping four separate logs (Parakeet STT, MLX-LLM hold, Qwen3 TTFA, speech-stopped → first audio). That works once. It doesn't scale when you're comparing configs.

So this PR adds a small structure: one INFO line when the response actually finishes (`_end_response`), keyed by `turn_id` / revision:

```
Turn turn_3 rev=0 latency: stt=0.08s llm=1.83s tts_ttfa=0.16s e2e=2.09s mlx_lock_wait=0.00s status=completed
```

Instrumenting the Mac default path only (Parakeet + Qwen3 + MLX lock). Other backends stay `n/a`. Existing logs are unchanged.

I found it relevant for *improving* latency, not just staring at it: once you can attribute a slow turn, the next default/model change is a decision instead of a guess. Same motion as bf16 → 4bit, just with a cleaner signal.

## Test plan
- [x] `uv run pytest tests/test_turn_latency.py -v`
- [x] `speech-to-speech local --mac-optimal-settings` — completed turns emit the line; a failed turn still logs `status=failed`

No rush either way.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/speech_to_speech/api/openai_realtime/service.py
+++ b/src/speech_to_speech/api/openai_realtime/service.py
@@ -71,8 +71,8 @@ from speech_to_speech.pipeline.events import (
 from speech_to_speech.pipeline.messages import GenerateResponseRequest
 from speech_to_speech.pipeline.queue_types import TextPromptItem
 from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
-from speech_to_speech.pipeline.turn_latency import TurnLatencyStore, TurnLatencyTracker
 from speech_to_speech.pipeline.transcript_logging import log_exception, transcript_for_log
+from speech_to_speech.pipeline.turn_latency import TurnLatencyStore
 from speech_to_speech.utils.utils import _generate_id
 
 logger = logging.getLogger(__name__)
@@ -236,7 +236,6 @@ class ConnState(BaseModel):
     speculative_user_speech_stopped_at_s: Optional[float] = None
     speculative_user_item_id: Optional[str] = None
     speculative_audio_duration_s: float = 0.0
-    turn_latency: TurnLatencyTracker = Field(default_factory=TurnLatencyTracker)
     # Client conversation.item.create items that arrived while a response was
     # generating. Applying them mid-generation races the LLM handler's chat
     # write-back (cross-thread), so they are buffered here and flushed in order
@@ -359,6 +358,7 @@ class RealtimeService:
                 st.response_usage.input_tokens += input_tokens
                 st.response_usage.output_tokens += output_tokens
             self.total_usage += st.response_usage
+            self.turn_latency_store.clear_session(conn_id)
             logger.info(
                 "Session %s unregistered — cumulative: input_tokens=%d, output_tokens=%d, audio=%.2fs",
                 conn_id,
@@ -369,6 +369,21 @@ class RealtimeService:
 
     def _state(self, conn_id: str) -> ConnState:
         return self._conns[conn_id]
+
+    def bind_response_latency_tracker(
+        self,
+        conn_id: str,
+        response_key: str,
+        *,
+        turn_id: str | None = None,
+        turn_revision: int | None = None,
+    ) -> None:
+        self.turn_latency_store.get_or_create_response(
+            response_key,
+            turn_id=turn_id,
+            turn_revision=turn_revision,
+            session_id=conn_id,
+        )
 
     @property
     def connection_ids(self) -> list[str]:
@@ -670,9 +685,6 @@ class RealtimeService:
             st.speculative_user_turn_id = event.turn_id
             st.speculative_user_turn_revision = event.turn_revision
             st.speculative_user_speech_stopped_at_s = event.speech_stopped_at_s
-            tracker = self.turn_latency_store.get_or_create(event.turn_id, event.turn_revision)
-            if tracker is not None:
-                st.turn_latency = tracker
 
         queue = self.text_prompt_queue
         if queue and transcript:
@@ -682,6 +694,12 @@ class RealtimeService:
                 turn_id=event.turn_id,
                 turn_revision=event.turn_revision,
                 speech_stopped_at_s=event.speech_stopped_at_s,
+            )
+            self.bind_response_latency_tracker(
+                conn_id,
+                request.response_key,
+                turn_id=event.turn_id,
+                turn_revision=event.turn_revision,
             )
             st.mark_response_pending(request.response_key)
             queue.put(request)
@@ -718,9 +736,6 @@ class RealtimeService:
             st.speculative_user_turn_id = event.turn_id
             st.speculative_user_turn_revision = event.turn_revision
             st.speculative_user_speech_stopped_at_s = event.speech_stopped_at_s
-            tracker = self.turn_latency_store.get_or_create(event.turn_id, event.turn_revision)
-            if tracker is not None:
-                st.turn_latency = tracker
 
         queue = self.text_prompt_queue
         if queue:
@@ -731,6 +746,12 @@ class RealtimeService:
                 turn_id=event.turn_id,
                 turn_revision=event.turn_revision,
                 speech_stopped_at_s=event.speech_stopped_at_s,
+            )
+            self.bind_response_latency_tracker(
+                conn_id,
+                request.response_key,
+                turn_id=event.turn_id,
+                turn_revision=event.turn_revision,
             )
             st.mark_response_pending(request.response_key)
             queue.put(request)
--- a/src/speech_to_speech/api/openai_realtime/service.py
+++ b/src/speech_to_speech/api/openai_realtime/service.py
@@ -71,6 +71,7 @@ from speech_to_speech.pipeline.events import (
 from speech_to_speech.pipeline.messages import GenerateResponseRequest
 from speech_to_speech.pipeline.queue_types import TextPromptItem
 from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
+from speech_to_speech.pipeline.turn_latency import TurnLatencyStore, TurnLatencyTracker
 from speech_to_speech.pipeline.transcript_logging import log_exception, transcript_for_log
 from speech_to_speech.utils.utils import _generate_id
 
@@ -235,6 +236,7 @@ class ConnState(BaseModel):
     speculative_user_speech_stopped_at_s: Optional[float] = None
     speculative_user_item_id: Optional[str] = None
     speculative_audio_duration_s: float = 0.0
+    turn_latency: TurnLatencyTracker = Field(default_factory=TurnLatencyTracker)
     # Client conversation.item.create items that arrived while a response was
     # generating. Applying them mid-generation races the LLM handler's chat
     # write-back (cross-thread), so they are buffered here and flushed in order
@@ -298,12 +300,14 @@ class RealtimeService:
         should_listen: ThreadingEvent | None = None,
         chat_size: int = 10,
         speculative_turns: SpeculativeTurnTracker | None = None,
+        turn_latency_store: TurnLatencyStore | None = None,
         default_instructions: str | None = None,
     ) -> None:
         self.text_prompt_queue = text_prompt_queue
         self.should_listen = should_listen
         self._chat_size = chat_size
         self.speculative_turns = speculative_turns
+        self.turn_latency_store = turn_latency_store or TurnLatencyStore()
         self._default_instructions = default_instructions
         self._conns: dict[str, ConnState] = {}
         self.total_usage = GlobalUsageMetrics()
@@ -666,6 +670,9 @@ class RealtimeService:
             st.speculative_user_turn_id = event.turn_id
             st.speculative_user_turn_revision = event.turn_revision
             st.speculative_user_speech_stopped_at_s = event.speech_stopped_at_s
+            tracker = self.turn_latency_store.get_or_create(event.turn_id, event.turn_revision)
+            if tracker is not None:
+                st.turn_latency = tracker
 
         queue = self.text_prompt_queue
         if queue and transcript:
@@ -711,6 +718,9 @@ class RealtimeService:
             st.speculative_user_turn_id = event.turn_id
             st.speculative_user_turn_revision = event.turn_revision
             st.speculative_user_speech_stopped_at_s = event.speech_stopped_at_s
+            tracker = self.turn_latency_store.get_or_create(event.turn_id, event.turn_revision)
+            if tracker is not None:
+                st.turn_latency = tracker
 
         queue = self.text_prompt_queue
         if queue:
--- a/src/speech_to_speech/pipeline/turn_latency.py
+++ b/src/speech_to_speech/pipeline/turn_latency.py
@@ -1,0 +1,124 @@
+from __future__ import annotations
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass, field
+from threading import Lock
+from typing import Literal
+
+TurnLatencyStatus = Literal["completed", "cancelled", "failed", "incomplete"]
+
+_active_tracker: ContextVar[TurnLatencyTracker | None] = ContextVar(
+    "active_turn_latency_tracker",
+    default=None,
+)
+
+
+def active_turn_latency_tracker() -> TurnLatencyTracker | None:
+    return _active_tracker.get()
+
+
+@contextmanager
+def bind_active_turn_latency_tracker(tracker: TurnLatencyTracker | None):
+    token = _active_tracker.set(tracker)
+    try:
+        yield tracker
+    finally:
+        _active_tracker.reset(token)
+
+
+@dataclass
+class TurnLatencyTracker:
+    turn_id: str | None = None
+    turn_revision: int | None = None
+    stt_s: float | None = None
+    llm_s: float | None = None
+    tts_ttfa_s: float | None = None
+    e2e_s: float | None = None
+    mlx_lock_wait_s: float = 0.0
+    status: TurnLatencyStatus = "completed"
+
+    def reset(
+        self,
+        turn_id: str | None = None,
+        turn_revision: int | None = None,
+    ) -> None:
+        self.turn_id = turn_id
+        self.turn_revision = turn_revision
+        self.stt_s = None
+        self.llm_s = None
+        self.tts_ttfa_s = None
+        self.e2e_s = None
+        self.mlx_lock_wait_s = 0.0
+        self.status = "completed"
+
+    def record_stt(self, seconds: float) -> None:
+        self.stt_s = max(0.0, seconds)
+
+    def record_llm(self, seconds: float) -> None:
+        self.llm_s = max(0.0, seconds)
+
+    def record_tts_ttfa(self, seconds: float) -> None:
+        self.tts_ttfa_s = max(0.0, seconds)
+
+    def record_e2e(self, seconds: float) -> None:
+        self.e2e_s = max(0.0, seconds)
+
+    def record_mlx_lock_wait(self, seconds: float) -> None:
+        if seconds > 0.0:
+            self.mlx_lock_wait_s += max(0.0, seconds)
+
+    @staticmethod
+    def _fmt(seconds: float | None) -> str:
+        if seconds is None:
+            return "n/a"
+        return f"{seconds:.2f}s"
+
+    def format_log_line(self) -> str | None:
+        if self.turn_id is None:
+            return None
+        revision = 0 if self.turn_revision is None else self.turn_revision
+        return (
+            f"Turn {self.turn_id} rev={revision} latency: "
+            f"stt={self._fmt(self.stt_s)} llm={self._fmt(self.llm_s)} "
+            f"tts_ttfa={self._fmt(self.tts_ttfa_s)} e2e={self._fmt(self.e2e_s)} "
+            f"mlx_lock_wait={self.mlx_lock_wait_s:.2f}s status={self.status}"
+        )
+
+
+class TurnLatencyStore:
+    """Thread-safe latency trackers keyed by (turn_id, turn_revision)."""
+
+    def __init__(self) -> None:
+        self._lock = Lock()
+        self._trackers: dict[tuple[str, int], TurnLatencyTracker] = {}
+
+    @staticmethod
+    def _key(turn_id: str, turn_revision: int | None) -> tuple[str, int]:
+        return turn_id, 0 if turn_revision is None else turn_revision
+
+    def get_or_create(
+        self,
+        turn_id: str | None,
+        turn_revision: int | None,
+    ) -> TurnLatencyTracker | None:
+        if turn_id is None:
+            return None
+        key = self._key(turn_id, turn_revision)
+        with self._lock:
+            tracker = self._trackers.get(key)
+            if tracker is None:
+                tracker = TurnLatencyTracker(turn_id=turn_id, turn_revision=key[1])
+                self._trackers[key] = tracker
+            return tracker
+
+    def pop(
+        self,
+        turn_id: str | None,
+        turn_revision: int | None,
+    ) -> TurnLatencyTracker | None:
+        if turn_id is None:
+            return None
+        key = self._key(turn_id, turn_revision)
+        with self._lock:
+            return self._trackers.pop(key, None)
--- a/src/speech_to_speech/pipeline/turn_latency.py
+++ b/src/speech_to_speech/pipeline/turn_latency.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+from collections import defaultdict
 from contextlib import contextmanager
 from contextvars import ContextVar
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from threading import Lock
 from typing import Literal
 
@@ -59,14 +60,21 @@ class TurnLatencyTracker:
         self.llm_s = max(0.0, seconds)
 
     def record_tts_ttfa(self, seconds: float) -> None:
-        self.tts_ttfa_s = max(0.0, seconds)
+        if self.tts_ttfa_s is None:
+            self.tts_ttfa_s = max(0.0, seconds)
 
     def record_e2e(self, seconds: float) -> None:
-        self.e2e_s = max(0.0, seconds)
+        if self.e2e_s is None:
+            self.e2e_s = max(0.0, seconds)
 
     def record_mlx_lock_wait(self, seconds: float) -> None:
         if seconds > 0.0:
             self.mlx_lock_wait_s += max(0.0, seconds)
+
+    def absorb_pending(self, pending: TurnLatencyTracker) -> None:
+        if pending.stt_s is not None:
+            self.stt_s = pending.stt_s
+        self.mlx_lock_wait_s += pending.mlx_lock_wait_s
 
     @staticmethod
     def _fmt(seconds: float | None) -> str:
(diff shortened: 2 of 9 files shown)


andimarafioti:
This is really cool, and I would want to merge it, but it's a bit harder to do than it looks like at face value. Turns get cancelled, interruptions happen, tools can call other tools can continue, the assistant can speak, call a tool, use the tool to continue speaking, get interrupted at any point. You get where I'm going. I gave this to codex and it obviously found some gaps (it always will), but several point in this direction. Here's what it wrote:
Reviewed current head `4ab6a1e`; the PR still has no comments, reviews, or inline threads.

## 1. Verdict

**Request changes** — non-interrupting VAD paths can still finalize the wrong turn/revision, later TTS segments overwrite first-audio latency, unfinished trackers leak across sessions, and the required lint job fails.

## 2. Scope contract

Based on [[#516](https://github.com/huggingface/speech-to-speech/issues/516)](https://github.com/huggingface/speech-to-speech/issues/516), its discussion, [[PR #533](https://github.com/huggingface/speech-to-speech/pull/533)](https://github.com/huggingface/speech-to-speech/pull/533), and the linked follow-ups:

- **User-visible problem:** Diagnosing slow Mac/local turns requires manually correlating separate Parakeet STT, MLX-lock, LLM, Qwen3 TTS, and end-to-end logs.
- **Minimum behavior:** Emit one coherent terminal record for the correct turn/revision containing accurate STT, LLM, first TTS audio, end-to-end, MLX lock-wait, and terminal-status values.
- **Out of scope:** The already-merged bf16→4bit default change; language/template behavior in [[#534](https://github.com/huggingface/speech-to-speech/issues/534)](https://github.com/huggingface/speech-to-speech/issues/534); GPU/Whisper/Parler work in [[#107](https://github.com/huggingface/speech-to-speech/issues/107)](https://github.com/huggingface/speech-to-speech/issues/107) and [[#72](https://github.com/huggingface/speech-to-speech/issues/72)](https://github.com/huggingface/speech-to-speech/issues/72); conversation “feel”; instrumenting every backend; aggregate mean/max/p90 reporting.
- **Relevant protocol requirements:** `response.done` terminates one particular response. Speech/transcription can overlap with a response when interruption is disabled. Response, item, turn, and content identities are distinct; incremental transcript deltas must be accumulated, and completion events across turns cannot be correlated by arrival order.

## 3. Blocking findings

### High — A non-interrupting VAD turn can cause the active response to finalize the newer tracker

**Location:** [`[response.py:78–91](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/response.py#L78-L91)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/response.py#L78-L91), with the latest user identity replaced at [`[service.py:669–675](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L669-L675)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L669-L675) and [`[service.py:716–723](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L716-L723)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L716-L723).

The ordinary interrupting barge-in path is safe: [`[on_speech_started()](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/audio.py#L165-L204)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/audio.py#L165-L204) synchronously finalizes the old response before updating turn state.

The problem remains in two supported non-interrupting paths:

- The session may configure `interrupt_response=false`.
- VAD itself emits `SpeechStartedEvent(interrupt_response=False)` when speech is only confirmed at soft end ([`[vad_handler.py:726–736](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/VAD/vad_handler.py#L726-L736)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/VAD/vad_handler.py#L726-L736)).

**Concrete failure:** An old response remains active. Late-confirmed speech completes STT and replaces `speculative_user_turn_id/revision`. When the old response reaches `response.done`, `_log_turn_latency()` pops the newer tracker. For a reopened turn, this is revision 0 versus revision 1; for a committed turn, it can be turn 1 versus turn 2.

Official OpenAI documentation explicitly states that with `interrupt_response=false`, the ongoing response continues until completion. See the [[Realtime client-events reference](https://developers.openai.com/api/reference/resources/realtime/client-events)](https://developers.openai.com/api/reference/resources/realtime/client-events).

**Smallest correction:** Associate the originating turn/revision tracker with the existing `response_key` when the response request is created. At `_end_response`, finalize by the completed `current_response_key`, never by mutable latest-turn state.

**Within original scope:** Yes. Correct attribution is the central behavior of per-turn metrics.

### High — Later TTS segments overwrite the response’s actual first-audio latency

**Location:** [`[turn_latency.py:61–65](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L61-L65)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L61-L65), called from [`[qwen3_tts_handler.py:715–720](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/TTS/qwen3_tts_handler.py#L715-L720)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/TTS/qwen3_tts_handler.py#L715-L720) and [`[qwen3_tts_handler.py:876–884](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/TTS/qwen3_tts_handler.py#L876-L884)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/TTS/qwen3_tts_handler.py#L876-L884).

Each assistant text part can produce a separate `TTSInput` ([`[lm_output_processor.py:205–223](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/LLM/lm_output_processor.py#L205-L223)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/LLM/lm_output_processor.py#L205-L223)). Qwen coalesces items already queued, but a later sentence can arrive after the first synthesis has begun.

**Concrete failure:** The first segment records `tts_ttfa=0.10s` and `e2e=1.20s`. A later segment records `0.35s` and `5.00s`, overwriting both fields. The terminal log now describes the last segment’s first chunk rather than the response’s first audio.

**Smallest correction:** Make `record_tts_ttfa()` and `record_e2e()` first-write-wins within a tracker.

**Within original scope:** Yes. The advertised TTFA/e2e fields otherwise measure the wrong event.

### Medium — Unfinished trackers survive session teardown and can be reused

**Location:** [`[turn_latency.py:89–124](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L89-L124)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L89-L124) and [`[service.py:351–368](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L351-L368)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/service.py#L351-L368).

The store has no session namespace or lifecycle cleanup.

**Concrete failure:** A disconnect, empty final transcription, or shutdown before `response.done` leaves `(turn_1, 0)` in the store. VAD resets its counter when the pipeline session ends. The next session’s `turn_1` retrieves the same tracker and inherits stale stage values.

**Smallest correction:** Clear trackers after the pipeline unit’s `SESSION_END` has drained, or namespace them by session and remove that session’s entries during unregister.

**Within original scope:** Yes. Cross-session values make the diagnostic record incorrect.

### Medium — The required Ruff CI job fails

**Locations:** [`[response.py:78](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/response.py#L78)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/api/openai_realtime/handlers/response.py#L78), [`[turn_latency.py:5](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L5)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/src/speech_to_speech/pipeline/turn_latency.py#L5), and the changed import blocks.

`ruff check` reports six errors:

- Undefined `ConnState`.
- Unused `dataclasses.field`.
- Import-order errors in `language_model.py`, `qwen3_tts_handler.py`, `service.py`, and `s2s_pipeline.py`.

CI runs this command at [`[.github/workflows/ci.yml:17–20](https://github.com/huggingface/speech-to-speech/blob/main/.github/workflows/ci.yml#L17-L20)`](https://github.com/huggingface/speech-to-speech/blob/main/.github/workflows/ci.yml#L17-L20).

**Smallest correction:** Import `ConnState` under `TYPE_CHECKING`, remove `field`, and apply Ruff’s import ordering.

**Within original scope:** Yes, as required build hygiene.

## 4. Scope audit

### Required

- `TurnLatencyTracker` and temporary per-turn storage.
- Parakeet final-STT timing.
- LLM duration.
- Qwen3 TTFA and speech-stop-to-first-audio timing.
- Logging at terminal response status.
- Correct turn/revision identity and cleanup.

### Supporting but justified

- `ContextVar` binding for instrumentation inside the handlers and MLX lock.
- Lock protection around shared tracker storage.
- MLX lock-wait accumulation.
- `n/a` formatting for stages that do not run.
- Pipeline wiring that shares the tracker store among the relevant handlers.

The same hooks can run outside the Mac configuration when those handler classes are selected. That is slightly broader than the description but introduces no separate feature or policy, so it is justified.

### Unnecessary or out of scope

- `ConnState.turn_latency` duplicates the store and serves as an identity-unsafe fallback. It should be removed when finalization becomes response-keyed.
- The unused `field` import is unnecessary.
- Do not add other TTS/STT backends, VAD percentile aggregation, model-default changes, or conversation behavior to this PR.

## 5. OpenAI Realtime protocol assessment

The PR does not alter public event schemas, item IDs, content indexes, transcript payloads, or audio-truncation behavior.

The problematic assumption is internal: it treats the connection’s latest user turn as the identity of the response being finalized. Official OpenAI documentation says:

- `response.done` is emitted when that response is completely finished and contains its final result. [[Realtime conversations](https://developers.openai.com/api/docs/guides/realtime-conversations)](https://developers.openai.com/api/docs/guides/realtime-conversations)
- With `interrupt_response=false`, an ongoing response continues until complete while VAD events may still occur. [[Realtime client events](https://developers.openai.com/api/reference/resources/realtime/client-events)](https://developers.openai.com/api/reference/resources/realtime/client-events)
- Transcript deltas contain newly available text, and completion ordering across different speech turns is not guaranteed; `item_id` must be used for correlation. [[Realtime transcription](https://developers.openai.com/api/docs/guides/realtime-transcription)](https://developers.openai.com/api/docs/guides/realtime-transcription)

The instrumentation therefore needs stable response-to-turn correlation even though it does not expose that mapping in protocol events.

The `e2e` field measures server-side speech-stop-to-generated-first-chunk latency, not client playback latency. That is appropriate for this server diagnostic, but later chunks must not overwrite it.

## 6. Test coverage assessment

Focused tests pass: **230 passed** across the new tracker tests, TTS coalescing tests, and Realtime service tests. Formatting passes. Ruff fails with the six errors above.

The new tests establish formatting, storage, reset, and an aligned happy path. They do not verify the complete stage-to-response flow. In particular, [`[test_turn_latency_log_emission](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/tests/test_turn_latency.py#L66-L87)`](https://github.com/huggingface/speech-to-speech/blob/4ab6a1e37b698134a1cb9f5cbb2ffd6dee5eb350/tests/test_turn_latency.py#L66-L87) manually repeats the logging logic instead of exercising the handler.

Before approval, add tests for:

- An active response followed by `SpeechStartedEvent(interrupt_response=False)` and completion of the newer turn; the old response must log only its own tracker.
- The same scenario using a reopened turn with a higher revision.
- Multiple `TTSInput` segments retaining the first TTFA/e2e values.
- Session teardown with an unfinished tracker, followed by a clean new `turn_1`.
- At least one failed or cancelled terminal status, as claimed in the PR test plan.

No tests are needed for the explicitly out-of-scope follow-up issues.

## 7. Optional follow-ups

None. Keep language/template behavior, conversation feel, additional backends, and aggregate percentile metrics outside this PR.

salignatmoandal:
Thanks, @andimarafioti — the review was spot on. I've pushed fixes on the same branch:

1. _Response-keyed correlation (high)_
Trackers are created when a GenerateResponseRequest is bound (bind_response_latency_tracker) and popped by `current_response_key` in  end response. STT runs before a `response_key` exists, so it writes to a short-lived pending (turn_id, rev) slot that gets merged when the response tracker is created. Removed the `ConnState.turn_latency` fallback so there's a single source of truth.

2. _First-write-wins TTFA / e2e (high)_
`record_tts_ttfa` / `record_e2e` are only set when still None, so later TTS segments don't overwrite the first audible output.

3. _Session teardown (medium)_
`TurnLatencyStore.clear_session(conn_id)` is called from unregister. Response trackers are indexed per session; pending turn slots are cleared when the last tracked session is removed (pipeline units are typically one active session at a time).

4. _CI / hygiene_
Ruff import ordering + ConnState under TYPE_CHECKING in `response.py`.

Tests : 
`tests/test_turn_latency.py` — first-write-wins, pending STT merge, pop/clear_session, and multi-session pending behavior.

I haven't added full real-time integration tests yet (interrupt off + overlapping response, multi-TTS segments, failed/cancelled status). If you'd rather merge this observability layer first and follow up with those, that works for me — otherwise I can add them in a follow-up commit on this PR.

Does this approach look right to you?

tavallaie (commented):
I looked at af8fefd against the four issues from the first review. Response-key binding and first-write-wins for TTFA/e2e look right. Two things still break the log: pending barge-in never discards the tracker, and the tests do not go through `_end_response`.

src/speech_to_speech/api/openai_realtime/service.py
```diff
@@ -366,6 +370,21 @@ def unregister(self, conn_id: str) -> None:
...
 
+    def bind_response_latency_tracker(
+        self,
+        conn_id: str,
+        response_key: str,
+        *,
+        turn_id: str | None = None,
+        turn_revision: int | None = None,
+    ) -> None:
+        self.turn_latency_store.get_or_create_response(
```
tavallaie, line 381: This bind creates a tracker before the response opens. `close_pending_responses` tombstones the key and leaves that tracker. If speech starts after STT has queued the LLM request, the turn never logs. Discard each pending `response_key` there, or pop it and log `status=cancelled`.

tests/test_turn_latency.py
```diff
@@ -0,0 +1,133 @@
...
+    assert store.active_session_count == 1
+
+    assert store.pop("resp_b", session_id="sess_2") is not None
+    assert store.active_session_count == 0
+
+    store.clear_session("sess_2")
+    assert store.get_or_create_for_turn("turn_9", 0).stt_s is None
+
+
+def test_turn_latency_log_emission(caplog: pytest.LogCaptureFixture) -> None:
```
tavallaie, line 112: This writes the log line by hand. It does not call `_end_response`. The first commit keyed by `turn_id` and this test still passed. Drive the INFO line through `finish_response`, and add a case that barges in while `response_pending` is true.

src/speech_to_speech/s2s_pipeline.py
```diff
@@ -47,6 +47,7 @@
     VADOutItem,
 )
 from speech_to_speech.pipeline.speculative_turns import SpeculativeTurnTracker
+from speech_to_speech.pipeline.turn_latency import TurnLatencyStore
```
tavallaie, line 50: Ruff I001: `turn_latency` is imported before `transcript_logging`. CI will fail this.

src/speech_to_speech/pipeline/turn_latency.py
```diff
@@ -0,0 +1,194 @@
...
+        with self._lock:
+            tracker = self._trackers.get(response_key)
+            if tracker is None:
+                revision = None if turn_revision is None else turn_revision
+                tracker = TurnLatencyTracker(turn_id=turn_id, turn_revision=revision)
+                self._trackers[response_key] = tracker
+                if session_id is not None:
+                    self._session_keys[session_id].add(response_key)
+                if turn_id is not None:
+                    pending = self._pending_turn.pop(self._turn_key(turn_id, turn_revision), None)
```
tavallaie, line 166: Merge copies pending STT into a new object. If bind happens before `record_stt`, the write lands on the discarded tracker and the log shows `stt=n/a`. Keep the pending tracker as the response tracker.

brijmohan:
Hey - this seems interesting and right where I am working right now,

Ran this at `af8fefd` on my Mac M3 Pro with `--mac-optimal-settings`, with a local tool that
sleeps a fixed 2s so tool turns stand out in the timings. Adding an observed case to
@tavallaie's review rather than repeating it.

A tool turn emits two lines under the same key:

```
01:46:06,235  Turn turn_4 rev=1 latency: stt=0.83s llm=2.15s tts_ttfa=n/a   e2e=n/a   ... status=completed
01:46:11,726  Turn turn_4 rev=1 latency: stt=n/a   llm=1.76s tts_ttfa=0.15s e2e=6.90s ... status=completed
```

First is the tool call, second is the spoken answer once the result comes back. Same on
`turn_6 rev=0`.

Two things I think this adds:

`(turn_id, rev)` is no longer unique. Two records share it with nothing to order or group
them, so a consumer can't tell which is which, or that they belong to the same user turn.

The `stt=n/a` here is structural rather than the race in @tavallaie's note. Response 1 pops
`_pending_turn` legitimately, so response 2 has nothing to absorb. Keeping the pending
tracker as the response tracker fixes the race, but response 2 of a tool turn would still
show `n/a`, since the turn has one STT and response 1 owns it.

For what it's worth, the `e2e` is right: 6.90s does cover the full wait including the 2s
tool sleep.

Since the store already keys on `response_key`, carrying it onto the tracker and printing
it would make the lines addressable without changing anything else.

Separately, the 2s of tool time isn't attributed anywhere. It's inside `e2e` but has no
field, so a slow tool and a slow model look identical. `mlx_lock_wait` looks like the
precedent if you wanted a `tool_wait`.

One other thing: the branch is 46 commits behind main (branched at `3986f45`), and #545
touches the response/content part lifecycle this hooks into. Might be worth a rebase before
more review, since some of the above may shift.

If any of it's worth picking up, I'm happy to take one, smallest first:

1. A regression test for the tool-turn case that drives the line through `_end_response`,
   which also covers @tavallaie's point about the current test writing the log by hand. No
   behaviour change.
2. Carrying `response_key` onto the tracker and emitting it. One field, additive, nothing
   downstream changes meaning.
3. A `tool_wait` field, if the server can see the gap between emitting the `function_call`
   and the `function_call_output` coming back. I haven't checked that it can.

Equally happy to just leave the repro here and let you take it. The tool module is about
40 lines if it's useful.

Cheers !

andimarafioti:
FWIW, codex still has many complaints. I still think this is necessary and would be great to have it, but it's a PIA to implement. I can definitely take over if you want:
<html><head></head><body><p><strong>1. Verdict: Request changes.</strong> At <code dir="ltr">af8fefd</code>, response attribution and first-audio timing are improved, but cancellation cleanup, failed-generation timing, ambiguous response records, and failing CI still need correction.</p><p><strong>2. Scope contract</strong></p><p>Based on <a href="https://github.com/huggingface/speech-to-speech/issues/516"><span><span>issue #516</span></span></a>, its discussion, and the PR discussions:</p><ul><li><strong>Problem:</strong> Diagnosing Mac turn latency requires correlating separate stage logs.</li><li><strong>Minimum behavior:</strong> Emit accurate, identifiable terminal records for the Mac pipeline’s STT, LLM, first TTS audio, end-to-end latency, and lock waits; clean up abandoned measurements.</li><li><strong>Non-goals:</strong> The model-default change already handled by #533; language/template work in #534; Whisper/Parler work in #107/#72; content-part lifecycle work in #545; additional backends, percentile aggregation, conversation tuning, and tool-duration instrumentation.</li><li><strong>Protocol requirements:</strong> Correlate measurements with the particular response being finalized. Preserve terminal statuses and support overlapping speech without confusing response identity. Transcript deltas are incremental; completion ordering across turns is not guaranteed.</li></ul><p><strong>3. Blocking findings</strong></p><p>All findings below are <strong>P2 / medium</strong> and within the original diagnostic scope.</p><ol start="1"><li><p><strong>Cancelled pending responses retain their trackers.</strong><br>Location: <a href="https://github.com/huggingface/speech-to-speech/blob/af8fefdc1367ec2e2b4c3193bf304b401288265a/src/speech_to_speech/api/openai_realtime/service.py#L695"><span><span>service.py:695–701</span></span></a>.</p><p>After transcription queues a response, speech can interrupt before its first output. The transport calls <code dir="ltr">close_pending_responses()</code> and flushes the request, but that method only tombstones the response key. The newly allocated tracker survives until session teardown. Repeated interruptions accumulate unreachable trackers during a long session.</p><p><strong>Verified:</strong> The response key was closed while its tracker remained in <code dir="ltr">_trackers</code>.</p><p><strong>Smallest correction:</strong> Discard the matching tracker when closing abandoned response keys. No new cancellation event or synthetic latency record is necessary.</p></li><li><p><strong>Failed LLM generation loses its elapsed duration.</strong><br>Location: <a href="https://github.com/huggingface/speech-to-speech/blob/af8fefdc1367ec2e2b4c3193bf304b401288265a/src/speech_to_speech/LLM/language_model.py#L705"><span><span>language_model.py:705–706</span></span></a>.</p><p><code dir="ltr">record_llm()</code> runs only when <code dir="ltr">_generate()</code> returns normally. If generation spends time working and then raises, the exception handler emits the terminal error without recording that time.</p><p><strong>Verified:</strong> A controlled generator failure after 50 ms produced a real terminal log containing <code dir="ltr">llm=n/a … status=failed</code>. The stage ran, but the diagnostic cannot report its duration.</p><p><strong>Smallest correction:</strong> Record elapsed generation time in a <code dir="ltr">finally</code> around the generation loop, before the failure terminal is yielded.</p></li><li><p><strong>Tool-follow-up records omit the identity that distinguishes them.</strong><br>Location: <a href="https://github.com/huggingface/speech-to-speech/blob/af8fefdc1367ec2e2b4c3193bf304b401288265a/src/speech_to_speech/api/openai_realtime/handlers/response.py#L83"><span><span>response.py:83–85</span></span></a>.</p><p>A tool-call response and its follow-up are separate response trackers with the same turn/revision. Both emit the same identifying prefix. A consumer cannot identify which response a record describes.</p><p><strong>Verified through the service’s tool-follow-up lifecycle:</strong> Two terminal records used <code dir="ltr">Turn turn_5 rev=0</code>; only the first carried STT timing.</p><p><strong>Smallest correction:</strong> Include the already-available <code dir="ltr">response_key</code> when logging. Keep turn/revision for grouping. Do not duplicate STT measurements or introduce whole-tool-chain aggregation.</p></li><li><p><strong>The added code fails Ruff and mypy checks.</strong><br>Locations: <a href="https://github.com/huggingface/speech-to-speech/blob/af8fefdc1367ec2e2b4c3193bf304b401288265a/src/speech_to_speech/s2s_pipeline.py#L50"><span><span>s2s_pipeline.py:50</span></span></a>, <a href="https://github.com/huggingface/speech-to-speech/blob/af8fefdc1367ec2e2b4c3193bf304b401288265a/src/speech_to_speech/api/openai_realtime/handlers/response.py#L82"><span><span>response.py:82</span></span></a>.</p><p>Ruff reports the misplaced <code dir="ltr">turn_latency</code> import. The <a href="https://github.com/huggingface/speech-to-speech/actions/runs/33550635925/job/106665063220"><span><span>mypy job</span></span></a> rejects assigning <code dir="ltr">_ResponseStatus</code>, which includes <code dir="ltr">in_progress</code>, to the terminal-only tracker status.</p><p><strong>Smallest correction:</strong> Order the import correctly and explicitly narrow the status to terminal values before assignment.</p></li></ol><p><strong>4. Scope audit</strong></p><div><div><div>
Classification | Changes
-- | --
Required | Stage timings, terminal logging, stable response binding, first-write-wins TTFA/e2e, and tracker cleanup.
Supporting but justified | Shared store wiring, pending STT measurements, context binding, store locking, session indexing, and speculative-prefetch discard handling.
Unnecessary | TurnLatencyTracker.reset() has no production caller; remove it and simplify its test to construct a fresh tracker. This is non-blocking.
Out-of-scope implementation | No substantive feature expansion found.

</div></div></div><p>I would <strong>not repeat the pending-STT-copy race finding</strong>: Parakeet records STT before yielding the transcript; the notifier and service create the response afterward. The normal pipeline does not demonstrate that race.</p><p><strong>5. OpenAI Realtime protocol assessment</strong></p><p>No new wire-protocol violation was demonstrated.</p><ul><li>Response-key finalization now matches the response-specific meaning and terminal statuses of <a href="https://developers.openai.com/api/reference/resources/realtime/server-events#response.done"><span><span><code dir="ltr">response.done</code></span></span></a>.</li><li>I verified that a newer, non-interrupting speech turn does not cause the older response to consume its tracker. This respects <a href="https://developers.openai.com/api/reference/resources/realtime/client-events#session.update"><span><span><code dir="ltr">interrupt_response=false</code></span></span></a>.</li><li>The diff leaves item IDs, content indexes, transcript payloads, and transcript ordering logic unchanged. Official documentation requires incremental delta handling and item-based correlation across potentially unordered transcription completions. <a href="https://developers.openai.com/api/docs/guides/realtime-transcription#handle-transcript-events"><span><span>Realtime transcription</span></span></a></li></ul><p>The duplicate tool-turn prefix is an internal observability problem, not an invalid protocol event. Also, <code dir="ltr">e2e</code> measures first generated audio, not browser playback; full-generation <code dir="ltr">llm</code> time is not an additive component of TTFA.</p><p><strong>6. Test coverage assessment</strong></p><ul><li><strong>244 focused tests passed</strong>, including all seven new tracker tests.</li><li><strong>75 WebSocket/WebRTC tests passed</strong> after rerunning socket-dependent checks outside the restrictive sandbox.</li><li>Ruff fails as described; CI confirms the mypy failure.</li><li>No live Mac model benchmark was performed.</li></ul><p>The new tests cover storage and formatting, but the log-emission test manually writes the line instead of exercising <code dir="ltr">_end_response</code>. Add focused service-path assertions for pending cancellation cleanup, failed-generation duration, overlapping turn/revision attribution, and distinguishable tool-follow-up records. Verify multiple TTS segments through the handler as well as the tracker.</p><p><strong>7. Optional follow-ups — non-blocking</strong></p><p>Remove the unused reset method described above. No additional features or refactoring are needed for this PR.</p><p>Review only: no source edits, commits, pushes, issues, or posted review comments.</p></body></html>

Deep-unlearning:
@andimarafioti I picked up the remaining items from your [latest review](https://github.com/huggingface/speech-to-speech/pull/539#issuecomment-5773437847) and prepared seven follow-up commits at [`fb9673c`](https://github.com/Deep-unlearning/speech-to-speech/commit/fb9673c5a3e7d0a281b2d9c866c3ca4e7b90c8e6).

The changes are on [`Deep-unlearning:fix/turn-latency-lifecycle`](https://github.com/Deep-unlearning/speech-to-speech/tree/fix/turn-latency-lifecycle). My account has no write access to this PR's source branch, so **this PR has not been updated yet**. @salignatmoandal or a maintainer with access can bring this branch into `feat/turn-latency-logging` with a normal merge. It preserves the original PR commits and includes a merge of main at `ca5c33c`; no history was rewritten.

The follow-ups address:

- **Cancellation cleanup:** discard unfinished trackers whenever response keys close, including stale speculative responses. LLM/TTS workers only look up existing trackers, so a late cancelled worker cannot recreate an orphaned tracker.
- **Failed generation timing:** record elapsed LLM time before yielding the failure terminal, including failures after partial output.
- **Tool follow-up identity:** include `response_key` in each terminal latency record. The tool call and spoken follow-up remain distinct; STT is not duplicated and no tool-duration/aggregation feature was added.
- **Ruff/mypy:** fix import ordering and narrow the logging status to terminal statuses.
- **Integration coverage:** exercise the actual service/response, WebSocket, LLM, and Qwen TTS paths for terminal statuses, cancellation races, overlapping turns/revisions, session cleanup/reuse, tool follow-ups, and first-audio timing across multiple TTS segments.

Validation at `fb9673c`: **1,707 tests passed, 3 skipped**; Ruff, formatting, and mypy passed. All six pinned Agents SDK adapter/WebSocket/WebRTC checks passed.

I also ran a **live MacBook smoke test** with `--mac-optimal-settings` (Parakeet TDT, local MLX LLM, Qwen3 TTS), including a local tool that waits two seconds before returning `violet-42`.

[**log.txt — server/client smoke-test extract**](https://gist.github.com/Deep-unlearning/77c9a70bc38def66508093601e30cd67#file-log-txt)

The recording shows pending and active interruptions recovering, completed short replies, the tool result spoken back, and clean session unregister. There are seven terminal latency records with seven distinct response keys. The tool call/follow-up share `turn_6 rev=0` but have different response keys; the follow-up correctly has `stt=n/a`. No server ERROR or traceback appears. The extract retains MLX contention warnings and an initial client connection refusal followed by a successful connection; local username paths are redacted.

Live validation is partial: reconnecting to the same server and letting a longer multi-sentence reply finish uninterrupted are not captured. Tracker cleanup and multi-segment timing are covered by the automated regressions; the live logs alone do not prove those properties. The runtime log does not embed a Git SHA, so the exact live revision cannot be verified from the log alone.

salignatmoandal:
@Deep-unlearning thanks — I brought `fix/turn-latency-lifecycle` into `feat/turn-latency-logging` with a fast-forward to [`fb9673c`](https://github.com/salignatmoandal/speech-to-speech/commit/fb9673c5a3e7d0a281b2d9c866c3ca4e7b90c8e6). The original PR commits stay as they were, including the merge of `main` at `ca5c33c`. No history was rewritten.

Locally on that commit: the turn-latency tests (`tests/test_turn_latency.py` and `tests/openai_realtime/test_turn_latency.py`) passed, 20 tests, and `mypy src/` reported no issues. I have not re-run the full 1,707-test suite.

`main` has moved three commits past `ca5c33c` since then. Happy to pick that up if you want it in this PR before another look.

PR #542, opened 2026-09-02
Improve Apple Silicon first-time setup

dusterbloom:
## Summary

- add a pinned one-line Apple Silicon installer with native arm64 environment checks
- add guided STT/LLM/TTS setup with disk and cache checks, OpenAI-compatible endpoint discovery, Keychain-backed authentication, and small curated defaults
- add saved profiles, managed llama.cpp support for Gemma GGUF, and a `doctor` command
- add a post-publish macOS installer smoke test

## Testing

- `PYTHONPATH=src ../../.venv/bin/python -m pytest tests/ -q` — 1414 passed, 1 skipped
- `ruff check src/ tests/`
- `mypy src/`
- `/bin/sh -n scripts/install-macos.sh`
- `git diff --check`

## Release prerequisite

The README installer command targets `v0.2.13` and becomes available after that tag and PyPI release are published. The publish workflow verifies the real installer after publication.


andimarafioti (changes requested):
This is a bit over the top. If you think that the first-time Apple silicon setup is hard, describe exactly what the issues are and let's try to address those in the most minimal way possible. Instead, this feels like you start from an abstract problem "how do we make this simpler?" and end up with a super complex solution.

shadymuwa (approved):

andimarafioti:
Closing since there was no follow up

PR #544, opened 2026-09-03
Add Qwen3-ASR STT backend

### LINKED ISSUE #337: Add Qwen3-ASR support as an STT backend

dcrey7:
Closes #337. Supersedes #382 by @poorvajasathasivam: same idea, rebased on the new backend registry. Credit to them for the first version.

## What

+ Backend - `--stt qwen3-asr` runs the `Qwen/Qwen3-ASR-*-hf` checkpoints through Transformers (`AutoModelForMultimodalLM`, `apply_transcription_request`).
+ Partials - progressive mode yields partial transcriptions, like the other local STT backends.
+ Language - `auto` reports the model's own language ID per final turn, with the `-auto` suffix like faster-whisper. Short partial windows reuse the last final language, because the model's ID flips on 1 s clips.
+ Defaults - `auto` device picks CUDA, then MPS, then CPU. `auto` dtype picks bfloat16 where the GPU supports it, else float16, and float32 on CPU.
+ Flags - `--qwen3_asr_model_name`, `--qwen3_asr_device`, `--qwen3_asr_torch_dtype`, `--qwen3_asr_language`, `--qwen3_asr_prompt`, `--qwen3_asr_gen_max_new_tokens`.
+ Deps - none new. `pyproject.toml` already pins a Transformers that knows `qwen3_asr`. On 5.14.1 (the macOS pin) the processor has no `prompt` parameter, so the handler drops it with a warning there.
+ Language table - comes from the Transformers processor, so there is no copy to keep in sync. `fil` (Filipino) added to the LLM language map.

## Tested

+ Tests - 38 new. `uv run pytest`: 1411 passed. `ruff check`, `ruff format --check`, `mypy src`: clean.
+ Linux rig - RTX 3090 24 GB (driver 580), Ryzen 7 7700, Python 3.13, torch 2.11 cu130, Transformers 5.16.1. 0.6B in bfloat16: setup 4.5 s, 1.6 GB VRAM, 4 s turn 227 ms, 15 s turn 522 ms, 1 s partial 52 ms.
+ Mac - Apple M5 Pro 64 GB, macOS 26.4.1, Python 3.11, torch 2.11, Transformers 5.14.1. 0.6B on MPS in float16: same transcripts, setup 4.4 s, 1.6 GB, 4 s turn 303 ms, 15 s turn 784 ms. The prompt is dropped with the warning as documented.

## Not in here

+ torch.compile - 2x faster once warm, but it recompiled on each new audio length (36 s stall). A follow-up could pad audio to fixed buckets.
+ Streaming - Qwen3-ASR supports it. Fits the streaming STT interface in #537 once it lands.

cc @andimarafioti @ebezzam

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ This installs the package in editable mode and makes the `speech-to-speech` CLI 
 | STT | [Lightning Whisper MLX](https://github.com/mustafaaljadery/lightning-whisper-mlx) | Apple Silicon | `whisper-mlx` |
 | STT | [MLX Audio Whisper](https://github.com/huggingface/mlx-audio) | Apple Silicon | built-in on macOS |
 | STT | [Paraformer](https://github.com/modelscope/FunASR) | CUDA / CPU | `paraformer` |
+| STT | [Qwen3-ASR](https://huggingface.co/Qwen/Qwen3-ASR-0.6B-hf) through Transformers | CUDA / CPU, Apple Silicon | built-in |
 | STT | OpenAI-compatible `/v1/audio/transcriptions` endpoint | local or remote HTTP server | built-in |
 | LLM | OpenAI-compatible API (`responses-api`, `chat-completions`) | hosted providers or self-hosted servers | built-in |
 | LLM | [Transformers](https://huggingface.co/models?pipeline_tag=text-generation&sort=trending) | CUDA / CPU | built-in |
--- a/src/speech_to_speech/LLM/utils.py
+++ b/src/speech_to_speech/LLM/utils.py
@@ -208,6 +208,8 @@ WHISPER_LANGUAGE_TO_LLM_LANGUAGE = {
     "kk": "kazakh",
     "sq": "albanian",
     "sw": "swahili",
+    # Qwen3-ASR reports Filipino as "fil"; Whisper only knows "tl" (Tagalog) above.
+    "fil": "filipino",
     "gl": "galician",
     "mr": "marathi",
     "pa": "punjabi",
--- a/src/speech_to_speech/STT/README.md
+++ b/src/speech_to_speech/STT/README.md
@@ -89,7 +89,7 @@ This document summarizes the Speech-to-Text (STT) implementations in the `STT/` 
   - With `auto`, the model identifies the language of each final turn and reports it with the `-auto` suffix
   - Progressive (partial) windows are short and fool the language ID, so they reuse the language of the last final turn
   - A forced language is passed on every request and reported as is
-- Needs `transformers>=5.13.0`, which `pyproject.toml` already requires
+- Transformers versions: `pyproject.toml` already requires a version that knows `qwen3_asr`. The prompt and true language forcing need `transformers>=5.15.1` (the Linux pin); with 5.14.1 (the macOS pin) the language is a hint and the prompt is ignored with a warning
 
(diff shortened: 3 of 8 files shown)


dcrey7:
CI note: the macOS install smoke failed in the pip step, not in the tests. The runner hit a PyPI read timeout (`ReadTimeoutError ... files.pythonhosted.org` while upgrading pip), then pip reported `pydantic-core` as having no matching distribution, which is the symptom of a failed index fetch. This PR adds no dependencies. The same job passed on main on 2 Sep and on #543 this morning.

I ran the exact job steps (`uv build`, fresh Python 3.11 venv, `pip install dist/*.whl`, `pip check`, `tests/install_smoke.py`) on an Apple M5 Pro: all pass.

Could a maintainer re-run the macOS job? Thanks.

dcrey7:
@andimarafioti when you have a minute: the macOS install smoke failure is a PyPI read timeout on the runner (`ReadTimeoutError files.pythonhosted.org` during the pip step, details one comment up), not a dependency change from this PR. The same job passed on #543 this morning, and the exact steps pass on an Apple M5 Pro. A re-run of that job should clear it. Sorry for the noise, and thanks for approving the workflow.

dcrey7:
@andimarafioti proof on the same runner image: I ran this exact CI workflow on my fork, and all 10 jobs pass, including `Install smoke (macos-arm64)` on macos-14-arm64: https://github.com/dcrey7/speech-to-speech/actions/runs/33770903480. The failure here was the runner's PyPI read timeout (the log has no `Downloading pydantic_core` line at all, only the conflict list). A re-run of the macOS job should go green.

andimarafioti (approved):
Really solid work! Thank you! I wanted this done for a while already 🤗

dcrey7:
its an honor 🥲! congratulations on the nvidia patnership!! 🤗
 
Sad The Office GIFz

PR #545, opened 2026-09-03
Emit Realtime content part lifecycle events

### LINKED ISSUE #457: Emit standard Realtime output-item and content-part lifecycle events
## Problem

The Realtime endpoint emits response deltas, per-modality done events, and `response.done`, but omits the standard per-item lifecycle events:

- `response.output_item.added`
- `response.content_part.added`
- `response.content_part.done`
- `response.output_item.done`

The [official Realtime text lifecycle](https://developers.openai.com/api/docs/guides/realtime-conversations#text-inputs-and-outputs) places these events around `response.output_text.delta` / `response.output_text.done` and before `response.done`. The [`response.output_item.done` reference](https://developers.openai.com/api/reference/resources/realtime/server-events#response.output_item.done) also specifies that it is emitted when a response is interrupted, incomplete, or cancelled.

This omission predates #453. Ordered multi-item responses make it more visible because a client waiting for item completion only discovers the final items inside `response.done`.

## Expected behavior

Emit the standard lifecycle for every assistant message or function-call output item while preserving the stable item IDs and response-wide `output_index` values already used by the endpoint.

## Acceptance criteria

- Emit `response.output_item.added` when each response output item begins.
- Emit `response.content_part.added` before streaming a message content part and `response.content_part.done` after its modality-specific done event.
- Emit `response.output_item.done` after every item completes, including items exposed before interruption, incomplete completion, or cancellation.
- Emit `response.done` only after all applicable item/content lifecycle events.
- Keep IDs and output indexes consistent across lifecycle events and `response.done.output`.
- Add WebSocket and WebRTC protocol-sequence contract tests covering text, `text → tool → text`, and cancellation.

dcrey7:
Closes #457. Follow-up to #544, same approach.

## What

+ Content parts - every assistant message now emits `response.content_part.added` right after its `response.output_item.added`, and `response.content_part.done` right after its modality done event (`response.output_text.done` or `response.output_audio_transcript.done`), then `response.output_item.done`. One shared open and one shared close in the response handler, both idempotent, so no item gets two of anything.
+ Cancel and errors - the close runs with item status `incomplete` on cancellation, failure, and interruption, before `response.done`.
+ Ordered tool calls - an ordered tool call closes the preceding message before its own item begins, like the reference stream. The early tool-call side channel keeps its existing order: the tool item goes out while the message audio is still playing, and the ordered copy closes the message later.
+ Usage - `response.created` and `response.done` now carry `usage.input_token_details` and `output_token_details` like the real API. The server does not count text and audio tokens apart, so all tokens go to `text_tokens` and the other fields are zero. Said in the code.
+ Function-call items - unchanged, no content parts, as in the reference.
+ Not in here - the `conversation.item.created` naming. Separate question, as discussed on the issue.

## Why

pipecat 1.3.0's Realtime client dies on the first reply from this server: its parser requires the usage detail objects, and it tracks parts through the content part events. With this change its parser accepts every event shape the server sends (checked against pipecat's `parse_server_event`).

## Tested

+ Tests - 6 new plus updated sequence expectations. `uv run pytest`: 1417 passed. `ruff check`, `ruff format --check`, `mypy src`: clean.
+ Contract tests - WebSocket and WebRTC cover text, text then tool then text, and cancellation, plus the openai SDK client sequence for tool calls.
+ Linux rig - RTX 3090, Python 3.13, Transformers 5.16.1.
+ Mac - Apple M5 Pro, Python 3.11, Transformers 5.14.1: Realtime suite 426 passed.
+ Fork CI - the upstream workflow on my fork, all 10 jobs green: https://github.com/dcrey7/speech-to-speech/actions/runs/33802217730

cc @andimarafioti


andimarafioti:
Interesting, how are you using pipecat here? Could you describe the setup?

Besides, codex found a couple of issues, these seem relevant:

## 3. Blocking findings

1. **High — Previously completed messages become incomplete in `response.done`.**  
   [[response.py:501](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/src/speech_to_speech/api/openai_realtime/handlers/response.py#L501-L509)](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/src/speech_to_speech/api/openai_realtime/handlers/response.py#L501-L509), [[response.py:686](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/src/speech_to_speech/api/openai_realtime/handlers/response.py#L686-L692)](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/src/speech_to_speech/api/openai_realtime/handlers/response.py#L686-L692)

   Concrete scenario: stream text, close it as `completed` when a tool begins, start another message, then cancel. The first message’s `response.output_item.done` says `completed`, but `_build_output_items()` assigns every message the response-wide `incomplete` status, so the same item appears as `incomplete` in `response.done.output`. I reproduced this directly.

   The official [`[response.output_item.done](https://developers.openai.com/api/reference/resources/realtime/server-events#response.output_item.done)`](https://developers.openai.com/api/reference/resources/realtime/server-events#response.output_item.done) event carries the terminal item, while [`[response.done](https://developers.openai.com/api/reference/resources/realtime/server-events#response.done)`](https://developers.openai.com/api/reference/resources/realtime/server-events#response.done) carries all generated output items; changing an already-terminal item’s status makes those representations contradictory.

   **Smallest correction:** record each message’s terminal status when `_finish_message_output()` closes it, and reuse that status when building `response.done`; derive status from the response only for messages still open at termination. Add the multi-item cancellation regression test.

   **Within original issue scope:** Yes.

4. **Medium — The WebSocket `text → tool → text` test does not contract-test the new lifecycle.**  
   [[test_websocket_router.py:1078](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/tests/openai_realtime/test_websocket_router.py#L1078-L1096)](https://github.com/huggingface/speech-to-speech/blob/42b77d8312e3217eabda4240b53c3e1363c2590b/tests/openai_realtime/test_websocket_router.py#L1078-L1096)

   This test checks transcript/audio indexes and final output but never asserts content-part or item lifecycle ordering. Omitting `content_part.added`, `content_part.done`, or either message’s `output_item.done` would still pass, so the WebSocket half of the explicit acceptance criterion is not demonstrated.

   **Smallest correction:** assert the complete response-event sequence and stable IDs/indexes in this existing test, matching the WebRTC sequence assertion.

   **Within original issue scope:** Yes.

dcrey7:
Thanks, both right. Fixed in 2d65215:

+ Status - each message keeps the status it closed with, and `response.done` reuses it. Only messages still open when the response ends take the response status. One old expectation moved with it: a message closed as completed before a cancel now stays completed in `response.done`, which was the contradiction you pointed at.
+ Tests - added the multi-item cancellation test, and the WebSocket text, tool, text test now checks the full event order with stable ids and indexes, like the WebRTC one. Fork CI green on all 10 jobs: https://github.com/dcrey7/speech-to-speech/actions/runs/33858550546

On pipecat, the short version:

+ What I have been building recently - tranzlato, an agent that helps people speak in another language. The app runs on pipecat, because it gives me the browser and phone plumbing and easy access to the online voice APIs.
+ Why this server - I also want an offline mode, on a single 3090 or a MacBook. I have worked with this repo before, and it already has what that needs: turn detection, barge-in, one CLI, and it speaks the OpenAI Realtime protocol. So the plan is pipecat as the app and speech-to-speech as the local engine behind it.
+ How I hit this - pipecat ships an OpenAI Realtime client. You give it a `base_url` and it should just talk to this server. I took the events this server sends and ran them through pipecat's event parser. It rejected two: the empty usage token details, and it needs the content part events to follow each part. That is what this PR adds.
+ What is left - two small things on pipecat's side, not here: it expects `conversation.item.added` where this server sends `conversation.item.created`, and it assumes 24 kHz audio. About ten lines in the app.
+ Next - I will wire it end to end and share the setup here when it works, in case it helps as an example.

cc @andimarafioti

andimarafioti:
### P2 — Early tool delivery loses the preceding text-only message’s completion boundary

Location: [[response.py:1035–1036](https://github.com/huggingface/speech-to-speech/blob/2d652151f247881f4efbf70d8fb15778db2fcd95/src/speech_to_speech/api/openai_realtime/handlers/response.py#L1035-L1036)](https://github.com/huggingface/speech-to-speech/blob/2d652151f247881f4efbf70d8fb15778db2fcd95/src/speech_to_speech/api/openai_realtime/handlers/response.py#L1035-L1036), with the failed lookup at [[response.py:671–675](https://github.com/huggingface/speech-to-speech/blob/2d652151f247881f4efbf70d8fb15778db2fcd95/src/speech_to_speech/api/openai_realtime/handlers/response.py#L671-L675)](https://github.com/huggingface/speech-to-speech/blob/2d652151f247881f4efbf70d8fb15778db2fcd95/src/speech_to_speech/api/openai_realtime/handlers/response.py#L671-L675).

Concrete reproduction, confirmed through the WebSocket router:

1. Request text-only output.
2. Deliver the tool’s early notification, then ordered `text → tool → text`.
3. Cancel while the second message remains open.

Early tool exposure skips closing the first message and replaces `current_output_kind/index` with the tool. Text mode has no pending-audio identity, so the ordered tool copy cannot find the preceding message either.

Consequently, item 0 emits no text/content/item completion until cancellation. Final statuses become:

`incomplete, completed, incomplete`

The first message had already finished streaming. This preserves the original observable problem—clients cannot recognize its completion—and misclassifies it on cancellation. The official [`response.output_item.done` definition](https://developers.openai.com/api/reference/resources/realtime/server-events#response.output_item.done) ties that event to completion of item streaming.

**Smallest correction:** close the preceding text-only message when exposing the early tool; retain deferred closure only for audio awaiting TTS. Extend the existing early-tool regression test to text-only output followed by cancellation.

**Within original issue scope:** Yes.

andimarafioti (approved):
Great work! thank you 👍

dcrey7:
thank you very much its a pleasure! 

Alt Text

Issue #235, opened 2026-04-16
Feature Request: Great new VAD and ASR models


D3vShoaib:
I’d like to suggest adding support for a newer VAD and ASR stack.

[FireRedVAD](https://github.com/FireRedTeam/FireRedVAD) is multilingual, ~97.5% F1, streaming, theoretically better than Silero VAD.

[NVIDIA Parakeet](https://huggingface.co/nvidia/parakeet-unified-en-0.6b) is streaming, English only, better than parakeet v2/v3.

Would be useful to have these as optional backends alongside the current setup.

pehonnet:
Note that FireRedVAD is "theoretically" better than silero vad in non streaming mode. There are no benchmarks on the streaming version yet - see https://github.com/FireRedTeam/FireRedVAD/issues/12

andimarafioti:
there's a nemotron streaming as well!

PR #549, opened 2026-09-05
feat: add optional FireRed streaming VAD

tavallaie:
Ticket: #235

## Problem

Issue 235 asked for FireRedVAD as an optional detector next to Silero. Silero stays the default.

## Changes

- `--vad firered` with extra `fireredvad` and `--vad_firered_model_dir`.
- `FireRedVadIterator` uses FireRed `is_speech_start` / `is_speech_end`. It does not subclass Silero `VADIterator`.
- Incoming audio stays float32 in `[-1, 1]`. FireRed gets int16-scale windows and a 400/160 overlap tail.
- Active speech is counted in 10 ms hops. Trailing hangover silence is not counted.
- A 20 s FireRed split can emit end then start in one 512-sample chunk. The new start stays buffered. The ended hop count is kept for `VADHandler`.
- Setup docs use `hf download`. `huggingface-cli` is gone in `huggingface_hub` 1.0.

## Behaviour after the change

- `--vad silero` is unchanged.
- `--vad firered` fails at setup if the extra or `--vad_firered_model_dir` is missing.
- No JSON key change.

## Verification

```
uv run mypy src/
Success: no issues found in 104 source files

uv run ruff check src/ tests/
All checks passed!

171 files already formatted

uv run pytest tests/test_firered_vad_iterator.py tests/test_cli_defaults.py tests/test_backend_registry.py -q
115 passed
```

## Follow-up

Parakeet Unified STT is #553. Nemotron streaming STT is #554.

### DIFF
--- a/src/speech_to_speech/VAD/README.md
+++ b/src/speech_to_speech/VAD/README.md
@@ -1,0 +1,31 @@
+# VAD
+
+Turn detection still runs in `VADHandler`. The speech-probability source is selectable.
+
+## Backends (`--vad`)
+
+- `silero` (default) — Silero VAD via `VADIterator`
+- `firered` — FireRed streaming VAD, same trigger/silence/pad logic as Silero
+
+Streaming FireRedVAD has no published benchmark yet. Silero remains the default.
+
+## FireRedVAD
+
+Install the optional extra and download Stream-VAD weights:
+
+```bash
+pip install "speech-to-speech[fireredvad]"
+huggingface-cli download FireRedTeam/FireRedVAD --local-dir ./pretrained_models/FireRedVAD
+```
+
+Then:
+
+```bash
+speech-to-speech local \
+  --vad firered \
+  --vad_firered_model_dir ./pretrained_models/FireRedVAD/Stream-VAD
+```
+
+`--vad_firered_use_gpu` runs the detector on CUDA. Audio is 16 kHz mono PCM, same as Silero.
+
+Missing extra or `--vad_firered_model_dir` fails at handler setup, before a session is accepted.
--- a/src/speech_to_speech/VAD/firered_vad_iterator.py
+++ b/src/speech_to_speech/VAD/firered_vad_iterator.py
@@ -31,7 +31,8 @@ class FireRedProbModel:
 
     def __call__(self, x: torch.Tensor, sampling_rate: int) -> torch.Tensor:
         samples = x.detach().cpu().contiguous().view(-1).numpy().astype(np.float32, copy=False)
-        results = self.streamer.detect_chunk(samples)
+        # FireRed fbank trains on int16 PCM. VADIterator feeds Silero-scale [-1, 1].
+        results = self.streamer.detect_chunk(samples * 32768.0)
         if not results:
             return torch.tensor(0.0, dtype=torch.float32)
         return torch.tensor(float(results[-1].smoothed_prob), dtype=torch.float32)
--- a/src/speech_to_speech/VAD/firered_vad_iterator.py
+++ b/src/speech_to_speech/VAD/firered_vad_iterator.py
@@ -1,0 +1,70 @@
+"""FireRed streaming VAD as a drop-in speech-probability source for VADIterator."""
+
+from __future__ import annotations
+
+from typing import Protocol
+
+import numpy as np
+import torch
+
+from speech_to_speech.VAD.vad_iterator import VADIterator
+
+
+class FireRedFrameResult(Protocol):
+    smoothed_prob: float
+
+
+class FireRedStreamer(Protocol):
+    def reset(self) -> None: ...
+
+    def detect_chunk(self, audio_chunk: np.ndarray) -> list[FireRedFrameResult]: ...
+
+
+class FireRedProbModel:
+    """Adapt FireRedStreamVad.detect_chunk to the Silero model() call shape."""
+
+    def __init__(self, streamer: FireRedStreamer) -> None:
+        self.streamer = streamer
+
+    def reset_states(self) -> None:
+        self.streamer.reset()
+
+    def __call__(self, x: torch.Tensor, sampling_rate: int) -> torch.Tensor:
+        samples = x.detach().cpu().contiguous().view(-1).numpy().astype(np.float32, copy=False)
+        results = self.streamer.detect_chunk(samples)
+        if not results:
+            return torch.tensor(0.0, dtype=torch.float32)
+        return torch.tensor(float(results[-1].smoothed_prob), dtype=torch.float32)
+
+
+class FireRedVadIterator(VADIterator):
+    """Same trigger/silence/pad behaviour as Silero, with FireRed speech scores."""
+
+    def __init__(
+        self,
+        streamer: FireRedStreamer,
+        *,
+        threshold: float = 0.5,
+        sampling_rate: int = 16000,
+        min_silence_duration_ms: int = 300,
+        speech_pad_ms: int = 30,
+    ) -> None:
+        super().__init__(
+            FireRedProbModel(streamer),
+            threshold=threshold,
+            sampling_rate=sampling_rate,
+            min_silence_duration_ms=min_silence_duration_ms,
+            speech_pad_ms=speech_pad_ms,
+        )
+
+
+def load_firered_streamer(model_dir: str, *, use_gpu: bool = False) -> FireRedStreamer:
+    try:
+        from fireredvad import FireRedStreamVad, FireRedStreamVadConfig
+    except ImportError as exc:
+        raise ImportError(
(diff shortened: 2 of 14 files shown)


src/speech_to_speech/VAD/firered_vad_iterator.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, streamer: FireRedStreamer) -> None:
+        self.streamer = streamer
+
+    def reset_states(self) -> None:
+        self.streamer.reset()
+
+    def __call__(self, x: torch.Tensor, sampling_rate: int) -> torch.Tensor:
+        samples = x.detach().cpu().contiguous().view(-1).numpy().astype(np.float32, copy=False)
+        # FireRed fbank trains on int16 PCM. VADIterator feeds Silero-scale [-1, 1].
+        results = self.streamer.detect_chunk(samples * 32768.0)
```
andimarafioti: Passing each disjoint 512-sample chunk directly to `detect_chunk` loses the overlap FireRed needs. Its frontend uses 400-sample windows with a 160-sample step and restarts feature extraction on each call. For 3.2 seconds of audio, this produces 100 frames instead of 318; with the real model and its English sample, speech detection starts about 130 ms later and ends about 350 ms later than continuous framing. Please retain the unconsumed waveform tail, feed complete windows at the required step, and clear that buffer on reset. A test covering successive 512-sample chunks would catch this—the current mock always returns one score. [FireRed's streaming example](https://github.com/FireRedTeam/FireRedVAD/blob/main/fireredvad/bin/stream_vad.py) shows the required overlap.

src/speech_to_speech/VAD/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Streaming FireRedVAD has no published benchmark yet. Silero remains the default.
+
+## FireRedVAD
+
+Install the optional extra and download Stream-VAD weights:
+
+```bash
+pip install "speech-to-speech[fireredvad]"
+huggingface-cli download FireRedTeam/FireRedVAD --local-dir ./pretrained_models/FireRedVAD
```
andimarafioti: `huggingface-cli` was removed in `huggingface_hub` 1.0, so this command fails on a fresh installation with current dependencies before the user can download the required weights. Please change it to `hf download FireRedTeam/FireRedVAD --local-dir ./pretrained_models/FireRedVAD`. [Migration docs](https://huggingface.co/docs/huggingface_hub/main/concepts/migration#cli-cache-commands).

andimarafioti (changes requested):
There are a couple of issues to fix before merging: the FireRed adapter loses analysis frames between audio chunks, and the setup instructions use a removed command. Details inline.

src/speech_to_speech/VAD/README.md
tavallaie: Addressed in 0e1faf9: the VAD README now uses `hf download FireRedTeam/FireRedVAD --local-dir ./pretrained_models/FireRedVAD`.

src/speech_to_speech/VAD/firered_vad_iterator.py
tavallaie: Addressed in 0e1faf9: FireRedProbModel now concatenates the unconsumed hop-160 tail before each detect_chunk call, feeds complete 400-sample windows, and clears that tail on reset. A test of 100 successive 512-sample chunks now yields 318 frames.

src/speech_to_speech/VAD/firered_vad_iterator.py
```diff
@@ -0,0 +1,226 @@
...
+                self._pre_speech_buffer.clear()
+                self._pre_speech_samples = 0
+                self.buffer.append(x)
+                chunk_in_buffer = True
+                self.active_speech_samples = self._num_samples(x)
+                self.last_utterance_active_speech_samples = 0
+            if frame.is_speech_end and self.triggered:
+                if not chunk_in_buffer:
+                    self.buffer.append(x)
+                return self._end_utterance()
```
andimarafioti: This return skips any remaining frame events in the chunk. With FireRed's default 20-second limit, frame 2000 emits a speech end and frame 2001 emits a new start; both can arrive in the same 512-sample chunk. The start gets discarded, leaving the iterator inactive while the user continues speaking. With the real FireRed postprocessor and 22 seconds of continuous speech scores, only the first 20.032 seconds were emitted and the rest was unbuffered. Please process the remaining events and preserve the subsequent start/audio before returning, with a regression test for this boundary.

src/speech_to_speech/VAD/firered_vad_iterator.py
```diff
@@ -0,0 +1,226 @@
...
+        # detect_chunk restarts fbank, so feed complete 400-sample windows at a 160-sample hop.
+        n_frames = (len(audio) - _FIRERED_WINDOW_SAMPLES) // _FIRERED_HOP_SAMPLES + 1
+        chunk_end = _FIRERED_WINDOW_SAMPLES + (n_frames - 1) * _FIRERED_HOP_SAMPLES
+        results = self.streamer.detect_chunk(audio[:chunk_end])
+        self._tail = audio[n_frames * _FIRERED_HOP_SAMPLES :]
+        return results
+
+    def _append_chunk(self, x: torch.Tensor) -> None:
+        self.buffer.append(x)
+        self.active_speech_samples += self._num_samples(x)
```
andimarafioti: This counts every buffered chunk as active speech, including silence while FireRed waits to emit its end event. With a 500 ms silence threshold, 100 ms of speech scores becomes 480 ms of "active speech", passes the handler's 384 ms minimum, and emits an interrupting speech-start event followed by a final utterance. A brief sound can therefore interrupt the assistant during the silence that follows it. Please count speech-classified frame duration separately from buffered audio duration, and test that trailing silence cannot make a short sound pass the minimum.

andimarafioti (changes requested):
The previous issues are fixed, but there are two issues in the new FireRed event handling: continuing speech can be dropped at the 20-second split, and trailing silence is counted as active speech. Details inline.

andimarafioti:
I also think it would be simpler if this were a few different PRs for each thing you're trying to add :)

src/speech_to_speech/VAD/firered_vad_iterator.py
tavallaie: Addressed in f79a576: active_speech_samples now counts FireRed speech hops (160 samples each), not trailing silence in the PCM buffer. A 10 ms speech hop plus hangover silence still reports 160 active samples.

src/speech_to_speech/VAD/firered_vad_iterator.py
tavallaie: Addressed in f79a576 and 6f6bbc6: after is_speech_end the iterator keeps scanning the same 512-sample chunk. A following is_speech_start stays buffered and triggered. last_utterance_active_speech_samples keeps the ended hop count so VADHandler still sees the first utterance's active speech.

tavallaie:
> I also think it would be simpler if this were a few different PRs for each thing you're trying to add :)

because I have it in single issue I did it in single PR. let me break it to few PR.

tavallaie:
Split as requested. This PR is FireRed VAD only. Parakeet Unified and Nemotron STT are in https://github.com/huggingface/speech-to-speech/pull/553.

andimarafioti:
Codex' still complaining:

3. **Blocking finding — medium severity, within scope.**

   **Speech before start confirmation is omitted from the duration count** at [[firered_vad_iterator.py:196–198](https://github.com/huggingface/speech-to-speech/pull/549/files)](codex://review?pr=https%3A%2F%2Fgithub.com%2Fhuggingface%2Fspeech-to-speech%2Fpull%2F549&path=src%2Fspeech_to_speech%2FVAD%2Ffirered_vad_iterator.py&line=197&side=right).

   FireRed requires eight consecutive speech frames before confirming a start. The iterator begins counting only on that confirmation, losing the preceding seven speech frames—70 ms.

   Reproducing this through FireRed’s actual postprocessor and the existing handler produced **430 ms of classified speech, counted as 360 ms**. The default 384 ms minimum then discarded it, producing no transcription or speech lifecycle events.

   Preserve the candidate speech-frame count and include it when the start is confirmed, without counting padding or hangover silence. Add a regression demonstrating that this short utterance reaches transcription.

tavallaie:
@andimarafioti  Thanks for catching this. FireRed confirms start after eight speech frames, and we were only counting from that confirmation.

`4bc4688` keeps a candidate hop count while waiting for start and includes it when the start fires. 430 ms of speech is now 43 hops, not 36, so it still clears the 384 ms minimum. Pad and hangover silence stay out of the count.

andimarafioti (approved):
Good work, thank you! In the end codex kept on complaining so I just had it work on it instead of going back and forth, nothing big!

PR #553, opened 2026-09-08
feat: add Parakeet Unified STT

tavallaie:
Ticket: #235

## Problem

Issue 235 asked for NVIDIA Parakeet Unified as an optional STT backend.

## Changes

- `--stt parakeet-unified` default model `nvidia/parakeet-unified-en-0.6b`.
- Shared `NemoASRSTTHandler`. Optional extra `nemo`.
- Each VAD utterance is transcribed with NeMo `ASRModel.transcribe`. This does not add cache-aware chunked decode.

## Behaviour after the change

- `--stt parakeet-tdt` stays the default.
- Missing extra fails at setup and names `speech-to-speech[nemo]`.
- No JSON key change.

## Verification

```
uv run mypy src/
Success: no issues found in 108 source files

uv run ruff check src/ tests/
All checks passed!

176 files already formatted

uv run pytest tests/test_nemo_asr_handler.py tests/test_backend_registry.py tests/test_cli_defaults.py -q
117 passed
```

## Follow-up

Nemotron streaming STT is #554 (stacked on this branch). FireRed VAD is #549.


tavallaie:
Nemotron streaming STT is now a separate PR stacked on this branch.

tavallaie:
Nemotron streaming STT is https://github.com/huggingface/speech-to-speech/pull/554. Merge this PR first.

Hotragn:
Contributor review, not a maintainer — and thanks for the careful review on #556, that caught a real mistake. Returning the favour on the language-plumbing side, since that's where I've spent most of my time in this repo. Nothing here is blocking; two suggestions and one thing I checked and think you got right.

**1. `SUPPORTED_LANGUAGES` and the coverage sweep.**

`tests/test_llm_utils.py::test_every_stt_language_has_an_llm_language_name` walks each bundled backend's `SUPPORTED_LANGUAGES` and asserts every code has an entry in `WHISPER_LANGUAGE_TO_LLM_LANGUAGE`. A missing entry means `--enable_lang_prompt` emits *nothing* for that language — silently, since the instruction is gated on the resolved name rather than the code. That's the bug #395 fixed for 17 of Parakeet TDT's 25 languages.

`NemoASRSTTHandler` doesn't declare `SUPPORTED_LANGUAGES`, so it isn't in that sweep (the module list is explicit). For the `-en-` checkpoint that's harmless today, but declaring even `["en"]` and adding the module to `_STT_HANDLER_MODULES` would mean a future multilingual unified checkpoint can't quietly drift out of the name map. Whoever adds `nvidia/parakeet-unified-*` for other languages then gets a CI failure instead of a silent no-op.

**2. The language flag accepts anything.**

`parakeet_unified_language` is a free-form `str` reported verbatim:

```python
yield Transcription(..., language_code=self.language)
```

The help text is accurate that it's "the language code reported", so this isn't a bug. But nothing validates it, and every other STT backend here takes `--language auto`, so `--parakeet_unified_language auto` is a natural thing to try. It produces a bare `"auto"` code:

```
resolve_auto_language('auto') -> ('auto', None)
resolve_auto_language('xx')   -> ('xx', None)
```

`None` name means no language instruction, and the TTS maps (`WHISPER_LANGUAGE_TO_KOKORO_LANG`, `WHISPER_LANGUAGE_TO_FACEBOOK_LANGUAGE`) are keyed on bare codes, so they'd fall back too. A one-line warning at `setup` when the code isn't in `WHISPER_LANGUAGE_TO_LLM_LANGUAGE` would turn a silent misconfiguration into a visible one. Same class of thing you flagged on mine — `"auto"` is a request, not a code.

**3. What I checked and think is right.**

No `on_session_end` override is needed here: the handler keeps no per-conversation language state (`self.language` is config, not detection), so there's nothing to reset when a unit is reclaimed — unlike the handlers in #556. And no `MLXLockContext`, correctly, since NeMo runs through torch rather than MLX; the global lock only guards MLX's shared Metal queue.

One process note: #554 carries `parakeet_unified_stt_arguments.py` and the same `nemo_asr_handler.py`, so it looks stacked on this one. Might be worth saying so in its description, or landing this first, so a reviewer doesn't read the shared handler twice.

tavallaie:
@Hotragn thanks for the review. Your note on #556 about `auto` being a request and not a language code stuck with me, so the language-flag part here was an easy fix once I knew what to look for.

`63e6efc` is up on the branch.

`SUPPORTED_LANGUAGES = ["en"]` on `NemoASRSTTHandler`, and I added the module to `_STT_HANDLER_MODULES`. The LLM name sweep should catch a future multilingual unified checkpoint if we forget a map entry, instead of `--enable_lang_prompt` doing nothing with no error.

For `--parakeet_unified_language`, setup now keeps the raw value in `start_language` but never emits bare `"auto"` on the transcription. `auto` logs a warning and reports `en` for the current `-en-` model. Unknown codes still pass through, with a warning if they are not in `WHISPER_LANGUAGE_TO_LLM_LANGUAGE`. Tests for both.

Still no `on_session_end` override and no MLX lock on my side. Same reasoning as in your comment.

On #554, I will keep this PR first so we are not re-reading the same handler twice.

Appreciate you taking the time on this.

andimarafioti (approved):
Good work, thank you!

PR #554, opened 2026-09-08
feat: add Nemotron streaming STT

tavallaie:
Ticket: #235

## Problem

Nemotron streaming ASR uses the same NeMo `ASRModel` handler as Parakeet Unified in #553. Keep it as its own backend flag.

## Changes

- `--stt nemotron-streaming` default model `nvidia/nemotron-speech-streaming-en-0.6b`.
- Override `nvidia/nemotron-3.5-asr-streaming-0.6b` for multilingual.
- Reuses `NemoASRSTTHandler` and extra `nemo` from #553.

## Behaviour after the change

- `--stt parakeet-tdt` stays the default.
- Missing extra fails at setup and names `speech-to-speech[nemo]`.
- No JSON key change.

## Verification

```
uv run mypy src/
Success: no issues found in 109 source files

uv run ruff check src/ tests/
All checks passed!

178 files already formatted

uv run pytest tests/test_nemotron_streaming_stt.py tests/test_nemo_asr_handler.py tests/test_backend_registry.py tests/test_cli_defaults.py -q
120 passed
```

## Follow-up

This branch sits on #553. Merge #553 first. The Nemotron-only commit is `f733795`. FireRed VAD is in #549.


andimarafioti (changes requested):
Requesting changes on the multilingual support. The documentation now explains the fixed-language setting, but that doesn't address the underlying gap in the integration.

Nemotron 3.5 can detect the language of each utterance, but `NemoASRSTTHandler` always sets `Transcription.language_code` to the configured `self.language`. It never uses the model's detected language. For example, with the multilingual checkpoint and the default settings, a French transcript is still labelled `en`. With `--enable_lang_prompt`, that asks the LLM to reply in English, and language-sensitive TTS also receives English. Setting the flag to `fr` works for a French-only session, but the same problem returns when the speaker switches to another language.

NVIDIA documents automatic detection via `target_lang=auto`, with the detected language included as a tag in the output: https://huggingface.co/nvidia/nemotron-3.5-asr-streaming-0.6b#supported-languages

Before we call the multilingual integration ready, we should use the detected language for each final transcription, remove the language tag from the text passed downstream, and cover two successive utterances in different languages with a focused test. The English-only checkpoint should keep its existing behavior. This doesn't require adding cache-aware streaming or a separate language detector.

If we'd rather keep this PR small, I'm also happy to keep it explicitly English-only and remove the multilingual support claim and example for now. The fixed-language example alone isn't enough to call that part done.

tavallaie:
Thanks. Nemotron 3.5 now uses `target_lang=auto`, strips the `<xx-XX>` tag, and reports the detected code on each final turn. The English-only checkpoint is unchanged. Covered in `1332dad`.

PR #556, opened 2026-09-08
Fix(stt): reset the detected language when a session ends

### LINKED ISSUE #555: STT handlers retain the detected language across sessions; only Parakeet resets it
## Summary

In `--language auto`, four of the five bundled STT handlers keep `self.last_language` when a session ends. Pipeline units are reused across clients, so the next client inherits the previous client's detected language.

`ParakeetTDTSTTHandler` — the default backend — already resets it, which is what makes this look like an oversight rather than a policy:

```python
# STT/parakeet_tdt_handler.py
def on_session_end(self) -> None:
    super().on_session_end()
    self.last_language = self.start_language if self.start_language else "en"
```

`BaseSTTHandler.on_session_end` only clears revision bookkeeping:

```python
def on_session_end(self) -> None:
    if hasattr(self, "_completed_final_revision_keys"):
        self._completed_final_revision_keys.clear()
```

| handler | resets `last_language` |
|---|---|
| `parakeet_tdt_handler` (default) | yes |
| `whisper_stt_handler` | no |
| `mlx_audio_whisper_handler` | no |
| `lightning_whisper_mlx_handler` | no |
| `qwen3_asr_handler` | no |

`on_session_end` is the documented place for this — `websocket_router._clean_unit` describes `SESSION_END` as *"the soft reset signal for stateful handlers"*, and flushes the queues specifically so work "cannot be picked up by handlers and leak into the next session that claims this unit."

## Impact, per backend

I want to be precise rather than overstate this, because it varies:

**Qwen3-ASR — user-visible.** Progressive requests are forced into `last_language`:

```python
# STT/qwen3_asr_handler.py:148
request_language = self.forced_language
if request_language is None and progressive:
    request_language = self.last_language
```

So with `--enable_live_transcription`, a new client's live partials are transcribed in the *previous* client's language until their first final turn re-detects. The handler's own docstring says progressive windows reuse "the language of the last final turn" — across a session boundary that's the wrong conversation's last final turn.

**Whisper family — fallback only.** `last_language` is consulted when detection yields nothing, so the leak surfaces on the new session's first turn only if detection is unavailable or returns an unsupported code. Milder. Note `process()` copies `gen_kwargs` (`dict(self.gen_kwargs)`), so the forced-language kwarg is *not* mutated at runtime — I checked, and only `last_language` persists.

## Reproduction

No model needed; this is handler state, and it is what `SESSION_END` is supposed to clear:

```python
from speech_to_speech.STT.whisper_stt_handler import WhisperSTTHandler
from speech_to_speech.STT.parakeet_tdt_handler import ParakeetTDTSTTHandler

w = object.__new__(WhisperSTTHandler)
w.start_language = "auto"; w.last_language = "de"; w.gen_kwargs = {}
w.on_session_end()
print(w.last_language)   # 'de'  <- survives into the next session

p = object.__new__(ParakeetTDTSTTHandler)
p.start_language = "auto"; p.last_language = "de"; p.enable_live_transcription = False
p.on_session_end()
print(p.last_language)   # 'auto' <- reset
```

End to end: run `serve` with `--stt qwen3-asr --language auto --enable_live_transcription`, connect a German speaker, disconnect, then connect a French speaker to the same pipeline unit. The second client's live partials come back German-forced until their first final turn.

## Suggested fix

Reset the per-session language state where the convention already lives. Either:

1. Give each handler an `on_session_end` matching Parakeet's, or
2. Do it once in `BaseSTTHandler.on_session_end` — `start_language`/`last_language` is a shared convention across all five handlers, so one reset covers every current backend and any future one, which is the part I'd value most given this was missed four times.

Happy to open a PR; I'd default to (2) with the contract documented on the base class, unless you'd rather keep it explicit per handler.

One adjacent thing I noticed while checking, not part of this: Parakeet resets to `self.start_language`, which is the literal string `"auto"` in auto mode rather than a language code. That's a separate question from whether the reset happens.

## Environment

- `speech-to-speech` v1.0.0, `main` at `16d7f98`
- Affects the realtime server, where `PipelineUnit`s are claimed and released across clients

Hotragn:
Fixes #555

## Problem

Pipeline units are reused across clients, and `SESSION_END` is the documented *"soft reset signal for stateful handlers"* — `websocket_router._clean_unit` flushes the queues specifically so a released session's work *"cannot be picked up by handlers and leak into the next session that claims this unit."*

`last_language` is that kind of state, but only `ParakeetTDTSTTHandler` reset it. `BaseSTTHandler.on_session_end` cleared revision bookkeeping and nothing else:

| handler | reset `last_language` before |
|---|---|
| `parakeet_tdt_handler` (default) | yes |
| `whisper_stt_handler` | no |
| `mlx_audio_whisper_handler` | no |
| `lightning_whisper_mlx_handler` | no |
| `faster_whisper_handler` | no |
| `qwen3_asr_handler` | no |

**Most visible on Qwen3-ASR**, which forces progressive requests with it:

```python
request_language = self.forced_language
if request_language is None and progressive:
    request_language = self.last_language
```

With `--enable_live_transcription`, a new client's live partials were transcribed in the *previous* client's language until their first final turn re-detected. The handler's docstring says progressive windows reuse "the language of the last final turn" — across a session boundary that's the wrong conversation's.

**For the whisper family the leak is narrower**, and I want to be accurate about it rather than inflate the finding: `last_language` is only a fallback consulted when detection yields nothing, and `process()` copies `gen_kwargs` (`dict(self.gen_kwargs)`), so no forced-language kwarg persists. I checked that specifically — my first read of this was wrong.

## Fix

One reset on the base class rather than six per-handler copies:

```python
def reset_session_language(self) -> None:
    self.last_language = self.start_language

def on_session_end(self) -> None:
    if hasattr(self, "_completed_final_revision_keys"):
        self._completed_final_revision_keys.clear()
    self.reset_session_language()
```

`start_language` / `last_language` is a convention every backend already shares, and the reset was missed in five of six handlers — so a new backend inheriting the correct behaviour is worth more here than keeping it explicit at each site. Both are declared on the base class so the reset also works for a handler that never sets them. `reset_session_language()` is public so a handler can reuse it without overriding `on_session_end`.

**Parakeet is unaffected**: its override still runs after `super()`, keeping its own `"en"` default. There's a test pinning that so this doesn't silently change the default backend.

Source diff is 20 lines in one file.

## Tests

Every bundled backend is parametrized over both directions — a *detected* language must not survive a session, a *user-configured* one must — plus the base-class contract, safety for a handler that never set a language, and that the pre-existing revision-bookkeeping clear still happens. The optional `faster_whisper` / `lightning_whisper_mlx` imports are stubbed so all six backends stay checkable on CI rather than skipping.

Ten fail against current `main`; the headline case fails with `assert 'de' != 'de'`.

Full suite: **1634 passed, 3 skipped**. `ruff check`, `ruff format --check`, `mypy src/` clean.

## Note

Parakeet resets to `self.start_language`, which is the literal `"auto"` in auto mode rather than a language code. I left that alone — it's pre-existing, orthogonal to whether the reset happens, and flagged in #555 rather than folded in here.

### DIFF
--- a/src/speech_to_speech/STT/base_stt_handler.py
+++ b/src/speech_to_speech/STT/base_stt_handler.py
@@ -21,6 +21,13 @@ class BaseSTTHandler(BaseHandler[STTIn, STTOut]):
     speculative_turns: SpeculativeTurnTracker | None = None
     final_revision_settle_s: float = 0.0
 
+    # Shared language convention: `start_language` is what the user configured and lives for
+    # the process, `last_language` is per-conversation detection state that on_session_end
+    # clears. Declared here so the reset applies to every backend, including ones that never
+    # set them.
+    start_language: Any = None
+    last_language: Any = None
+
     def should_process_input(self, item: STTIn) -> bool:
         mode = getattr(item, "mode", None)
         turn_id = getattr(item, "turn_id", None)
@@ -208,6 +215,19 @@ class BaseSTTHandler(BaseHandler[STTIn, STTOut]):
         while len(completed) > self._MAX_COMPLETED_FINAL_REVISIONS:
             completed.popitem(last=False)
 
+    def reset_session_language(self) -> None:
+        """Drop the language detected for the finished conversation.
+
+        Pipeline units are reused across clients, so a detected language left behind is the
+        previous conversation's. Backends consult `last_language` when detection yields
+        nothing, and Qwen3-ASR forces progressive requests with it, so a new client would be
+        transcribed in the old client's language until their first final turn re-detects.
+        Resetting to `start_language` restores whatever the user configured -- including
+        `None`, which means "detect per turn".
+        """
+        self.last_language = self.start_language
+
     def on_session_end(self) -> None:
         if hasattr(self, "_completed_final_revision_keys"):
             self._completed_final_revision_keys.clear()
+        self.reset_session_language()
--- a/tests/test_stt_session_language_reset.py
+++ b/tests/test_stt_session_language_reset.py
@@ -1,0 +1,152 @@
+"""A finished conversation's detected language must not reach the next one.
+
+Pipeline units are reused across clients: `websocket_router._clean_unit` flushes the queues
+so pending work "cannot be picked up by handlers and leak into the next session that claims
+this unit", and `SESSION_END` is "the soft reset signal for stateful handlers".
+
+`last_language` is exactly that kind of state. Backends consult it when detection yields
+nothing, and Qwen3-ASR forces progressive requests with it, so leaving it set transcribes a
+new client in the previous client's language. Only Parakeet reset it; the reset now lives on
+`BaseSTTHandler` so it covers every backend, including future ones.
+"""
+
+from __future__ import annotations
+
+import importlib
+import sys
+import types
+from contextlib import ExitStack
+from unittest import mock
+
+import pytest
+
+from speech_to_speech.STT.base_stt_handler import BaseSTTHandler
+
+# (module, class, extra attributes process()/on_session_end() needs)
+_STT_HANDLERS = [
+    ("speech_to_speech.STT.parakeet_tdt_handler", "ParakeetTDTSTTHandler", {"enable_live_transcription": False}),
+    ("speech_to_speech.STT.whisper_stt_handler", "WhisperSTTHandler", {}),
+    ("speech_to_speech.STT.mlx_audio_whisper_handler", "MLXAudioWhisperSTTHandler", {}),
+    ("speech_to_speech.STT.lightning_whisper_mlx_handler", "LightningWhisperSTTHandler", {}),
+    ("speech_to_speech.STT.faster_whisper_handler", "FasterWhisperSTTHandler", {}),
+    ("speech_to_speech.STT.qwen3_asr_handler", "Qwen3ASRSTTHandler", {}),
+]
+
+# Optional third-party imports stubbed so every backend stays checkable without its extra.
+_STUBS = {
+    "speech_to_speech.STT.faster_whisper_handler": ("faster_whisper", "WhisperModel"),
+    "speech_to_speech.STT.lightning_whisper_mlx_handler": ("lightning_whisper_mlx", "LightningWhisperMLX"),
+}
+
+
+def _load(module_name: str, class_name: str):
+    stub = _STUBS.get(module_name)
+    with ExitStack() as stack:
+        if stub is not None and stub[0] not in sys.modules:
+            package, attribute = stub
+            module = types.ModuleType(package)
+            setattr(module, attribute, object)
+            stack.enter_context(mock.patch.dict(sys.modules, {package: module}))
+            stack.callback(sys.modules.pop, module_name, None)
+        try:
+            return getattr(importlib.import_module(module_name), class_name)
+        except ImportError:
+            return None
+
+
+def _handler(cls, extra, *, start_language, last_language):
+    handler = object.__new__(cls)
+    handler.start_language = start_language
+    handler.last_language = last_language
+    for key, value in extra.items():
+        setattr(handler, key, value)
+    return handler
+
+
+@pytest.mark.parametrize(("module_name", "class_name", "extra"), _STT_HANDLERS)
+def test_detected_language_does_not_survive_the_session(module_name, class_name, extra):
+    """In auto mode the next client must start from "detect", not the last detection."""
+    cls = _load(module_name, class_name)
+    if cls is None:
+        pytest.skip(f"{module_name} requires an optional dependency")
+
+    handler = _handler(cls, extra, start_language=None, last_language="de")
+
+    handler.on_session_end()
+
+    assert handler.last_language != "de", (
+        f"{class_name} carried the previous session's detected language into the next one"
+    )
+
+
+@pytest.mark.parametrize(("module_name", "class_name", "extra"), _STT_HANDLERS)
+def test_configured_language_survives_the_session(module_name, class_name, extra):
+    """A user-configured language is process-level config, not per-conversation state."""
+    cls = _load(module_name, class_name)
+    if cls is None:
+        pytest.skip(f"{module_name} requires an optional dependency")
+
+    handler = _handler(cls, extra, start_language="de", last_language="fr")
+
+    handler.on_session_end()
+
+    assert handler.last_language == "de"
+
+
+def test_parakeet_keeps_its_english_fallback():
+    """Parakeet already reset, with its own default. That behaviour is unchanged."""
+    cls = _load("speech_to_speech.STT.parakeet_tdt_handler", "ParakeetTDTSTTHandler")
+    handler = _handler(cls, {"enable_live_transcription": False}, start_language=None, last_language="de")
+
+    handler.on_session_end()
+
+    assert handler.last_language == "en"
+
+
+# --- the base-class contract --------------------------------------------------------------
+
+
+def test_base_handler_declares_the_language_attributes():
+    """Declared on the base so the reset works for a backend that never sets them."""
+    assert BaseSTTHandler.start_language is None
+    assert BaseSTTHandler.last_language is None
+
+
+def test_reset_is_safe_for_a_handler_that_never_set_a_language():
+    class Bare(BaseSTTHandler):
+        pass
+
+    handler = object.__new__(Bare)
+
+    handler.on_session_end()  # must not raise
+
+    assert handler.last_language is None
+
+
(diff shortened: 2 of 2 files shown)


src/speech_to_speech/STT/base_stt_handler.py
```diff
@@ -208,6 +215,19 @@ def _mark_completed_final_revision(self, output: Transcription) -> None:
...
+        """Drop the language detected for the finished conversation.
+
+        Pipeline units are reused across clients, so a detected language left behind is the
+        previous conversation's. Backends consult `last_language` when detection yields
+        nothing, and Qwen3-ASR forces progressive requests with it, so a new client would be
+        transcribed in the old client's language until their first final turn re-detects.
+        Resetting to `start_language` restores whatever the user configured -- including
+        `None`, which means "detect per turn".
+        """
+        self.last_language = self.start_language
```
tavallaie: **[bug]** `self.last_language = self.start_language` restores a detect-mode `"auto"` as if it were a language code. Whisper and MLX-audio-whisper initialize `last_language` with `language if language != "auto" else None`. After `on_session_end()` with `start_language="auto"`, `last_language` is `"auto"`. Whisper's detection-failed path then prefers `"auto"` over English and reports `"auto-auto"`. The new tests use `start_language=None` and only assert `last_language != "de"`, so this case passes.

**Suggestion:** Treat `"auto"` the same way Whisper/MLX `setup()` does: `start = self.start_language; self.last_language = None if start in (None, "auto") else start`. Keep Parakeet's override as-is. Add a test that constructs Whisper and MLX-audio-whisper with `start_language="auto"` and `last_language="de"`, calls `on_session_end()`, and asserts `last_language is None`.

src/speech_to_speech/STT/base_stt_handler.py
```diff
@@ -208,6 +215,19 @@ def _mark_completed_final_revision(self, output: Transcription) -> None:
         while len(completed) > self._MAX_COMPLETED_FINAL_REVISIONS:
             completed.popitem(last=False)
 
+    def reset_session_language(self) -> None:
+        """Drop the language detected for the finished conversation.
```
tavallaie: **[suggestion]** The `reset_session_language` docstring and the class-attribute comment narrate pipeline reuse, quote the Qwen3 progressive path, and restate the assignment. The rest of this file has almost no comments. The claim that `None` means detect-per-turn is also incomplete: the documented CLI value for that mode is `"auto"`.

**Suggestion:** Keep a short WHY (units are reused, so detected language is session state). Drop the Qwen3 walkthrough and the restatement of the assignment. If the detect-mode sentinel stays in the docstring, name both `None` and `"auto"`.

src/speech_to_speech/STT/base_stt_handler.py
tavallaie, line 228: Whisper and MLX setup map "auto" to None. This assignment would put "auto" back into last_language after --language auto, and a failed detect reports auto-auto instead of English.

Would the same map work here (None if start is None or "auto")? A test with start_language="auto", last_language="de", expecting None after on_session_end, would pin that.

src/speech_to_speech/STT/base_stt_handler.py
tavallaie, line 219: The rest of this file is pretty spare with comments. Would a shorter note be enough, that units get reused so we drop the detected language?

xprilion:
Thanks for working on this @Hotragn!

I verified @tavallaie's point regarding `--language auto`: resetting `self.last_language = self.start_language` puts `auto` back into `last_language`, which causes Whisper and MLX to emit `auto-auto` on turns where detection falls back.

While checking this against the handlers, I noticed two related details:

1. Parakeet's override still assigns `auto`:
   In `ParakeetTDTSTTHandler.on_session_end()`, line 644 runs after `super()`:
   ```python
   self.last_language = self.start_language if self.start_language else "en"
   ```
   Because `auto` is truthy, it overrides whatever `BaseSTTHandler` does and re-assigns `auto`. It needs a guard so `auto` maps to `en`:
   ```python
   self.last_language = self.start_language if self.start_language and self.start_language != "auto" else "en"
   ```

2. Qwen3 doesn't set `start_language`:
   `Qwen3ASRSTTHandler.configure_language()` sets `forced_language` and `last_language`, but never `start_language`. When configured with a fixed language (e.g. `--language de`), `self.start_language` remains `None` (the base class default), so `reset_session_language()` wipes `last_language` to `None`. 

Adding `self.start_language = language` in `configure_language()` should resolve it.

3. **Test coverage:**
   In `test_stt_session_language_reset.py`, adding a test case with `start_language="auto"` asserting `handler.last_language is None` (or `en` for Parakeet) will catch the `auto` leak. Currently, `assert handler.last_language != "de"` passes when `last_language` becomes `auto`.

I suggest the following in the `BaseSTTHandler`:

```python
def reset_session_language(self) -> None:
    start = getattr(self, "forced_language", None) or self.start_language
    self.last_language = None if start in (None, "auto") else start
```

Hotragn:
Thanks both — @tavallaie and @xprilion, this was a real catch and all of it is addressed in `022e3e2`.

The underlying mistake: I treated the reset as "restore `start_language`" when it should be "restore what `setup` established". Those differ precisely for `auto`, which `setup` maps to `None` because it's a request to detect, not a language code. So the reset was putting back a value `setup` had deliberately rejected.

**The `auto` normalization** now lives in the base:

```python
self.last_language = None if self.start_language == "auto" else self.start_language
```

**Qwen3 records `start_language`** (@xprilion's point 2). It only set `forced_language`, so the base default applied and a fixed `--language de` was wiped on session end.

Worth flagging that these two have to land together, which I didn't appreciate until I traced it: *before* recording `start_language`, Qwen3 was accidentally safe — the base default `None` meant the reset produced `None`. Adding `start_language = language` on its own would have introduced `auto-auto` via `_final_language_code`. There's now a test pinning `en-auto`.

**Parakeet's override** (@xprilion's point 1) is guarded. Its assignment runs after `super()` and `auto` is truthy, so it re-assigned `auto` over the base reset.

One correction on the stated symptom, since I checked each handler rather than assuming it applied uniformly: **`auto-auto` is only reachable in Qwen3-ASR.** Whisper and MLX-audio both gate on `if last_language in SUPPORTED_LANGUAGES`, so `auto` is filtered out and they report `en-auto`; Lightning stores `last_language` but never reads it back for output. Parakeet's is a bare `auto` rather than `auto-auto`, because it doesn't append the suffix. The fix is still right for all of them — restoring a value `setup` rejects is wrong regardless — but the user-visible `auto-auto` is Qwen3 only.

**On the test assertion** (@xprilion's point 3) — you were right and it mattered more than it looks. `assert handler.last_language != "de"` passed while the reset left `auto` behind. Now asserting the exact expected value per backend, via a table, plus explicit `start_language="auto"` cases.

That also exposed a second weakness in my own test file: it skipped on `ImportError` with no guard, and a half-installed local environment was silently skipping **four of the six** backends while the suite reported green. Skips now `fail` for anything that should be importable, matching the precedent in `test_llm_utils.py`. Locally that took it from 22 passed / 10 skipped to **32 passed, 0 skipped**.

@tavallaie — also trimmed the base-class comment; you were right that it was out of step with the file.

Full suite: **1649 passed, 3 skipped**; ruff, format and mypy clean. Twelve of the new cases fail against the previous commit.

tavallaie:
Re-checked `022e3e2` , this addresses the `auto` reset, Qwen3 `start_language`, Parakeet override, and the tightened tests. LGTM.
  Optional follow-up (not blocking): normalize `start_language` in Qwen3 the same way you normalize `auto` for `forced_language`, so reset stays
  consistent if `auto` ever arrives with different casing.

tavallaie (approved):

andimarafioti (approved):
Good work! Thank you 👍

Hotragn:
Thanks for the re-review @tavallaie, and for the merge.

Followed up on your optional note in #574. It turned out to be worth more than consistency: on the transformers Whisper backend a non-canonical spelling reaches `generate()` as a language *name* and raises on the first turn.

```
generate(language='en')   -> OK
generate(language='AUTO') -> ValueError: Unsupported language: auto.
```

It also defeated the reset this PR added — `--language AUTO` left `"AUTO"` in `last_language` for the next client, which is the leak we just closed:

```
whisper --language 'auto' -> after_reset=None
whisper --language 'AUTO' -> after_reset='AUTO'
```

Qwen3 needed no change in the end: @andimarafioti's merged version stores `self.start_language = self.forced_language` rather than the raw value I had, so the normalization you suggested is already there via `configure_language`. Nice improvement on what I wrote — it puts the canonical value in one place instead of two.

PR #561, opened 2026-09-13
Add configurable startup buffering to browser WebSocket playback

### LINKED ISSUE #557: Browser realtime demo can underrun TTS playback without a startup jitter buffer
## Summary

The browser realtime demo can produce unnatural / glitchy speech when the TTS backend streams audio close to or slower than realtime.

The issue appears to be in the browser playback path rather than in TTS generation itself.

In my case, adding a small startup jitter buffer in `demo/s2s-realtime-client.js` fixed the issue.

## Setup

* Hugging Face `speech-to-speech`
* Browser demo using WebSocket transport
* OpenAI-compatible local TTS backend
* VoxCPM2 as TTS
* TTS output: PCM s16le
* Pipeline output sample rate: 24 kHz
* Windows + Docker Desktop
* NVIDIA consumer GPU

## Symptoms

Without buffering, speech sounds mostly correct, but some syllables / words become unnatural or glitchy during realtime playback.

The problem is not present in the generated audio itself.

## Tests

I tested the same pipeline at several points:

1. VoxCPM native `generate_streaming()` output saved directly to WAV

   * sounds natural

2. Raw PCM returned by the local OpenAI-compatible TTS HTTP endpoint

   * sounds natural

3. PCM after the speech-to-speech backend resampling/output path

   * sounds natural when saved and played offline

4. Browser WebSocket realtime playback

   * some words / syllables sound unnatural

5. Browser playback changed to buffer the whole response before playback

   * sounds natural

6. Browser playback changed to use a startup buffer

   * 800 ms: much better, but occasional glitches remain
   * 1200 ms: playback becomes natural

This strongly suggests the browser playback queue is occasionally underrunning.

## Current browser behavior

The browser client currently forwards incoming PCM chunks to the AudioWorklet immediately from `_onAudio()`.

Conceptually:

```js
this._playbackNode.port.postMessage({
  kind: "audio",
  samples
}, [samples.buffer]);
```

There does not appear to be enough startup buffering to absorb TTS generation jitter.

## TTS realtime performance

Typical warm TTS performance in my setup is approximately:

* first chunk latency: ~0.28-0.35 s
* average RTF: ~1.1-1.2

So generation is close to realtime, but not always fast enough to continuously feed an immediately-started playback queue.

Even when average RTF is close to 1, individual chunk latency can vary enough to cause playback underruns.

## Workaround

I added a 1200 ms startup buffer to `demo/s2s-realtime-client.js`.

The client first accumulates around 1.2 seconds of assistant PCM audio, then starts playback. After that, incoming chunks are forwarded normally.

This completely fixed the audible glitches in my setup.

## Suggestion

Would it make sense for the browser demo to support a configurable playback jitter buffer, similar in purpose to `--playback-buffer-ms` used by the packaged local playback client?

Possible options:

* configurable startup playback buffer
* rebuffering when the AudioWorklet queue approaches underrun
* adaptive low-watermark jitter buffer

The browser demo currently seems to assume that the TTS backend can continuously provide audio at realtime speed or faster.

That assumption works well with fast TTS backends, but can cause audible artifacts with local TTS models running close to realtime on consumer GPUs.

karatarassul4-max:
## Summary

Adds configurable startup buffering for assistant audio in the browser WebSocket realtime demo.

The browser previously forwarded each incoming PCM chunk to the playback AudioWorklet immediately. When TTS generation runs close to realtime, temporary producer gaps can exhaust the playback queue and cause audible underruns.

This change adds a response-scoped startup reserve before playback begins, while preserving the existing immediate-playback behavior by default.

Fixes #557.

## What changed

* Add a WebSocket-only `playbackBufferMs` setting to the browser demo.
* Buffer PCM by accumulated sample duration before starting playback.
* Release buffered chunks in order once the threshold is reached.
* Flush short successfully completed responses even if they never reach the threshold.
* Discard pending audio on interruption, cancellation, failure, disconnect, or close.
* Prevent late audio/completion events from an old response from affecting a newer response.
* Keep WebRTC playback unchanged.
* Fix the AudioWorklet `clear` → immediate new audio boundary so a pending fade-out cannot strand the new queue.
* Document the latency/underrun tradeoff and the 1200 ms value reported in #557 as a tuning example rather than a universal default.

The default remains `0 ms`, so existing browser playback latency is unchanged unless buffering is explicitly configured.

## Testing

Added deterministic browser playback coverage that exercises the actual realtime client and playback worklet without requiring a TTS model, GPU, browser audio device, or timing sleeps.

The new tests cover:

* buffering below the startup threshold
* exact threshold crossing and chunk ordering
* completed short responses
* interruption before and during playback
* cancelled, failed, and incomplete responses
* stale response events
* disconnect and close handling
* zero-buffer compatibility
* WebRTC isolation
* bounded producer gaps
* fractional buffer durations
* settings persistence and validation
* `clear` followed immediately by new audio

Validation:

* `node demo/tests/playback.test.mjs` — **18 passed**
* focused demo Python tests — **36 passed**
* Agents SDK adapter tests — **3 passed**
* transcription lifecycle tests — **7 passed**
* `npm --prefix demo run test:agents` — **passed**
* broader relevant Python subset — **145 passed, 1 pre-existing failure**
* `ruff check` — **passed**
* `ruff format --check` — **passed**
* `mypy src/` — **passed**
* `git diff --check` — **passed**

The one broader-suite failure (`test_response_cancel_flushes_queues`) reproduces unchanged on a clean upstream checkout and is unrelated to this patch.


andimarafioti:
This looks complementary to #589 rather than overlapping with it. We recently investigated repeated audible crunches in a deployed WebSocket pipeline and initially suspected playback-buffer starvation as well. However, with a 196 ms startup buffer we measured zero simulated underruns, while the server was delivering audio at about 0.25× realtime (roughly 4× faster than playback). The glitches still occurred throughout each response.

The cause in that case was server-side 16→24 kHz conversion: `resample_poly` was being restarted independently for each outbound batch (typically about 192 ms of source audio), introducing discontinuities at batch boundaries. Keeping the resampler state across chunks in #589 reduced the measured boundary-step p95 from 2,405.5 to 1,407.3 PCM units; after the fix it fell below the same output waveform’s ordinary p99 of 2,385.0. The audible crunches disappeared in a live test.

So I think the two changes protect against distinct failure modes:

- this PR prevents genuine client playback underruns when producer/network gaps consume the queue;
- #589 prevents discontinuities already embedded in the PCM stream, which additional buffering cannot repair.

The `clear` → immediate-audio worklet fix here also covers another useful boundary case around interruption. The PRs touch separate files, so they should compose cleanly. Our deployed results support keeping both layers of protection, with a moderate configurable startup reserve rather than treating a large buffer as a substitute for continuous server-side resampling.

PR #566, opened 2026-09-16
fix: keep active response playing on listener backchannels

### LINKED ISSUE #565: Listener backchannels abort active response mid-playback
While testing the browser realtime demo, I noticed that uttering natural listener acknowledgements/backchannels (such as "mm-hmm", "uh-huh", "alright", or "go ahead") while the assistant is speaking immediately aborts the in-progress response, clears the audio playback, and triggers a brand-new reply answering the acknowledgement.

I might be missing an existing setting or flag for this, but I wanted to check if this is intended behavior or if there is interest in allowing short passive backchannels to pause and resume playback instead of triggering a destructive cancel.

---

### Steps to Reproduce

1. Start the server: `speech-to-speech serve`.
2. Connect via the browser demo at `http://localhost:8765` (or connect via `talk`).
3. Ask a question that produces a long response (say "Explain how the solar system formed").
4. While the assistant is actively speaking sentence 1, say "mm-hmm" or "right".

---

### Observed Behavior

1. The assistant abruptly cuts off mid-sentence.
2. The queued audio for the rest of the explanation is cleared.
3. The model receives "mm-hmm" as a new user prompt and begins answering it from scratch.

---

### Expected Behavior

Brief listener feedback/backchannels should ideally not abort the active assistant turn. Instead, playback could briefly pause, and then resume the existing response once the short vocalization ends (saving full cancellations for substantive interruptions or queries).

---

### Environment

* **Version:** `speech-to-speech` v1.0.0 (`main` at `16d7f98`)
* **Platform:** macOS (Apple Silicon) / Linux

### Screenshot

Image

xprilion:
Fixes #565

## Problem

When the assistant is speaking, any user vocalization exceeding `_active_speech_min_ms` (384 ms) - including passive acknowledgements like `"mm-hmm"`, `"right"`, or `"great"` - triggers an immediate hard cancellation. The active response is aborted mid-playback, unplayed audio is discarded, and the LLM is prompted to reply to the backchannel.

On Apple Silicon, this also caused lock starvation: `Qwen3TTS` held the global `_mlx_lock` continuously across the entire multi-second response, causing `ParakeetSTT` to time out after 5.0 seconds with `Failed to acquire compute lock for final transcription` and freeze the pipeline.

## Solution

1. Language-Agnostic Backchannel Classification (`LLM/utils.py`):
   * Evaluates structural invariants without hardcoded wordlists: non-question, non-stop-command, `≤ 2` words, and duration ceiling (`≤ 2.0s`).
   * Backchannels allow active playback to continue uninterrupted; questions and substantive commands (`len(words) > 2`, `?`, `!`) trigger immediate cancellation.

2. Continuous Playback (Zero Gap / Zero Audio Skipping):
   * The assistant continues speaking without pausing or withholding audio buffers during passive feedback.
   * On confirmed interruptions, playback is cleared immediately on `response.done(status="cancelled")`.

3. Chunk-Level MLX Lock Yielding (`TTS/qwen3_tts_handler.py`):
   * Wrapped Qwen3-TTS generator steps in `_MLXChunkIterator` so `_mlx_lock` is held only for ~150 ms per chunk and released in between.
   * `ParakeetSTT` can acquire the lock in <150 ms without starving or timing out.

4. Turn Commitment on Backchannel Discard (`service.py`):
   * Once a backchannel is identified, `speculative_turns.commit()` marks the turn closed so VAD flushes its audio prefix and does not concatenate the backchannel into subsequent user inputs.

## Validation

- [x] `uv run pytest tests/test_backchannel_detection.py tests/openai_realtime/test_backchannel_gating.py` - **9 passed**
- [x] `node --test demo/tests/agents-adapter.test.mjs demo/tests/agents-websocket.test.mjs` - **5 passed**
- [x] Full test suite: `uv run pytest tests/` - **1,626 passed, 3 skipped**
- [x] `uv run ruff check src/ tests/` - **clean**
- [x] `uv run ruff format --check src/ tests/` - **175 files formatted**
- [x] `uv run mypy src/` - **clean**
- [x] Verified live end-to-end on macOS with the browser demo: assistant continues speaking continuously through `"Right."` and `"Mm-hmm."`, and cancels cleanly on questions or commands.


xprilion:
This approach doesn't scale well for languages like Chinese/Japanese, or for languages with strong interruption words that can be said in less than 1.2s. I guess the solution, if required, needs more discussion.

PR #570, opened 2026-09-19
Stop sending idle WebRTC audio packets

### LINKED ISSUE #558: Sending silence on RTC causes noticable per-client background CPU usage
I've noticed that when the AI isn't speaking, speech-to-speech has a noticable CPU usage. I've found two causes for this, the major one being the fact that, when on a WebRTC connection, s2s constantly sends 50 packets/s of silent audio (I think this is standard in WebRTC).

The network usage seems to be negligeble (~150 bytes/s).

This is quite annoying in setups where s2s is running 24/7, only speaking every now and then.

I have managed to mitigate this locally by disabling the sender when it runs out of data to send:
https://github.com/huggingface/speech-to-speech/blob/16d7f98ff712fb082d53497937f5456667c84680/src/speech_to_speech/api/openai_realtime/webrtc_session.py#L146-L147

and enabling it when there's more data to send:
https://github.com/huggingface/speech-to-speech/blob/16d7f98ff712fb082d53497937f5456667c84680/src/speech_to_speech/api/openai_realtime/webrtc_session.py#L120-L121

What I'm not sure about is the best way to disable it.

The way to completely eliminate the CPU usage would be to call `stop` on the `RTCRtpSender`, then recreate it when it's needed. I'm not sure if this requires renegotiation, or if there's too much of an overhead to starting/stopping a sender too often (once every few seconds in a normal conversation).

Another way would be to use `replaceTrack` to add/remove the track. Currently in aiortc, [this will result in a task that wakes up 50 times per second to check the track](https://github.com/aiortc/aiortc/blob/8a2864630bf417d977a06dda8d0254b1c1501bfb/src/aiortc/rtcrtpsender.py#L364-L367), which is not ideal, but is much better than the current implementation.

Would you accept a PR that implements one of these approaches?

ousamabenyounes:
## Summary

- pause outbound WebRTC audio reads when the PCM buffer is empty, so aiortc does not continuously encode and transmit silent frames
- wake blocked reads when audio arrives or the track stops
- preserve RTP media-clock timestamps and normal pacing when audio resumes after an idle period

Fixes #558

## Test verification (RED → GREEN)

- RED, upstream `main` production plus the regression test: `pytest tests/openai_realtime/test_webrtc.py::TestPipelineAudioTrack::test_recv_waits_for_audio_when_idle -q` failed at `assert not recv_task.done()` because the old track returned a silent frame immediately (`1 failed`)
- RED review follow-up: the encoder-level test failed with `assert (3840 - 960) == 960`, proving aiortc's Opus encoder saw a discontinuous `2880`-sample gap between the two resumed packets (`1 failed`)
- RED race follow-up: the wait-setup race test failed on upstream production because the event-backed wakeup protocol did not exist (`1 failed`)
- GREEN, patched tree: `pytest tests/openai_realtime/test_webrtc.py::TestPipelineAudioTrack -q` completed with `7 passed`
- full local suite, baseline → final: Ruff checks passed; mypy passed across 106 source files; WebRTC pytest improved from 22 to 25 passing tests; changed executable lines reached 100% diff coverage; the Agents SDK WebRTC test remained 1/1 passing
- the aggregate formatter check reports the same pre-existing drift in two untouched files (`src/speech_to_speech/LLM/language_model.py` and `tests/test_speculative_turns.py`) on both baseline and final runs


src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -131,12 +140,26 @@ async def recv(self) -> av.AudioFrame:
         if self.readyState != "live":
             raise MediaStreamError
 
+        waited_for_audio = False
+        while not self._buffer:
+            self._data_available.clear()
+            if self._buffer:
```
bolshoytoster, line 146: Is this check needed? Does `self._data_available.clear()` block?

src/speech_to_speech/api/openai_realtime/webrtc_session.py
```diff
@@ -131,12 +140,26 @@ async def recv(self) -> av.AudioFrame:
...
+            await self._data_available.wait()
+            if self.readyState != "live":
+                raise MediaStreamError
+
+        now = time.time()
         if self._start is None:
-            self._start = time.time()
+            self._start = now
             self._timestamp = 0
+        elif waited_for_audio:
```
bolshoytoster, line 157: Should this branch still sleep like the else branch? Does this even need its own branch?

src/speech_to_speech/api/openai_realtime/webrtc_session.py
ousamabenyounes: No — `asyncio.Event.clear()` is synchronous and never blocks, and no `await` runs between the `while not self._buffer` condition and `clear()`, so in single-threaded asyncio the buffer cannot change in between. The recheck was dead, removed it in 8f7fdbf.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
ousamabenyounes: It shouldn't sleep: the first frame after an idle gap must go out immediately. And it no longer needs its own branch — folded into the paced `else` in 8f7fdbf. The timestamp advance is shared, and after re-anchoring `self._start` the paced `wait` resolves to 0, so no sleep happens on resume.

ousamabenyounes:
Addressed both review threads in 8f7fdbf: removed the redundant post-`clear()` buffer recheck (single-threaded asyncio, no `await` in between, and `clear()` doesn't block) and folded the post-idle branch into the paced `else` (shared timestamp advance; after re-anchoring `self._start` the wait resolves to 0, so resume still emits immediately).
Behavior is unchanged — verified locally: `ruff check`/`ruff format --check`/`mypy` clean on the changed file, and pytest shows 7/7 `TestPipelineAudioTrack` + 25/25 `test_webrtc.py` passing, with the 3 regression tests still RED on unmodified `main`.
Still open: CI on the new commit and maintainer review.

src/speech_to_speech/api/openai_realtime/webrtc_session.py
bolshoytoster: > It shouldn't sleep: the first frame after an idle gap must go out immediately.

I know it's probably quite unlikely, but what if the gap is <20ms? We could probably ignore that case, since it probably won't happen often, and even if it did, it shouldn't cause much trouble.

